PR #1, opened 2018-09-30
Add Docker + Docker Compose

George334456:


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ruby:2.5-alpine
+
+# Add tzdata because the Gemfile doesn't successfully add the dependency via geminstall.
+RUN apk update && apk add build-base nodejs postgresql-dev tzdata
```
padabap: Separate packages onto separate lines (with backslash) for easier reading + editing in future

padabap (changes requested):

padabap (approved):

PR #3, opened 2018-10-08
Update schema and create seed data

padabap:


db/migrate/20181005194048_add_columns_to_applicant.rb
```diff
@@ -12,6 +12,7 @@ def change
     add_column :applicants, :commentary, :text
     add_column :applicants, :dept, :string
     add_column :applicants, :year_in_program, :integer
-    add_column :applicants, :full_time, :string
+    add_column :applicants, :is_full_time, :boolean
+    add_column :applicants, :is_grad, :boolean
```
George334456: Let's spell it out in full.

db/schema.rb
```diff
@@ -16,20 +16,21 @@
...
-    t.string 'last_name'
-    t.string 'email'
-    t.string 'phone'
-    t.text 'address'
-    t.text 'commentary'
-    t.string 'dept'
-    t.integer 'year_in_program'
-    t.string 'full_time'
+  create_table "applicants", force: :cascade do |t|
+    t.datetime "created_at", null: false
```
George334456, line 19: This is probably legacy, but `rubocop` should change these to single quotes

db/seeds.rb
```diff
@@ -1,9 +1,52 @@
...
-#   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
-#   Character.create(name: 'Luke', movie: movies.first)
+# This file should contain all the record creation needed to seed the database with its default
+# values. The data can then be loaded with the rails db:seed command (or created alongside the
+# database with db:setup).
+
+a = Applicant.create(
+  first_name: 'George',
+  last_name: 'Wu',
+  utorid: 'wugeorge',
```
George334456, line 10: What's the difference between utorid and student number?

George334456 (changes requested):
Just some questions

db/seeds.rb
padabap: both are unique, but i presume HR wants both. Or at least HR needs student number, and CS department would need utorid

db/schema.rb
padabap: oops i forgot to run rubocop manually before commiting. your pre-commit will fix this :)

db/seeds.rb
George334456: Well I meant that something like wugeorge vs ########. I hope both are unique. xD

George334456 (approved):

PR #4, opened 2018-10-08
Added validations

George334456:
Added more validations

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -5,6 +5,9 @@
 class Applicant < ApplicationRecord
   has_many :preferences
   has_many :positions, through: :preferences
+
+  validates_presence_of :first_name, :last_name, :email, :student_number, :utorid
+  validates_uniqueness_of :student_number, :utorid
 end
 
 # == Schema Information
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -30,3 +30,8 @@ end
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
 #
+# Indexes
+#
+#  index_applicants_on_student_number  (student_number) UNIQUE
+#  index_applicants_on_utorid          (utorid) UNIQUE
+#
--- a/app/models/preference.rb
+++ b/app/models/preference.rb
@@ -5,6 +5,9 @@
 class Preference < ApplicationRecord
   belongs_to :applicant
   belongs_to :position
+
+  validates_presence_of :priority, :applicant, :position
+  validates :priority, numericality: { only_integer: true, less_than_or_equal_to: 3 }
 end
 
 # == Schema Information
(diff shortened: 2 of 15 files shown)


app/models/applicant.rb
```diff
@@ -5,6 +5,9 @@
 class Applicant < ApplicationRecord
   has_many :preferences
   has_many :positions, through: :preferences
+
+  validates_presence_of :first_name, :last_name, :email, :student_number, :utorid
+  validates_uniqueness_of :student_number, :utorid
```
padabap, line 10: Validations should be done on both the db schema and application level.

Recommend reading both these articles for background:
- https://robots.thoughtbot.com/the-perils-of-uniqueness-validations 
- https://robots.thoughtbot.com/validation-database-constraint-or-both

app/models/session.rb
```diff
@@ -4,6 +4,14 @@
 # an applicant that applies to Fall 2018, round 1 applies to the "Fall 2018 Round 1" session.
 class Session < ApplicationRecord
   has_many :positions
+
+  # TODO: Since we're using enums for semester, we might have to change semester to integer for type.
```
padabap, line 8: is this optional or a must do? :/

app/models/preference.rb
```diff
@@ -5,6 +5,9 @@
 class Preference < ApplicationRecord
   belongs_to :applicant
   belongs_to :position
+
+  validates_presence_of :priority, :applicant, :position
+  validates :priority, numericality: { only_integer: true, less_than_or_equal_to: 3 }
```
padabap: Perhaps we should make this an enum too? something like `%i[low medium high]` or at least ensure that it can't be negative?

padabap (changes requested):

app/models/preference.rb
George334456: I kept it as 1 - 3 because we might have other priority.

app/models/session.rb
George334456: Probably must do.

padabap: Hm, I just tried to insert into db on your branch, and it seems to work without needing to change to an int?

app/models/applicant.rb
George334456: Hmm, maybe we should revisit this regarding `belongs_to` validations. Also we need to talk about uniqueness constraints regarding multiple scopes.

app/models/session.rb
George334456: 🤔

George334456: Can you see what it's saved as postgres side?

padabap: Insertions work for me. Will create ticket to double check if this TODO can be removed

padabap (approved):

PR #6, opened 2018-10-10
Add api for positions

padabap:


app/controllers/api/v1/positions_controller.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Api::V1
+  # Controller for Positions
+  class PositionsController < ApplicationController
```
George334456, line 5: We might have to be prepared for authorization.

George334456: Also please document (perhaps in readme or wiki) on API documentation.

IE
`GET`
Make a table and a description

Sample JSON response would be great as well.

app/controllers/api/v1/positions_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    private
+
+    # Use callbacks to share common setup or constraints between actions.
+    def set_position
+      @position = Position.find(params[:id])
+    end
+
+    # Only allow a trusted parameter "white position" through.
+    def position_params
+      params.require(:position).permit(:title, :excerpt, :description, :upvotes)
```
George334456: Plz explain.

George334456: Is it like a position ID that you're requiring?

Also what's an upvote.

app/controllers/api/v1/positions_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    # GET /positions/1
+    def show
+      render json: @position
+    end
+
+    # POST /positions
+    def create
+      @position = Position.new(position_params)
```
George334456, line 22: For this POST, please provide a sample JSON body to give. Also a sample response JSON if applicable.

app/controllers/api/v1/positions_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+      if @position.update(position_params)
+        render json: @position
+      else
+        render json: @position.errors, status: :unprocessable_entity
+      end
+    end
+
+    # DELETE /positions/1
+    def destroy
+      @position.destroy
```
George334456, line 42: Is destroy safe here? Should we be hard deleting?

app/controllers/api/v1/positions_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+        head :no_content, status: :ok
+      else
+        render json: @position.errors, status: :unprocessable_entity
+      end
+    end
+
+    private
+
+    # Use callbacks to share common setup or constraints between actions.
+    def set_position
```
George334456, line 53: Indentation for these methods?

app/controllers/api/v1/positions_controller.rb
George334456: Looking at the tutorial, are you sure these things being permitted even exist?

app/models/position.rb
```diff
@@ -12,6 +12,16 @@ class Position < ApplicationRecord
 
   validates_presence_of :course_code, :openings, :session
   validates :openings, numericality: { only_integer: true }
+
+  def as_json(_options = {})
+    super(
+      except: [:session_id],
```
George334456: Shouldn't we use `%i[session_id]`?

I thought the linter would catch this.

app/models/position.rb
```diff
@@ -12,6 +12,16 @@ class Position < ApplicationRecord
...
+      include: {
+        session: { only: %i[id year round semester] },
+        instructors: { only: %i[id first_name last_name] }
```
George334456, line 21: Should we include `utorid`?

script/pre-commit
```diff
@@ -33,7 +30,7 @@ then
 	fi
 	
 	# Run rubocop on the staged files
-	bin/bundle exec rubocop ${FILES}
+	rubocop ${FILES}
```
George334456: I kinda made this change locally, but is this what Eugene suggested?

This won't work on staged files (as we've discussed), and this changes behaviour to only the ones we include.

George334456 (changes requested):
Documentation, styling, and some comments.

app/controllers/api/v1/positions_controller.rb
padabap: ahehehehe my bad will fix

app/controllers/api/v1/positions_controller.rb
padabap: https://github.com/uoft-tapp/tapp/wiki/APIs

app/controllers/api/v1/positions_controller.rb
padabap: don't think they should be indented? rubocop is happy with it

app/models/position.rb
padabap: guess linter doesn't complain if it's just one element, but will update

app/models/position.rb
padabap: i can't think of a use case for it, this can always be modified later though

app/controllers/api/v1/positions_controller.rb
padabap: @reidka what do you think?

reidka: I used to be very hesitant to delete anything from the database, but I think in this context it will be useful.  I'm thinking of the case where we upload some positions and then realize that we uploaded the wrong ones.

We will want to make sure that there are no other connections to the position before deleting.  In particular, no "assignments".  Primarily for database consistency.

app/controllers/api/v1/positions_controller.rb
George334456: Could you also make it more in line with this:
https://gsa.github.io/api-documentation-template/api-docs/console/

We can talk about parameters, status codes, and sample responses on success/failure.

script/pre-commit
padabap: will do some more testing with this. We ran rubocop by calling `ruby rubocop.rb` before in flipp project.

app/controllers/api/v1/positions_controller.rb
padabap: Ooh looks cool! Yes, for sure. More effort though, so will put it as a new task on trello

PR #13, opened 2018-10-30
Add user table and model for roles

padabap:


app/models/user.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# A class representing a user. This holds information regarding known users and their roles
+# to determine what they should be allowed access to.
+class User < ApplicationRecord
+  enum role: %i[admin instructor]
+  validates_presence_of :utorid
+  validates_uniqueness_of :utorid
```
George334456, line 8: Please have a database constraint for this.

For future reference, you can run `consistency_fail` as a gem in order to find out what unique indices you need to back up your db with based on our models..

spec/factories/users.rb
```diff
@@ -0,0 +1,24 @@
+FactoryBot.define do
+  factory :user do
+    utorid { "utorid#{Faker::Number.number(2)}" }
+    
+    trait :admin do
+      role { 0 }
```
George334456: I think we should do symbols instead of integers. It's hard to read otherwise and we need to look back. Also, I don't like magic constants especially if ruby decides to change how the symbol to integer mapping is changed.

George334456 (changes requested):
Make sure database side is consistent w/ our validations.

app/models/user.rb
padabap: o good catch

George334456 (approved):

PR #14, opened 2018-10-30
Update time to_json

padabap:


config/initializers/time_with_zone.rb
```diff
@@ -0,0 +1,5 @@
+class ActiveSupport::TimeWithZone
```
George334456, line 1: Does this class change the configuration for everything to now just year/month/day?

padabap: it makes it so when you call `to_json` on a datetime object, you only get the year/month/day

PR #15, opened 2018-11-01
file suffix should have been .sh

jmzaleski:


script/init-script.sh
```diff
@@ -0,0 +1,3 @@
+gem install rubocop
```
jmzaleski: on my laptop this had to be sudo

script/init-script.sh
```diff
@@ -0,0 +1,3 @@
...
+ln -s ./script/pre-commit .git/hooks/pre-commit
```
padabap: @jmzaleski This would point to `./script/pre-commit` from the `.git/hooks` directory. 

```
~/tapp > ls -l .git/hooks
total 10414 Nov 08:43 pre-commit -> ./script/pre-commit
```
It should be from the perspective of where `script` is located from the `.git/hooks` directory, thus the navigating back up to the root directory from `.git/hooks` to get `../../script/pre-commit`

script/init-script.sh
padabap: If you update this line to add the `sudo`, I can merge it in :)

PR #16, opened 2018-11-11
TAPP-31: Added API route for Instructors and unit tests

George334456:
Added API route for instructors.

To add API documentation and things like that to the wiki.


spec/controllers/api/v1/instructors_controller_spec.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+RSpec.describe Api::V1::InstructorsController, type: :api do
+  let!(:instructor) { FactoryBot.create(:instructor) }
+
+  # This is the minimal set of attributes required to create a valid Position.
```
padabap: Update comment from position to instructor

spec/controllers/api/v1/instructors_controller_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      "first_name": 'Julie',
+      "utorid": 'wongjulie305',
+      "email": 'julie.wong@mail.utoronto.ca'
+    }
+  end
+
+  # This should return the minimal set of values that should be in the session
+  # in order to pass any filters (e.g. authentication) defined in the
+  # InstructorsController. Be sure to keep this updated.
+  let(:valid_session) { [] }
```
yyjoycewoo: Does instructor need session object at all? Don't think there's an association, so you can get rid of this, plus all things related to sessions in the tests

PR #32, opened 2018-11-29
New position

padabap:


frontend/src/modules/positions/components/EditPositionModal.jsx
```diff
@@ -66,6 +66,7 @@ class EditPositionModal extends React.Component {
         )
     render() {
         const { position, show, handleHide } = this.props
+        console.log(this.props)
```
chrislessard: Is there a reason for this ?

frontend/src/modules/positions/components/NewPosition.jsx
```diff
@@ -1,42 +1,164 @@
 import React from "react"
 import { connect } from "react-redux"
 import { createNewPosition } from "../actions"
+import { fetchInstructors } from "../../instructors/actions"
 import { Grid, Col, Button } from "react-bootstrap"
+import moment from "moment"
```
chrislessard: could we stick to convention and use `import { moment } from "moment"` ?

frontend/src/modules/positions/components/NewPosition.jsx
```diff
@@ -1,42 +1,164 @@
...
+const InstructorInput = ({ curValue, label, onChange, required }) => <div>
+        <label style={{ width: "100px" }}>{required ? label + '*' : label}</label>
+        <input value={curValue} 
+               onChange={onChange} 
+               required={required} />
+    </div>
+
+const InstructorList = ({ curValue, instructors, setInstructor }) => {
+    const valid = validInstructor(curValue, { instructors })
+    // console.log(thing, curValue, instructors)
```
chrislessard: please drop this comment

frontend/src/modules/positions/components/NewPosition.jsx
```diff
@@ -1,42 +1,164 @@
...
+                                                />
+                                            { this.state.instructor !== '' &&
+                                            <InstructorList 
+                                                curValue={this.state[value]} 
+                                                setInstructor={this.setInstructor} 
+                                                key='instructorList' 
+                                                instructors={this.props.instructors} /> }
+                                        </div>
+
+                                case 'qualifications':
```
chrislessard: is this needed ?

chrislessard (commented):
Just a few small things to change, otherwise looks good to me !

frontend/src/modules/positions/components/NewPosition.jsx
padabap: In switch statements, when a `case` is matched, it runs what's in the `case`, and all following statements as well. So in this case, it will run the `return <TextboxInput>` block for both qualifications and duties.

More info / live example of an empty case clause here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

frontend/src/modules/positions/components/EditPositionModal.jsx
padabap: I think you commented on an older commit - should already be fixed

frontend/src/modules/positions/components/NewPosition.jsx
padabap: I think you commented on an older commit - should already be fixed

frontend/src/modules/positions/components/NewPosition.jsx
padabap: Updated!

frontend/src/modules/positions/components/NewPosition.jsx
chrislessard: Ah, I see, thanks for clarifying.

chrislessard (approved):
Looks good to me !

PR #43, opened 2019-01-16
fix(documentation): Update the README

### LINKED ISSUE #42: README needs a few updates
We have a few small issues pertaining to the readme that should be updated:

1. One of our prerequisites is listed as "For Linux users only, Docker Compose". Later in the document, we ask that users invoke `docker-compose ...`. I see this being confusing to new members on mac or windows who don't realize that docker-compose comes w/ docker on those OS installations. We could resolve this with a small rephrasing indicating that if the dev is on mac or win, they already have it. 

2. Invoking `./script/init-setup.sh` out of the box doesn't work out, as the script won't likely have executable permissions. It's nitpicky, but we should write "Run the script with your shell. In bash, for example, invoke `bash script/init-setup.sh`.

3. Under "Running in a docker container" we first instruct the user to invoke `docker-compose up`. This will fail if docker isn't actually running on the users machine, we should specify to make sure it's running first. 

4. The information under "Accessing Docker Images through a shell command" seems to be both somewhat redundant and at odds with the earlier mention of using `docker-compose exec tapp sh' to run the tapp container and access the app. Why do we advise docker in one case and docker-compose in the other ?

5. There are grammatical errors throughout the document, mostly related to sentence structure and word choice.

At the time of writing, we're also experience an issue w/ psql invocations. That's being tracked in #40

chrislessard:
This patch makes a number of changes to the README, notably:

- Rephrasing of sentences and grammatical corrections
- Update of prerequisites
- General clarification of docker related steps
- Introduction of section on running the frontend

Resolves #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ After cloning the repo and navigating into the cloned directory, run the init sc
 in your shell. To do so in bash, for example, invoke:
 
 ```
-bash script/init-setup.sh
+bash script/init-script.sh
 ```
 
 ### Running the backend with Docker
@@ -87,7 +87,7 @@ the DB container, then invoking:
 $ psql tapp_development tapp
 ```
 
-## Running the frontend
+## Running the React frontend
 
 To run the front end, open a new terminal instance and navigate into the `frontend` directory. Once there, invoke
 
--- a/README.md
+++ b/README.md
@@ -2,59 +2,108 @@
 TA application, assignment, and matching.
 
 ## Getting Started
-These instructions will get you a copy of the project up and running on your
-local machine for development and testing purposes. The entire application is
-Dockerized, using Ruby 2.5.1 and a Postgres database.
+These instructions will get a copy of the project up and running on your
+local machine for development and testing purposes. Currently, the backend 
+(using Ruby 2.5.1 and a Postgres database) is dockerized and the frontend 
+must be run manually.
 
 ### Prerequisties
 
 1. [Docker](https://docs.docker.com/install/#supported-platforms)
-2. For Linux users only, [Docker Compose](https://docs.docker.com/compose/install/)
+2. [Docker Compose](https://docs.docker.com/compose/install/) (Must be installed seperately on Linux)*
 3. [Ruby](https://www.ruby-lang.org/en/documentation/installation/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,68 +2,117 @@
...
 ```
-docker-compose exec tapp sh
+docker-compose exec [service] sh
 ```
 
-### Initializing DB
-To create and setup your local development database, simply navigate into the
-rails container and run the following rake task. It will create your local database,
-run all migrations, and populate the db with seed data. 
+To interact with the rails app, invoke `docker-compose exec tapp sh` and to
```
jmzaleski, line 68: what means "interact" ??

README.md
```diff
@@ -2,68 +2,117 @@
...
+with seed data. 
+
+Once your DB is setup, you can gain SQL access to it by first navigating into 
+the DB container, then invoking:
+
 ```
 $ psql tapp_development tapp
 ```
 
+## Running the frontend
```
jmzaleski: I know this is a hard question.. but what is the "front end" in this usage?
I reckon it's how we serve the javascript of our app -- but I don't understand how we might start up the "front end" in production differently than from dev. 

Maybe I'm below the threshold of a reasonable knowledgeable reader?

README.md
chrislessard: @jmzaleski It means that you can enter an interactive terminal session within the container, I took the wording from docker resources. Under such a session, you could run our rails specs, for example. Do you want this to be clarified further in the document ?

README.md
chrislessard: @jmzaleski Given how most people tend to use the term these days I don't think it will pose a problem. That said, the "frontend" here is how we serve the collection of *both* html, stylings (i.e. css) and javascript (note that rails does not serve any of these things). In other words, it's the presentational aspect of the project. 

I specified in the document that this is separate from the "backend" (the logical/state aspect of the app). In retrospect, this distinction might deserve its own section altogether.

dingtaoliu: I think to clarify this point, maybe you can change "frontend" to "react frontend app" since the frontend is sort of its own app, separate from the Rails API server?

chrislessard: Sure, sounds good to me. I'll add a commit addressing that.

Issue #41, opened 2019-01-14
the apps's front end component is not containerized


chrislessard:
As it stands, only the backend portion of our app is containerized. To see the frontend code, I need to navigate manually into the frontend directory, perform a `yarn install` and `yarn start`. 

It would be nice if I could perform a single invocation to run the container, spawning both the backend server and the frontend server at the same time.

PR #46, opened 2019-01-21
Containerize the frontend

chrislessard:
Hi,

These commits effectively containerize the "frontend" of our app (that is, the React app). This will allow us to spin three containers, one for the DB, api (rails app) and frontend (react app) in a single `docker-compose up --build` invocation.

Solves #41


dingtaoliu (approved):
Looks good to me!  I will merge the pull request.

PR #60, opened 2019-01-27
TAPP-57: fixing delete button functionality on edit position popup

dingtaoliu:
Fixed deletePosition function in positions/action.js to make a delete request to the api and to pass the payload object to the reducer.


chrislessard (approved):
Looks good to me, thanks for solving this.

Issue #54, opened 2019-01-25
[Api] Drop rounds from the schema


chrislessard:
In our last meeting, we decided to drop the notions of "rounds" from our database schema. This will have to be refactored out of the backend code.

PR #74, opened 2019-02-09
[Api] Drop rounds from the schema 

chrislessard:
This series of commits refactors our database scheme, dropping the rounds table from it altogether. The necessary refactoring are put in place to associate positions directly with sessions.

Solves #54


dingtaoliu (approved):
Looks great to me!  Models, migrations, and tests have all been refactored accordingly to remove rounds.

Issue #76, opened 2019-02-10
[Update] update ajv version in the package.json


Yifang-Zhang:
The old version ajv is causing problems when we do the docker-compose up. Update the ajv version in the package.json.

Issue #72, opened 2019-02-09
Drop issue number commit convention


chrislessard:
We should drop the convention to mention issue numbers with a hashtag, since this is the token for comments in GIT. If someone needs to find an issue linked to some commit, they can do so by looking in the associated PR.

PR #77, opened 2019-02-10
Update: update ajv version in package.json

Yifang-Zhang:
This commit updates the ajv version, thus solving the problem when I do the docker-compose up.

Solves #76

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -39,5 +39,8 @@
     ],
     "devDependencies": {
         "redux-devtools-extension": "^2.13.5"
-    }
+    },
+    "resolutions": {
+        "ajv": "6.8.1"
+      }
 }


chrislessard (commented):
Hi @Yifang-Zhang, to be honest I'm not totally comfortable merging this in quite yet. 

For starters, I'm not actually able to reproduce the issue described in #72. I just checked out a fresh copy of the latest master (head is 2460fc40d0423c9cb2743b40477bae81eda6c97a), did a fresh `docker compose up --build` and was able to visit all pages of the frontend. No apparent issues seemed to come up. Can you please provide the exact steps for bug reproduction ?

I'm also a bit skeptical about the wording of updating ajv because we never had it in `package.json` in the first place. It appears to be a json schema validator. Are you using it as part of your work on data importation ?

Issue #70, opened 2019-02-03
[Frontend] Create a button to upload .csv file when creating a new position


Yifang-Zhang:
For this ticket, we should create a button to upload .csv file locally, and read the data from that file, and feed it into the input box of related areas.

Yifang-Zhang:
In this issue, I will also fix a small bug resulting from the delete call in position. It has "no content" errors in the notification, but this should not be an issue for delete calls.

PR #82, opened 2019-02-18
[Frontend] create a button to upload .csv file to create a new position

Yifang-Zhang:
This pr deals with the frontend files. It will read csv files, and fetch the new route call in the backend, which will be included in another pr.  #70


frontend/src/modules/positions/components/NewPosition.jsx
```diff
@@ -95,7 +96,27 @@ class NewPosition extends React.Component {
...
+                hours: data[i][7],
+                openings: data[i][8],
+                start_date: data[i][9],
+                end_date: data[i][10]
+            });
+            this.props.importNewPosition(this.state)
+        }
+        
+    }
+
```
dingtaoliu, line 125: I'm a bit lost here, why do we need to save the position information in the state before we dispatch it as an action? Could we just directly dispatch the position?  It seems like the position information will be lost in the next iteration

dingtaoliu (changes requested):
Just one comment about setting the state with the position information as you parse through the csv, other than that it looks good to me!

frontend/src/modules/positions/components/NewPosition.jsx
Yifang-Zhang: Yeah, I could directly dispatch the position. I will change it.

Issue #81, opened 2019-02-18
[Backend] Add new route for uploading csv file at frontend


Yifang-Zhang:
The new route will receive the data from the frontend, verify the validity of data and do a save once it is verified. If the data record already exists in the database, the route shall update its fields. If not valid, then show a notification to remind the user of the corresponding errors.

PR #83, opened 2019-02-19
[Backend] Add a new route for uploading csv file at frontend

Yifang-Zhang:
This pr added a new route to import data from csv file. It will check whether the data is valid first, and create a new object if so. If the data entered has duplicate fields (course_code and session_id), which has a record in the database, the route will update its fields accordingly. If the data is not valid, it will return an error in the notification.

Note: the csv file cannot be empty in * areas, like session_id, course_code.
#81


api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -28,6 +28,27 @@ def create
       end
     end
 
+    # POST /positions/import
+    def import
+      @position = Position.new(position_params)
```
chrislessard, line 33: Are you absolutely certain that position_params is well formed, and that the .new() method won't throw an error ? If something goes wrong, the client could be left there hanging. A try/catch block might be a good idea here. What do you think ?

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -28,6 +28,27 @@ def create
...
+
+      if @position.valid?
+        if @position.save
+          render json: @position, status: :created
+        else
+          render json: @position.errors, status: :unprocessable_entity
+        end
+      else
+        @already_exists = Position.where('course_code = ? AND session_id = ?', params[:course_code], params[:session_id])
```
chrislessard: Two things feel funny to me here. First of all, positions get validated on three criteria: 

1. the presence of a `course_code`, `openings` and `session` 
2. the numericality of `openings` 
3. the uniqueness of `course_code`

Your code is only checking whether a position already exists based on course code and session_id, which also seems funny since uniqueness on course_code suggests that it the only thing you should need to check. This code is not robust against situations where .valid? might be false for other reasons, i.e. the lack of numericality, or the lack of certain fields. You definitely need to handle these cases as well.

Second, and to be fair i'm not sure if this is really beyond the scope of your PR, you're making use of a pattern whereby you first check if a position is invalid because uniqueness is violated, then use this situation to enact an update. I feel like there should be a better way to check if an update is necessary, but I'm not an expert. @dingtaoliu could you please weigh in ?

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -28,6 +28,27 @@ def create
...
+        if @already_exists.present?
+          @already_exists.update(position_params)
+          render json: @already_exists
```
chrislessard: This is a bit strange. Isn't `@already_exists` a boolean ? I would prefer a response closer to what you have above, with `render json: @position, status: :created` if possible.

chrislessard (changes requested):
This is a great push, but it's not quite at the point we need it to be. You can find details in my comments. 

Can you please refactor your commit names and message bodies to be in line with our commit standard ? Its in the README if you need it.

api/app/controllers/api/v1/positions_controller.rb
Yifang-Zhang: That makes sense to me, so basically I need another validation before calling update in the duplicate course_code and session id case. I wonder for the update call, you also need to check on which condition you are allowed to update, right? Say if you are failing to satisfy some requirements when you make an update call, that should be automatically caught by the rails validation. If I can use the same mechanism to check before it really updates the data, then I don't need to add the validation here.

api/app/controllers/api/v1/positions_controller.rb
Yifang-Zhang: Sorry for the misleading name. The @already_exists is just a try to find the position data record according to that where condition. If it finds it, it would be a position object, if not, then just null and fails the .present call. I would try to find a better name for that.

chrislessard: Ah I see, my bad I should have understood that. If a better name comes to you i'd say go for it, but its ok to leave otherwise.

api/app/controllers/api/v1/positions_controller.rb
chrislessard: Yes, sounds like you've got it right. There are a few ways to enforce it, it might need to be mechanical if Rails doesn't provide anything convenient for us to use here. 

Interesting point about update failing. I would still prefer a more mechanical approach to error checking so that we can have a bit more flexibility in the errors we return or how we choose to handle specific cases. Let me know if you need any help moving forward with this !

api/app/controllers/api/v1/positions_controller.rb
Yifang-Zhang: I think even if position_params are corrupted, .new() method will still proceed and we will have a few error messages back upon validation. Locally I test several cases like passing null or "dfasdfdasf" to the position_params, and the rails will return a meaningful message like "session": [  "must exist", "can't be blank"]. So I think we are safe here without try/catch.

Yifang-Zhang: It's a good catch, as I thought about before. I searched online and didn't find anyone using try/catch, maybe it's one of good things from rails validation.

chrislessard:
@Yifang-Zhang this is nearly complete - you still have one commit `[Backend] Add new route for uploading csv file at frontend (fc4bc9a)` that needs to be reformated.

Yifang-Zhang:
Ok. It's reformated now.

chrislessard:
@Yifang-Zhang its still a bit off. `Feature: Add new route for uploading csv file at frontend` should read `feat(api): add new route for uploading csv file at frontend`

PR #85, opened 2019-03-04
Tapp 66: Create an application form component which renders under the route /application

dingtaoliu:
This form allows applicants to enter basic personal information and stores it in a persisted redux store.  Further features include saving the form information to the backend and creating a course selection interface.

### DIFF
--- a/frontend/src/modules/application/components/ApplicationForm.jsx
+++ b/frontend/src/modules/application/components/ApplicationForm.jsx
@@ -1,0 +1,123 @@
+import React from "react"
+import { connect } from "react-redux"
+import { updateField } from "../actions"
+
+import { personalInformationFields, 
+    currentProgramInformationFields, 
+    currentStatusFields, 
+    customQuestions } from "./FormFields"
+
+const DefaultInput = ({ label, curValue, onChange, required }) => <div>
+    <label style={{ width: "20vw" }}>{required ? label + ':' : label}</label>
+    <input style={{ width: "30vw" }} value={curValue} onChange={onChange} required={required} />
+</div>
+
+const TextboxInput = ({ label, curValue, onChange, required }) => <div>
+    <label style={{ width: "20vh" }}>{required ? label + ':' : label}</label>
+    <textarea style={{ width: "30vw" }} 
+              value={curValue} 
+              onChange={onChange} 
+              required={required}/>
+</div>
+
+const BinaryRadioInput = ({ label, curValue, onChange, required }) => <div>
+    <label style={{ width: "20vw" }}>{required ? label + ':' : label}</label>
+    <label>
+        <input type="radio" value="true" checked={curValue === "true"} onChange={onChange}/>
+        Yes
+    </label>
+    <label>
+        <input type="radio" value="false" checked={curValue === "false"} onChange={onChange}/>
+        No
+    </label>
+</div>
+
+class ApplicationForm extends React.Component {
+
+    // dispatches a redux action to update the store
+    handleChange = field_name => event => {
+        this.props.updateField({ "key": field_name,
+                                "value": event.target.value})
+    }
+
+    // TODO TAPP-67: used for submitting the form and saving it in the backend
+    handleSubmit = (e) => {
+        e.preventDefault()
+    }
+
+    // Renders a simple field component
+    render_field_type = ({ field_name, label, required, type }) => {
+        switch(type) {
+            case "textbox": 
+                return <TextboxInput
+                    curValue={this.props.fields[field_name]} 
+                    key={label} 
+                    label={label} 
+                    field_name={field_name} 
+                    required={required}
+                    onChange={this.handleChange(field_name)} 
+                />
+            case "binary":
+                if (field_name !== "grad_student" && this.props.fields["grad_student"] !== "false") {
+                    return
+                }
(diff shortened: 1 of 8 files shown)


frontend/src/modules/application/components/ApplicationForm.jsx
```diff
@@ -1,11 +1,19 @@
...
+    currentStatusFields, 
+    customQuestions } from "./FormFields"
+
+// Default input field component
 const DefaultInput = ({ label, curValue, onChange, required }) => <div>
     <label style={{ width: "20vw" }}>{required ? label + ':' : label}</label>
     <input style={{ width: "30vw" }} value={curValue} onChange={onChange} required={required} />
 </div>
 
+// Textbox input field component
```
chrislessard: I think the name of the const is enough to explain this, personally. The comment feels a bit superfluous to me.

frontend/src/modules/application/components/ApplicationForm.jsx
```diff
@@ -14,64 +22,65 @@ const TextboxInput = ({ label, curValue, onChange, required }) => <div>
               required={required}/>
 </div>
 
-const BinaryRadioInput = ({ label, curValue,required }) => <div>
+// Binary radio input field component
```
chrislessard: Same as above.

frontend/src/modules/application/components/ApplicationForm.jsx
```diff
@@ -14,64 +22,65 @@ const TextboxInput = ({ label, curValue, onChange, required }) => <div>
...
-
-const allFields = [].concat.apply([], 
-    [personalInformationFields, 
-    currentProgramInformationFields,
-    currentStatusFields, 
-    customQuestions])
-
-const initialState = allFields.reduce((acc, cur) => ({ ...acc, [cur.value]: "" }), {})
-
+// Application form component that holds multiple individual simple components defined earlier
```
chrislessard: I feel like this comment is trying to explain how the framework functions, which isn't really our responsibility. Not sure it's really necessary to keep.

frontend/src/modules/application/components/ApplicationForm.jsx
```diff
@@ -14,64 +22,65 @@ const TextboxInput = ({ label, curValue, onChange, required }) => <div>
...
-    state = { ...initialState }
-
-    handleChange = prop => event => this.setState({ [prop]: event.target.value })
+    // dispatches a redux action to update the store
+    handleChange = field_name => event => {
+        this.props.updateField({ "key": field_name,
+                                "value": event.target.value})
+    }
 
+    // TODO: used for submitting the form and saving it in the backend
```
chrislessard: Im ok with leaving a todo in the comments, but can we make sure that an associated issue gets created for it ? That's easier for the team to track imo.

frontend/src/modules/application/components/ApplicationForm.jsx
```diff
@@ -14,64 +22,65 @@ const TextboxInput = ({ label, curValue, onChange, required }) => <div>
...
     handleSubmit = (e) => {
         e.preventDefault()
     }
 
-    render_field_type = ({ value, label, required, type }) => {
-        if (type === "textbox") {
+    // Renders a simple field component
```
chrislessard, line 48: Same comment as above.

frontend/src/modules/application/components/FormFields.jsx
```diff
@@ -1,29 +1,44 @@
+// This file contains all of the fields for the application form.
```
chrislessard: I like the explanation, but please convert this to a multiline comment for readability. `{/* A JSX comment */}`

frontend/src/modules/application/components/ApplicationForm.jsx
```diff
@@ -0,0 +1,114 @@
...
+    </label>
+</div>
+
+
+
+const allFields = [].concat.apply([], 
+    [personalInformationFields, 
+    currentProgramInformationFields,
+    currentStatusFields, 
+    customQuestions])
```
chrislessard: If I understand your intentions correctly, you're trying to concatenate these various arrays into one ? You can simplify this code; concat allows you to achieve this by passing several parameters, i.e. `[].concat([1], [2],[3])`

chrislessard (changes requested):
This is great work, I have a few small nitpicks that you'll find in my comments. Can you rename your first two commits, and their associated messages, to be in line with our standard ? Admittedly this is partially my fault, since I was the one who asked if you could push your work early ;) 

Theres a number of bits on the comments you've included. Personally, I think the code (namely, the function/const names) speaks for itself in many of these situations. I don't think we should be explaining how react components work, or what they are achieving when the code can already do so, we should reserve our comments for why we make particular decisions that might not be so clear to a dev.

I know we've talked about the transience of the team, so maybe keeping very explicit comments might be helpful to people who are just starting out with React. I'm open to discussion about it. 

Otherwise, great work !

chrislessard:
@dingtaoliu I've also noticed that one of our routes does not appear to be working under your PR. I've attached a screenshot; you can reproduce by running your work and navigating to `http://localhost:8000/tapp/applicants` 

screen shot 2019-03-07 at 11 08 27 pm

chrislessard:
LGTM !

PR #86, opened 2019-03-08
Use a more specific bootstrap version in package.json

freeatnet:
This PR adjusts bootstrap version specified in package.json to be more specific. Since #84 updated the version in yarn.lock, I assume 3.4.1 or more recent is a reasonable choice for us here.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -4,7 +4,7 @@
     "proxy": "http://api:3000",
     "private": true,
     "dependencies": {
-        "bootstrap": "3",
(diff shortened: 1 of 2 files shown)


chrislessard:
Hi @freeatnet,

This change is fine with me. Can you please modify the commit header and message to be in line with the standard defined in the readme ?

freeatnet:
@chrislessard How's this?

chrislessard:
@freeatnet great, thanks for this !

Issue #68, opened 2019-02-03
[Api] update/implement the Application model


dingtaoliu:
Flesh out the Application model in rails and create api endpoints to handle CRUD operations.

chrislessard:
I'll be working on this alongside your corresponding frontend work @dingtaoliu

PR #87, opened 2019-03-10
Update the Applicant model and define a new route

chrislessard:
Hi,

This series of patch introduces a new route for POST-ing and PUT-ing applicants. It's meant to correspond to @dingtaoliu's work on the frontend. One commit also introduces a few new fields to the applicant model. 

Solves #68


dingtaoliu (approved):
Sorry for the late review, the model and controller changes look good to me, and I like that you followed the convention when using context and it in the rspecs!  Just a small nitpick, can you add some dummy values for dept_fields and program into the factory for applicants?  Thanks!

PR #89, opened 2019-03-11
Introduce eslint and refactor init script

chrislessard:
Hi,

This patch adds eslint to the project to serve as the main linter for the frontend, and sets up a new init script that will install a git precommit hook that applies linters to the project.


dingtaoliu:
Good work on the linter!

Issue #90, opened 2019-03-11
Update rails bin folder


chrislessard:
I've run into some trouble using the rails developer tools, and some online investigation has suggested that I update my binstubs by regenerating the project's bin folder. I noticed that a few things in there could use an update as well, so I figure I should make the change more permanent.

PR #91, opened 2019-03-11
Update project bin folder

chrislessard:
This patch updates the project's bin folder. 

Solves #90


dingtaoliu:
I'm good with these changes.  I did some research and found that rails and bundler will generate their own binstubs, and since we were still able to run rails command line tools after restructuring the folder, the binstubs could have been affected due to the dockerization changes to how the bundler installs the gems.

Issue #104, opened 2019-05-16
[Fix] Make TAPP work on Windows 10 


michellemtchai:
The `db` container has permission issue on Windows 10 because the postgres directory has no permission to access `tmp/data`. To fix this, Docker has to know the `PGDATA=/tmp` as its environment variable.

PR #105, opened 2019-05-16
[Fix] TAPP-104: Make TAPP work on Windows 10

michellemtchai:
[#104] adds `PGDATA=/tmp` to the file `/api/.env`


michellemtchai:
Please check if the app still runs on other OS.

dingtaoliu:
This works fine on Mac.  I'm all good with this PR since it only adds a PGDATA environment variable which sets a subdirectory for Postgres data only for the dev environment, which is unlikely to cause problems beyond making you recreate your local db.

Issue #103, opened 2019-05-16
[App] CP initial implementation


michellemtchai:
This involves creating tables, routes, and UI for contract presentment (CP).

For the table, the following tables was created:
- `assignment`
  - `applicant_id`: foreign key
  - `position_id`: foreign key
  - `hours`: int
  - `pay1`: float
  - `pay2`: float
  - `start_date`: date
  - `end_date`: date
  - `status`: `unsent` (0), `sent` (1), `accepted` (2), `rejected` (3), `withdrawn` (4)
- `offers`
  - `assignment_id`: foreign key
  - `hours`: int
  - `pay1`: float
  - `pay2`: float
  - `start_date`: date
  - `end_date`: date
  - `status`: `unsent` (0), `sent` (1), `accepted` (2), `rejected` (3), `withdrawn` (4)

For the routes, the following routes was created:
- GET `/api/v1/assignments`: returns all assignments
- GET `/api/v1/assignments/{id}`: returns assignment with `id`
- POST `/api/v1/assignment`: create a new assignment with `position_id`, `applicant_id`, `hours`, `pay1`, `pay2`
- PUT `/api/v1/assignment/{assignment_id}`: can modify the attributes in assignment
- GET `/api/v1/offers`: returns all offers
- GET `/api/v1/offers`: returns offer with `id`

For the UI, the design should follow that of the old CP and new ABC view for reference.

chrislessard:
Do we have a more detailed plan or list of to-do's for this ? It would be nice to enumerate them here with links to the relevant issues so that those interested can organize themselves more straightforwardly.

michellemtchai:
Should I move the description I made on the corresponding PR here? A lot of the plans for this portion is still on the planning stage in Slack, so nothing is really concrete as of now.

chrislessard:
Please do. It would be best that we try to keep everything documented via GH issues, as this is the easiest way for everyone to track which work is available and how things are progressing in an organized fashion.

michellemtchai:
I've moved the description over. Does that mean I continue to write in the description here when I work on the PR?

chrislessard:
No, you don't need to do that :) All I was really looking for here was a bit more specificity into what constitutes an "initial implementation", as its not immediately clear. The goal for our commits is to be as upfront as possible about the specific work involved.

PR #106, opened 2019-05-16
[APP] Tapp-103: CP initial implementation

michellemtchai:
**Please merge this after PR #105**

Please see Issue #103 for details.

### DIFF
--- a/api/app/controllers/api/v1/assignments_controller.rb
+++ b/api/app/controllers/api/v1/assignments_controller.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+module Api::V1
+  # Controller for Assignments
+	class AssignmentsController < ApplicationController
+    before_action :set_assignment, only: %i[show update]
+    # GET /assignments
+    def index
+      @assignment = Assignment.order(:id)
+
+      render json: @assignment
+    end
+
+    # GET /assignments/1
+    def show
+      render json: @assignment
+    end
+
+    # POST /assignments
+    def create
+      position = Position.find(params[:position_id])
+      applicant = Applicant.find(params[:applicant_id])
+      params[:start_date] = position[:start_date]
+      params[:end_date] = position[:end_date]
+      params[:status] = 0
+      @assignment = Assignment.new(assignment_params)
+      if @assignment.save
+        render json: @assignment, status: :created
+      else
+        render json: @assignment.errors, status: :unprocessable_entity
+      end
+    end
+
+    # PATCH/PUT /assignments/1
+    def update
+      updated_at = @assignment[:updated_at]
+      if @assignment.update(update_assignment_params)
+        # creates a copy of assignment if changed are made after being sent
+        if @assignment[:status] > 0
+          log_offer
+        end
+        render json: @assignment
+      else
+        render json: @assignment.errors, status: :unprocessable_entity
+      end
+    end
+
+   	private
+    def set_assignment
+      @assignment = Assignment.find(params[:id])
+    end
+
+   	def assignment_params
+      params.permit(
+        :id,
+        :applicant_id,
+        :position_id,
+        :hours,
+        :pay1,  
+        :pay2,
+        :status,
+        :start_date,
+        :end_date,
+      )
+    end
+    def update_assignment_params
+      params.permit(
+        :hours,
+        :pay1,  
+        :pay2,
+        :start_date,
+        :end_date,
+      )
+    end
+
+    def log_offer
+      offer = Offer.new(
+        assignment_id: @assignment[:id],
+        hours: @assignment[:hours],
+        pay1: @assignment[:pay1],  
+        pay2: @assignment[:pay2],
+        status: @assignment[:status],
+        start_date: @assignment[:start_date],
+        end_date: @assignment[:end_date],
+      )
+      if !offer.save
+        render json: offer.errors, status: :unprocessable_entity
+      end
+    end
+	end
+end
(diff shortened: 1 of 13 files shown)


api/db/migrate/20190516205625_add_columns_to_assignment.rb
```diff
@@ -0,0 +1,10 @@
+class AddColumnsToAssignment < ActiveRecord::Migration[5.2]
+  def change
+    add_column :assignments, :hours, :integer
+    add_column :assignments, :pay1, :float
+    add_column :assignments, :pay2, :float
```
chrislessard, line 5: It's unclear to me what is meant by "pay1" and "pay2". Can this be clarified with a comment or change of names ?

api/db/migrate/20190516205813_add_columns_to_offer.rb
```diff
@@ -0,0 +1,10 @@
+class AddColumnsToOffer < ActiveRecord::Migration[5.2]
+  def change
+    add_column :offers, :hours, :integer
+    add_column :offers, :pay1, :float
+    add_column :offers, :pay2, :float
```
chrislessard, line 5: Same as above.

api/db/schema.rb
```diff
@@ -36,6 +36,21 @@
...
   end
 
+  create_table "assignments", force: :cascade do |t|
+    t.bigint "applicants_id"
+    t.bigint "positions_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.integer "hours"
+    t.float "pay1"
+    t.float "pay2"
```
chrislessard, line 46: Same as above.

api/db/schema.rb
```diff
@@ -53,6 +68,23 @@
...
 
+  create_table "offers", force: :cascade do |t|
+    t.bigint "applicants_id"
+    t.bigint "positions_id"
+    t.bigint "assignments_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.integer "hours"
+    t.float "pay1"
+    t.float "pay2"
```
chrislessard, line 78: Same as above.

api/app/controllers/api/v1/assignments_controller.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module Api::V1
+  # Controller for Assignments
+	class AssignmentsController < ApplicationController
```
chrislessard: Please fix this formatting.

api/app/assets/javascripts/api/v1/assignments.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
```
chrislessard: Why are we generating .coffee files ? Are there plans to use this ? I would prefer these to be dropped these if not.

chrislessard (changes requested):
Hi Michelle, 

Thanks for all this great work ! I do have a few other nits : 

1. Please follow the commit format mentioned in the project README. Yours aren't far off, but we try to indicate both the nature of the commit and the affected region in the header, such as in "fix(docker): add PGDATA to .env" to make the commit long more consistent and clear. 

2. This is creeping towards the upper bound for PR size, as it becomes harder to review everything thoroughly in a reasonable time frame. IMO it would have been ok to keep the DB stuff and the route/code/test boilerplate separate, especially since DB work tends to be more contested.

api/db/migrate/20190516205625_add_columns_to_assignment.rb
michellemtchai: Actually, I plan to remove these attributes. Even though it was discussed during the meeting on Thursday that we should use them, there's apparently a plan to move them to a new table called `fund_source`. I'm kind of waiting for a final decision in case there's anymore changes...

api/app/controllers/api/v1/assignments_controller.rb
michellemtchai: Do you mean formatting as in the tabbing?

api/app/assets/javascripts/api/v1/assignments.coffee
michellemtchai: These were generated when I made the models in rails. Do you have a list of files that should be removed if I generate model through the rails command? Or is it preferred that I make these files manually?

michellemtchai:
I've rewritten the commit messages. Does this look okay now?

api/app/assets/javascripts/api/v1/assignments.coffee
michellemtchai: I've removed the non-controller files. Does it look okay now?

api/db/migrate/20190516205625_add_columns_to_assignment.rb
chrislessard: I see. Is there any harm in removing them now since they'll be find-able under version control anyways ? I would be tempted to remove them for now and keep a note somewhere when fund_source gets created. We can always put them back.

api/app/assets/javascripts/api/v1/assignments.coffee
chrislessard: This looks good to me !

api/app/controllers/api/v1/assignments_controller.rb
chrislessard: Yes !

chrislessard:
@michellemtchai 

Hi Michelle, 

Sorry for the delay. This all looks great to me 🙂Thanks for all your hard work !

chrislessard (approved):

PR #107, opened 2019-05-17
Upgrade and add libraries

siefkenj:
Upgrade react to 16.8.5 so hooks can be used; add `wicked_pdf` so that html templates can be rendered to pdf automatically; added the liquid template templating engine; rename the folder where database information is persisted (it was  previously named `tmp/db`, but it is not a temporary directory!).


michellemtchai:
I'm getting this error when I try to run `docker-compose up`:
image

siefkenj:
What happens if you completely reclone the branch? Do you still get the same error?

George334456 (commented):
We should run a regression on the new TAPP port on dev/QA before we merge into Master. Everything else looks good to me.

siefkenj:
@michellemtchai Did you run `docker-compose build` before `up`?

michellemtchai:
@siefkenj I used the command you mentioned. Everything looks fine, aside from the frontend code linking to the port.

dingtaoliu:
Everything works on my side as well, I am able to start the containers and tested that the frontend and api are communicating properly through the docker network.

@michellemtchai could you explain what you meant by frontend code linking to the port?

michellemtchai:
@dingtaoliu Sorry, that was a mistake. I thought the routes wouldn't work on the frontend because I remember seeing `fetch` that included the whole url (host+port+route). But apparently, I was remembering the wrong app, so there isn't actually any issue. Please just ignore my other statement.

siefkenj:
Sounds like this is ready to merge. Would someone do the honors?

michellemtchai:
@siefkenj You might want to talk to @George334456. He mentioned there were things that had be merged before this PR(?) I'm not sure what the details are, though.

thezane:
I tested the PR with the latest version of react-boostrap (1.0.0-beta.9).   I'm getting the following warning when building the frontend:

warning react-bootstrap > react-overlays > react-popper > create-react-context > fbjs > core-js@1.2.7: core-js@<2.6.8 is no longer maintained. Please, upgrade to core-js@3 or at least to actual version of core-js@2.

Right now, the frontend isn't displaying with the upgrade to react-boostrap.  The only error msg is

Proxy error: Could not proxy request /sockjs-node/140/h2df1b2d/websocket from localhost:8000 to http://api:3000 ... (ENOTFOUND).

The above error msg is also present before the upgrade to react-boostrap but the frontend displayed properly.  This leads me to conclude the core-js versioning is the problem.  Should we upgrade it?

George334456:
Could it be because we've changed ports?

PR #108, opened 2019-05-20
TAPP-101: Added more stuff to Dockerfile

George334456:


docker-compose.prod.yml
```diff
@@ -0,0 +1,28 @@
...
+      - 'db'
+    container_name: 'tapp_api'
+    command: bundle exec rails s -p 3000 -b "0.0.0.0"
+    build: ./api
+    ports:
+      - '3000:3000'
+    env_file:
+      - './api/.env'
+  frontend:
+    container_name: 'tapp_fe'
```
dingtaoliu: The frontend container does not need to run in production because the frontend is served by the rails app.

docker-compose.prod.yml
```diff
@@ -0,0 +1,28 @@
...
+    command: yarn start
+    build: ./frontend
+    ports:
+      - '8000:8000'
+    volumes:
+      - './frontend:/app'
```
dingtaoliu: We don't need the majority of the stuff here because it is just repeating what is already in the base compose file.  In the production file, we should be setting special keys such as restart to always.

docker-compose.yml
```diff
@@ -19,6 +21,7 @@ services:
     env_file:
       - './api/.env'
```
dingtaoliu, line 20: We need to make two environment files that this base compose file will read from, one for dev and one for prod.  I can add this in the original branch.

docker-compose.yml
```diff
@@ -19,6 +21,7 @@ services:
...
   frontend:
+    container_name: 'tapp_fe'
```
dingtaoliu: The frontend container should only run in the dev compose file because the container is an instance of webpack-dev-server.

dingtaoliu (changes requested):

api/credentials.yml
```diff
@@ -0,0 +1,6 @@
+# aws:
+#   access_key_id: 123
+#   secret_access_key: 345
+
+# Used as the base secret for all MessageVerifiers in Rails, including the one protecting cookies.
+secret_key_base: 5dd936603469858c82acbdc979497918584be11fb6b8d9ae83cc5f507dbbe1e99869b6b50db8f2267210592608febc1e1d088843d24ba9cb8dd16499dd2fe2d8
```
siefkenj, line 6: Is it safe for this to be here?

frontend/Dockerfile
```diff
@@ -3,10 +3,9 @@ FROM node:8.15-alpine
...
 ENV PATH=$PATH:/node_modules/.bin
 
-COPY package.json .
-
-RUN yarn
-
 RUN mkdir /app
 WORKDIR /app
 
+COPY . ./
```
siefkenj, line 9: What does this do?

Issue #111, opened 2019-05-23
[Database] Remove pay1 and pay2 from Assignments and Offers


michellemtchai:
It's been decided in this week's meeting that the `pay1` and `pay2` attributes will be moved to a new table for keeping track of the rates in the assignments and offers. So, these attributes will be removed as unused attributes.

PR #113, opened 2019-05-24
[Database] TAPP-111: Remove pay1 and pay2 from Assignments and Offers

michellemtchai:
Please refer to issue #111

### DIFF
--- a/api/db/schema.rb
+++ b/api/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_05_24_054519) do
+ActiveRecord::Schema.define(version: 2019_05_24_061229) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -42,8 +42,6 @@ ActiveRecord::Schema.define(version: 2019_05_24_054519) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.integer "hours"
-    t.float "pay1"
-    t.float "pay2"
     t.date "start_date"
     t.date "end_date"
     t.integer "status"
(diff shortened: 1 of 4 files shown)


dingtaoliu:
There is a merge conflict with the schema file because I just merged in a pr that included migrations which changed the schema version.  It seems like we will have this problem every time there are parallel branches that modify the schema, should we just remove the schema file from version control?

michellemtchai:
I think we should if `schema.rb` can generated by rails instead of just be append to by it. Also, I noticed that when you switch from a branch that has deleted table a to a branch that still contain table a, table a doesn't come back on the second branch. This may become a bigger issue later on. Should I make another issue for that?

michellemtchai:
I've fixed the conflict.

michellemtchai:
I've fixed the conflict.

dingtaoliu:
For the issue that you brought up, the problem is that the migrations are out of sync between the two branches.  I think we can fix this without resetting the entire database by dropping the 'schema_migrations' table that keeps track of migrations that have been run.  After this, we can run db:migrate again to get the db to the state we want it to be in a particular branch.

api/db/schema.rb
```diff
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+<<<<<<< HEAD
```
George334456: Is this supposed to be here?

George334456 (changes requested):
I think you left something inside, unless I'm looking at the wrong thing.

api/db/schema.rb
michellemtchai: No, it's not supposed to be here. Great catch!

michellemtchai:
I've removed the leftover line you caught.

George334456 (approved):
Good stuff

Issue #109, opened 2019-05-22
[Backend] Remove unused routes in Rails


michellemtchai:
There are a lot of routes that are unused in Rails such as:
GET `/api/v1/applicants/new`
GET `/api/v1/applicants/edit`
They need to be removed.

PR #114, opened 2019-05-24
[Backend] TAPP-109: Remove unused routes in Rails

michellemtchai:
Please refer to issue #109 for more details.


dingtaoliu (approved):
Good catch that there are no controller methods for these routes

Issue #116, opened 2019-05-24
[Frontend] create shared file-related functions for components


michellemtchai:
This app will likely have features involving uploading and downloading files, which will need to work across different components. To prevent repetitive code, a folder `lib` is created for the `/frontend`, which will serve as the location, where shared functions go.

PR #117, opened 2019-05-24
[Frontend] TAPP-116: create shared file-related functions for components

michellemtchai:
Please refer to issue #116 for details.

Tested in the following browsers:
- [x] Firefox
- [x] Chrome
- [x] Edge
- [ ] IE  `// needs polyfill for fetch, promise, arrow function`
- [ ] Safari

As a starter, a file called `fileManager.js` is made for action involving files, such as:
 - `readFile`: a function that takes an input file and reads it as either text or JSON
 - `downloadFile`: a function that fetches a blob given at a certain route and triggers a download file response

Note: both of these functions output a JSON, using the second function argument, with the following format:
```
{
  success: boolean,
  content: string OR JSON, // content of file OR error message
}
```
Furthermore, both functions takes in a function as the second parameter. The function is assumed to take in only one argument. This single argument should be the aforementioned JSON.

Here is an example on the use of `readFile()` in HTML:
```
<input type='file' onchange='readFile(this, console.log)' />
```
Please note that the use of `onchange` is very important for getting the content of the file when you click on the upload button.

Here is an example on the use of `downloadFile()` in HTML:
```
<button onclick='downloadFile("https://www.wikipedia.org/portal/wikipedia.org/assets/img/sprite-556af1a5.svg", console.log)'>Download</button>
```
For a quick test of these functions, you can make a HTML like this:
```
<!DOCTYPE html>
<html>
<head>
	<script type="text/javascript" src="./fileManager.js"></script>
</head>
<body>
	<p>Upload Test</p>
	<input type='file' onchange='readFile(this, console.log)' />
	<button onclick='downloadFile("https://www.wikipedia.org/portal/wikipedia.org/assets/img/sprite-556af1a5.svg", console.log)'>Download</button>
</body>
</html>
```
Note: You will need to put this file into `/frontend/src/lib/` for it to test it on your browser. You will also need to comment out the `export` before the functions in `fileManager.js` because it's an independent JS file.

### DIFF
--- a/frontend/src/lib/fileManager.js
+++ b/frontend/src/lib/fileManager.js
@@ -1,0 +1,81 @@
+export const readFile = (component, loadDataFunc) => {
+    let files = component.files;
+    if (files.length > 0) {
+        const reader = new FileReader();
+        let importFunc = importChoices(files[0].name, loadDataFunc);
+        reader.readAsText(files[0]);
+        reader.onload = event => importFunc(event.target.result);
+    }
+    component.value = '';
+}
+
+export const downloadFile = async (route, loadMessage) => {
+    try {
+        const res = await fetch(route);
+        if (res.status === 200) {
+            const blob = await res.blob();
+            let filename = getFilename(res);
+            if (window.navigator && window.navigator.msSaveOrOpenBlob) {
+                window.navigator.msSaveOrOpenBlob(blob, filename); //special case for Edge & IE
+            }
+            else{
+                let url = URL.createObjectURL(blob),
+                    a = document.createElement('a');
+                a.href = url;
+                a.download = filename;
+                a.target = "_self" ; //required in FF
+                a.style.display = 'none';
+                document.body.appendChild(a); //required in FF
+                a.click();
+                URL.revokeObjectURL(url);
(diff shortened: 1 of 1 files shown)


michellemtchai:
Please check if this code works on Safari if you have a Mac.

frontend/src/lib/fileManager.js
```diff
@@ -0,0 +1,83 @@
+export const readFile = (component, loadDataFunc) => {
+    let files = component.files;
+    if (files.length > 0) {
+        const reader = new FileReader();
+        let importFunc = importChoices(files[0].name, loadDataFunc);
+        reader.readAsText(files[0]);
+        reader.onload = event => importFunc(event.target.result);
+        component.value = '';
+    } else {
```
dingtaoliu: We can pull component.value out of the if block and get rid of the else to be more concise

dingtaoliu (changes requested):
Just one small change, everything else looks good!  I like how generalized these functions are, good work!

frontend/src/lib/fileManager.js
michellemtchai: I've fixed the redundancy. Thanks.

PR #119, opened 2019-05-28
feat(app): added deployment script for production

dingtaoliu:
This commit makes several changes to the docker setup to create a deployment process for production.  The docker compose files are now split into four files:
	- docker-compose.yml: defines the api and db services
	- docker-compose.dev.yml: defines development specific docker behaviour
	- docker-compose.prod.yml: defines production specific docker behaviour
	- docker-compose.frontend.yml: defines a container used only to build optimized webpack bundles

Environment variables for rails and postgres can be found in prod.env and dev.env.

To deploy, pull the most recent master branch into the server and run the deploy script.  This will spin up the rails and postgres containers in detached mode.
To run the app locally, run the start_dev script.  This will spin up a rails, webpack, and postgres server in development mode with hot-reloading enabled.

### DIFF
--- a/deploy
+++ b/deploy
@@ -4,4 +4,3 @@ docker-compose -f docker-compose.yml -f docker-compose.prod.yml down --remove-or
 docker-compose -f docker-compose.frontend.yml up --build --force-recreate
 cp -a frontend/build/. api/public/
 docker-compose -f docker-compose.yml -f docker-compose.prod.yml up --build --force-recreate -d
-docker-compose exec api sh -c "rails db:migrate RAILS_ENV=production"
--- a/deploy
+++ b/deploy
@@ -1,0 +1,7 @@
+#!/bin/bash
+cp prod.env .env 
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml down --remove-orphans
+docker-compose -f docker-compose.frontend.yml up --build --force-recreate
+cp -a frontend/build/. api/public/
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml up --build --force-recreate -d
+docker-compose exec api sh -c "rails db:migrate RAILS_ENV=production"
--- a/start_dev
+++ b/start_dev
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+docker-compose -f docker-compose.yml -f docker-compose.dev.yml down
+cp dev.env .env 
+docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build
+docker-compose exec api sh -c "rails db:migrate RAILS_ENV=development"
+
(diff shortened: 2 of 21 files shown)


michellemtchai:
Hey, I got this problem when I tried to run `./deploy`:
image
Was I suppose to run anything else before this?

deploy
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+cp prod.env .env 
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml down --remove-orphans
+docker-compose -f docker-compose.frontend.yml up --build --force-recreate
+cp -a frontend/build/. api/public/
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml up --build --force-recreate -d
+docker-compose exec api sh -c "rails db:migrate RAILS_ENV=production"
```
michellemtchai: Actually, you can shorten this last command by calling 
`docker-compose exec api rails db:migrate RAILS_ENV=production` 
instead.

michellemtchai:
Also, when I ran `./start_dev`, I saw this warning:
image
Were these variables intentionally left as blank?

start_dev
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+docker-compose -f docker-compose.yml -f docker-compose.dev.yml down
+cp dev.env .env 
+docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build
+docker-compose exec api sh -c "rails db:migrate RAILS_ENV=development"
```
michellemtchai: When I ran this script, the last line was never called because the process on line 5 never finished.

michellemtchai:
I've ran `./deploy` after I ran `docker system prune --all`, but it seems like the `frontend/build` was never made. As a result, the `frontend` container keeps crashing. In fact, from the output, it looks like the container also got to the point where it was being built, but never showed signs of `docker-compose up`. It reaches the last line of the script just fine, though. It did the migration just fine.

Since the docker image didn't seem to be up, I ran:
```
docker-compose -f docker-compose.yml -f docker-compose.prod.yml up
```
From the message, it seem like it only managed to start up `tapp_db_1` and `tapp_api_1`. The frontend portion is completely missing. Furthermore, `localhost:3000` didn't displayed a `This page isn't working`. The problem seems to stem from the unsuccessful creating of `frontend/build`. Here's what I saw on the console:
image

I've also ran the `./start_dev` after `docker system prune --all`. It doesn't seem to have the problem of crashing containers. In fact, all three containers seems to be running properly. Though, I had to run the migration command separately. Also, I got this error once the migration warning cleared up:
image

start_dev
dingtaoliu: This is a mistake on my end, I have moved migrations to an entry point file that is run every time the container starts so we no longer have to explicitly run them with a docker command.

dingtaoliu:
For the warnings about environment variables not set, I intentionally left them blank in the dev environment because I wanted the default behaviour from rails but didn't know what the corresponding values should be.

michellemtchai (approved):

Issue #121, opened 2019-05-29
[Database] Add tables and attributes according to new schema plan


michellemtchai:
The design of the new schema follows the following diagram:
![image](https://user-images.githubusercontent.com/22383586/58597660-fdddbd80-8246-11e9-9ab7-948cd4e712dd.png)

PR #122, opened 2019-05-30
[Database] TAPP-121: Add tables and attributes according to new schema plan 

michellemtchai:
Please see issue #121 for details.


api/app/models/position.rb
```diff
@@ -13,11 +13,10 @@ class Position < ApplicationRecord
   has_many :assignments
   has_many :position_preferences
   has_many :applications, through: :position_preferences
+  has_one :position_data_for_ad
+  has_one :position_data_for_matching
 
 
-  validates_presence_of :course_code, :openings, :session
```
George334456, line 13: Is there a reason we're deleting these rails validations? I don't think we have the database validations for these either.

api/db/migrate/20190606151443_create_position_data_for_ads.rb
```diff
@@ -0,0 +1,14 @@
+class CreatePositionDataForAds < ActiveRecord::Migration[5.2]
+  def change
+    create_table :position_data_for_ads do |t|
```
George334456, line 3: Did we ever come to a conclusion on whether or not we should be on one singular position table or multiple position tables?

George334456 (commented):
Couple of comments, please respond to them.

Also, please start an issue for unit tests.

api/app/models/position.rb
michellemtchai: I deleted these validation because the columns used for the validation no longer exist. The new schema plan got rid of them, so I wasn't if the exact same constraints should apply to which attribute.

api/db/migrate/20190606151443_create_position_data_for_ads.rb
michellemtchai: @reidka mentioned in the meeting yesterday to separate position into multiple tables.

michellemtchai:
I've made a new issue #125 for the unit tests.

George334456 (approved):
Goooooooooooooooooooooooooooooooooooood stuff

Issue #123, opened 2019-06-06
[Database] Add gem to auto-generate ER diagram


michellemtchai:
`rails-erd` is a gem that can generate the ER diagram of the model without relying on external tools such as RubyMine (IDE) or other online tools. Furthermore, unlike `railroady`, which requires a separate command to generate the diagrams, this gem can be configured to auto-generate the diagram whenever `rake db:migrate` is called.

PR #124, opened 2019-06-06
Enhancement [Database]: Tapp-123: Add gem to auto-generate ER diagram

michellemtchai:
Please refer to Issue #123 for details.


api/Gemfile
```diff
@@ -61,6 +61,7 @@ group :development do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   gem 'spring-watcher-listen', '~> 2.0.0'
+  gem 'rails-erd'
```
George334456: Comment what this is for please.

michellemtchai: Do you mean comment on the file? Or do you mean comment on what it is used for? If it's the latter, I've already described the use in issue #123.

George334456: Sorry, comment in the file similar to the annotations.

michellemtchai: I've added the annotations.

michellemtchai:
I've rebase with `master` to create the output with the ER diagram of the latest schema.

dingtaoliu:
Since we are adding the generated diagram to version control, should we be generating new diagrams everytime migrations are run?

michellemtchai:
Is this about the situation when two PRs conflict or something(?) We shouldn't be changing the schema too much after PR #126. Can you elaborate on the problem with using version control on the new diagram? I'm a bit confused about the main part of your concern.

dingtaoliu:
If we version control the diagram, we want it to represent the current schema.  My concern is that if we add new migrations, would the existing diagram in the repo be outdated?  Should we generate new diagrams for each PR that involve migrations?

michellemtchai:
The main reason why I chose this gem was because the diagram is generated every time `rake db:migrate` is called. The diagram generation is automatic. There's no need to run a separate command to update the diagram each time. It does so on its own.

dingtaoliu:
That's perfect, I'm good to merge this in.

PR #133, opened 2019-06-13
Use `docker-compose.override.yml` for dev environment

siefkenj:
Docker will automatically pull in `docker-compose.yml`
	and `docker-compose.override.yml` when running `docker-compose
	up`. The default for `docker-compose up` should be to start
	the development environment.

	Also fixes a bug due to the an old `I18n` configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 TA application, assignment, and matching.
 
 ## Contributing
-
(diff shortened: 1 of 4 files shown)


George334456 (approved):
Looks good to me.

PR #141, opened 2019-06-18
Change default api port from 3000 to 3023

siefkenj:
The api is currently served on port 3000, which is the default port for create-react-app and so it will conflict with other apps being developed on the same computer. 3023 is a random port that no body else uses (to my knowledge).


George334456 (approved):
Looks good to me.

Issue #128, opened 2019-06-08
[Frontend] Add positions list to the applications component folder 


dingtaoliu:
When a student wants to apply for TA positions, they need to fill out an application form and select course preferences.  As part of a previous PR (https://github.com/uoft-tapp/tapp/pull/97), Yifang has created a modified positions list that has a drop down for preference selection.  This component currently has its own separate redux store, but we want the preference selection component and the application form to share the same redux store so that when the application form is submitted, the course preferences can be sent in the same request.  

To do this, we can:
  1) move the components in applicant positions into the application directory 
  2) move and combine the redux actions and reducer from applicant positions with the actions 
  and reducer in application
  3) connect the redux store in application to the components moved over from applicant 
  positions

You may want to change the name of the component 'PositionsApplied' to something that reflects a list of position advertisements.

PR #148, opened 2019-06-20
[Frontend] Refactor positions list to the applications

jacobljh:
Please see #128 and #145 for details

Refactor positions applied to applicants so that one application is associated with all the positions applied with preference level and the applicant's info.


frontend/src/modules/application/actions.js
```diff
@@ -1,10 +1,14 @@
...
 import { error, success  } from "react-notification-system-redux"
 import { errorProps } from "../notifications/constants"
 
 // an action generator function that returns an action object
 export const updateField = data => ({ type: UPDATE_FIELD, data })
 
+// action used by applicant form view
 export const createNewApplication = payload => async dispatch => {
     const res = await fetch("/api/v1/applicants", {
         method: "POST",
```
siefkenj: We need to add the credential sending to this request. Once Shibboleth is activated, all requests sent without credentials will fail.

frontend/src/modules/application/actions.js
```diff
@@ -24,4 +28,17 @@ export const createNewApplication = payload => async dispatch => {
...
     }
-}
\ No newline at end of file
+}
+
+// action used by applicant positions view
+export const fetchPositionsSuccess = payload => ({ type: FETCH_POSITIONS_SUCCESS, payload })
+export const fetchPositions = () => async dispatch => {
+    const res = await fetch("/api/v1/positions")
+    const data = await res.json()
```
siefkenj: This needs to be in a `try .. catch` block. `await` turns failed promises into thrown errors.

frontend/src/modules/application/actions.js
```diff
@@ -24,4 +28,17 @@ export const createNewApplication = payload => async dispatch => {
...
         }
     }
-}
\ No newline at end of file
+}
+
+// action used by applicant positions view
+export const fetchPositionsSuccess = payload => ({ type: FETCH_POSITIONS_SUCCESS, payload })
+export const fetchPositions = () => async dispatch => {
+    const res = await fetch("/api/v1/positions")
```
siefkenj: Credentials.

frontend/src/modules/application/actions.js
```diff
@@ -24,4 +28,17 @@ export const createNewApplication = payload => async dispatch => {
...
+    const data = await res.json()
+    if (res.status === 200) {
+        dispatch(fetchPositionsSuccess(data))
```
siefkenj: Wouldn't it be better to await the JSON in here after you know the fetch was a success?

frontend/src/modules/application/actions.js
```diff
@@ -33,10 +33,18 @@ export const createNewApplication = payload => async dispatch => {
...
+    } catch(err) {
+        alert(err);
+    }
     if (res.status === 200) {
-        dispatch(fetchPositionsSuccess(data))
+        try{
+            const data = await res.json()
+            dispatch(fetchPositionsSuccess(data))
+        } catch(err) {
+            alert(err);
```
siefkenj: Don't use `alert`. You should be dispatching any errors

frontend/src/modules/application/actions.js
```diff
@@ -33,10 +33,18 @@ export const createNewApplication = payload => async dispatch => {
 // action used by applicant positions view
 export const fetchPositionsSuccess = payload => ({ type: FETCH_POSITIONS_SUCCESS, payload })
 export const fetchPositions = () => async dispatch => {
-    const res = await fetch("/api/v1/positions")
-    const data = await res.json()
+    try{
```
siefkenj: Run your code through `prettier`. There should be a space before `{`

frontend/src/modules/application/actions.js
```diff
@@ -33,10 +33,18 @@ export const createNewApplication = payload => async dispatch => {
...
+        var res = await fetch("/api/v1/positions")
```
siefkenj: We need to send credentials with every `fetch` otherwise it will fail when shibboleth is active.

siefkenj (commented):
See if you can remove `erd.svg` from your PR. It gets rebuilt automatically, and it will be annoying to have merge conflicts because of it...

Issue #150, opened 2019-06-26
[Database] remove preference_level from schema


michellemtchai:
Remove the preference_level table from the database schema.

Issue #147, opened 2019-06-19
[Backend] Update schema for missing fields


George334456:

michellemtchai:
`nag_count` and `email_count` needs to be added to offers table.

PR #151, opened 2019-06-26
[Database] fix errors in schema

michellemtchai:
See details in issue #150, and #147 for more details.


jmzaleski (approved):
looks good. only issue (which Jacob and I caused) is the nag count field.

jmzaleski (approved):

Issue #146, opened 2019-06-19
[Backend] Add API


George334456:

michellemtchai:
Follows the API plans in this document: https://docs.google.com/document/d/18nNYgLbXvL2pycjWi9kIZg35XBKbdgdoOOSVn7JEpTk/edit

PR #152, opened 2019-06-26
[Backend] Add API

michellemtchai:
See issue #146 for more details.

**Please merge this PR after PR #151**

### DIFF
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -19,13 +19,16 @@ module Api::V1
             if invalid_id(Session, :session_id) then return end
             application = Application.new(application_params)
             if not application.save # does not pass Application model validation
+                application.destroy!
                 render_error(application.errors)
+                return
             end
             params[:application_id] = application[:id]
-            message = valid_applicant_matching_data
+            message = valid_applicant_matching_data(application.errors.messages)
             if not message
                 render_success(application)
             else
+                application.destroy!
                 render_error(message)
             end
         end
@@ -86,13 +89,14 @@ module Api::V1
             end
         end
 
-        def valid_applicant_matching_data
+        def valid_applicant_matching_data(errors)
             if invalid_id(Applicant, :applicant_id) then return end
             matching = ApplicantDataForMatching.new(applicant_data_for_matching_params)
             if matching.save
                 return nil
             else
-                return matching.errors
+                matching.destroy!
+                return errors.deep_merge(matching.errors.messages)
             end
         end
     end
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -10,17 +10,13 @@ module Api::V1
                 render_success(Application.order(:id))
                 return
             end
-            if invalid_primary_key(Session, :session_id)
-                return
-            end
+            if invalid_id(Session, :session_id) then return end
             render_success(applications_by_session)
         end
 
         # POST /applications
         def create
-            if invalid_primary_key(Session, :session_id)
-                return
-            end
+            if invalid_id(Session, :session_id) then return end
             application = Application.new(application_params)
             if not application.save # does not pass Application model validation
                 render_error(application.errors)
@@ -60,9 +56,7 @@ module Api::V1
         end
 
         def valid_applicant_matching_data
-            if invalid_primary_key(Applicant, :applicant_id)
-                return
-            end
+            if invalid_id(Applicant, :applicant_id) then return end
             matching = ApplicantDataForMatching.new(applicant_data_for_matching_params)
             if matching.save
                 return nil
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -30,6 +30,20 @@ module Api::V1
             end
         end
 
+        # PUT/PATCH /applications/:id
+        def update
+            application = Application.find(params[:id])
+            matching = ApplicantDataForMatching.find(application[:id])
+            application_res = application.update_attributes!(application_update_params)
+            matching_res = matching.update_attributes!(matching_data_update_params)
+            errors = application.errors.messages.deep_merge(matching.errors.messages)
+            if application_res and matching_res
+                render_success(application)
+            else
+                render_error(errors)
+            end
+        end
+
         private
         def application_params
             params.permit(
@@ -37,6 +51,7 @@ module Api::V1
                 :session_id,
             )
         end
+
         def applicant_data_for_matching_params
             params.permit(
                 :program, 
@@ -49,6 +64,22 @@ module Api::V1
             )
         end
 
+        def application_update_params
+            params.permit(
+                :comments,
+            )
+        end
+
+        def matching_data_update_params
+            params.permit(
+                :program, 
+                :department, 
+                :previous_uoft_ta_experience, 
+                :yip, 
+                :annotation,
+            )
+        end
+        
         def applications_by_session
             return Application.order(:id).select do |entry|
                 entry[:session_id] == params[:session_id].to_i
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+module Api::V1
+    # Controller for Applications
+    class ApplicationsController < ApplicationController
+
+        # GET /applications
+        def index
+            if not params.require(:session_id)
+                render json: { status: 'success', message: '', payload: Application.order(:id) }
+                return
+            end
+            if Session.exists?(id: params[:session_id])
+                render json: { status: 'success', message: '', payload: applications_by_session }
+            else
+                render json: { status: 'error', message: 'Invalid session_id', payload: {} }
+            end
+        end
+
+        # POST /applications
+        def create
+            if not Session.exists?(id: params[:session_id])
+                render json: { status: 'error', 
+                    message: 'Invalid session_id', payload: applications_by_session }
+                return
+            end
+            application = Application.new(application_params)
+            if not application.save # does not pass Application model validation
+                render json: { status: 'error', message: application.errors, payload: {} }
+            end
+            params[:application_id] = application[:id]
+            message = valid_matching
+            if not message
+                render json: { status: 'error', message: '', payload: application }
+            else
+                render json: { status: 'error', message: message, payload: {} }
+            end
+        end
+
+        private
+        def application_params
+            params.permit(
+                :comments,
+                :session_id,
+            )
+        end
+        def applicant_data_for_matching_params
+            params.permit(
+                :program, 
+                :department, 
+                :previous_uoft_ta_experience, 
+                :yip, 
+                :annotation,
+                :applicant_id,
+                :application_id,
+            )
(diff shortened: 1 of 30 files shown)


api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -1,60 +1,32 @@
...
+
+        # GET /instructors
+        def index
+            render json: { status: 'success', message: '', payload: Instructor.order(:id) }
+        end
+
+        # POST /instructors
+        def create
+            instructor = Instructor.new(instructor_params)
+            if instructor.save
```
siefkenj: What is this?

michellemtchai: Which line are you referring to?

siefkenj: `if instructor.save`

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+                :id,
+                :session_id,
+                :offer_template,
+                :position_type,
+            )
+        end
+
+        def position_templates_by_session
+            position_templates = []
+            PositionTemplate.order(:id).each do |entry|
```
siefkenj: You can change this to
```
return PositionTemplate.order(:id).select do |entry|
end
```

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -1,92 +1,83 @@
...
-      if @position.update(position_params)
-        render json: @position
-      else
-        render json: @position.errors, status: :unprocessable_entity
-      end
-    end
+        # POST /positions
+        def create
+            req_keys = [:position_code, :position_title]
+            if Session.exists?(id: params[:session_id])
```
siefkenj: I think it makes code more readable to have early returns instead of nested `if`s.

e.g.,
```
if not Session.exists?(id: params[:session_id])
    render json: { status: 'error', message: 'Invalid session_id', payload: {} }
    return
end
if not (req_keys & params.keys.map(&:to_sym) == req_keys)
    render json: { status: 'error',  message: 'Missing position_code or position_title', payload: {} }
    return
end

position = Position.new(position_params)
...
```

api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -1,61 +1,52 @@
...
-    end
+    # Controller for Instructors
+    class InstructorsController < ApplicationController
+
+        # GET /instructors
+        def index
+            if Position.exists?(id: params[:position_id])
+                render json: { status: 'success', message: '', payload: instructors_by_position }
+            else
+                render json: { status: 'error', message: 'Invalid position_id', payload: {} }
```
siefkenj: What if `:position_id` isn't provided because this is used from a different route?

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: This checks if the `Instructor` created passes all the validation for the model `Instructor`.

api/app/controllers/api/v1/position_templates_controller.rb
michellemtchai: Great idea. I'll change it. Thanks.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: I've removed the nested `if`s.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: I've made a new route `GET /api/v1/instructors` for the `index` function. Is this what you mean?

siefkenj: Yep, that's what I mean.

api/app/controllers/api/v1/instructors_controller.rb
siefkenj: `.save` seems like a really awkward name for `.passes_validation`...

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -1,92 +1,79 @@
...
-      end
-    end
+        # POST /positions
+        def create
+            req_keys = [:position_code, :position_title]
+            if not Session.exists?(id: params[:session_id])
+                render json: { status: 'error', message: 'Invalid session_id', payload: {} }
+                return
+            end
+            if (req_keys & params.keys.map(&:to_sym)) != req_keys
```
siefkenj: https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require looks to be the standard method to require particular `params`. Is there a reason you didn't use it?

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -1,92 +1,79 @@
...
+                render json: { status: 'error', 
+                    message: 'Missing position_code or position_title', payload: {} }
+                return
+            end
+            position = Position.new(position_params)
+            if position.save
```
siefkenj: I like this the other way around, too.
```
if not position.save
   ...
end
...rest of code
```

api/app/controllers/api/v1/sessions_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    class SessionsController < ApplicationController
+
+        # GET /sessions
+        def index
+            render json: { status: 'success', message: '', payload: Session.order(:id) }
+        end
+
+        # POST /sessions
+        def create
+            if not params.include?(:name)
```
siefkenj: See previous comment about required params

api/app/controllers/api/v1/instructors_controller.rb
siefkenj: Ah, I understand. `instructor.save` does the action of saving an instructor returning `true` if the save was successful. Can you add a comment to this effect for new rails people?

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: Actually, I did not know about this function. I'll change it to `require`. Thanks.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: I've move the `not position.save` to the top. I've also added a validation for `position_data_for_matching` and `position_data_for_ad`. Should I move that up as well?

api/app/controllers/api/v1/sessions_controller.rb
michellemtchai: Done.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: Okay. I'll do that.

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -1,92 +1,92 @@
...
+                return
+            end
+            if not params.require(req_keys)
+                render json: { status: 'error', 
+                    message: 'Missing position_code or position_title', payload: {} }
+                return
+            end
+            position = Position.new(position_params)
+            if not position.save # does not pass Position model validation
+                render json: { status: 'error', message: position.errors, payload: {} }
```
siefkenj: Need a `return` here

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -1,92 +1,92 @@
...
-        :start_date, :end_date
-      )
+        def valid_ad_and_matching
+            ad = PositionDataForAd.new(position_data_for_ad_params)
+            matching = PositionDataForMatching.new(position_data_for_matching_params)
+            if ad.save and matching.save
+                return nil
+            elsif ad.save and not matching.save
+                return matching.errors
+            elsif not ad.save and matching.save
```
siefkenj: Does this code re-execute `*.save` a bunch of times? Will it save items to the database multiple times??

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+
+        # POST /add_position_template
+        def create
+            if not Session.exists?(id: params[:session_id])
+                render json: { status: 'error', 
+                    message: 'Invalid session_id', payload: position_templates_by_session }
+                return
+            end
+            if not params.require(:offer_template)
```
siefkenj: `params.require` documentation says it raises an error if the requirement is not met, so I don't think this is the right way to use it.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: Great catch. Thanks. I'll fix that.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: I don't know if this re-execute a bunch of times, so I'll just save the outcome to a local variable.

api/app/controllers/api/v1/position_templates_controller.rb
michellemtchai: Okay. I've gotten rid of this and wrote a rescue from `ParameterMissing` instead.

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+                return
+            end
+            application = Application.new(application_params)
+            if not application.save # does not pass Application model validation
+                render json: { status: 'error', message: application.errors, payload: {} }
+            end
+            params[:application_id] = application[:id]
+            message = valid_matching
+            if not message
+                render json: { status: 'error', message: '', payload: application }
```
siefkenj: should this be `status: 'success'`?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+                render json: { status: 'error', 
+                    message: 'Invalid session_id', payload: applications_by_session }
+                return
+            end
+            application = Application.new(application_params)
+            if not application.save # does not pass Application model validation
+                render json: { status: 'error', message: application.errors, payload: {} }
+            end
+            params[:application_id] = application[:id]
+            message = valid_matching
```
siefkenj: This is a confusing name...can you come up with something better?

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Yes. It should be. Thanks for catching that.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Is `valid_applicant_matching_data` a good alternative? Or should I come up with something else?

michellemtchai:
I've added the routes:
```
GET /sessions/:session_id/applicants
GET /assignments/:assignment_id/wage_chunks
```

siefkenj:
Is this ready to merge?

michellemtchai:
I guess it is ready to merge if you don't see any glaringly missing route/error.

siefkenj:
I'll go ahead and merge. It looks mostly good. I am sure some bugs will be revealed when we have seed data and unit tests, but since we don't have a full product yet, let's push forward for rapid progress!

PR #158, opened 2019-06-27
[frontend] Added `apiGET` and `apiPOST` utility functions

siefkenj:
All GET and POST requests can (and should) be accessed through `apiGET` and `apiPOST` utility functions in `src/libs/apiUtils.js`. Most bare `fetch()` instances have been replaced with these calls.


frontend/src/libs/apiUtils.js
```diff
@@ -0,0 +1,107 @@
...
+        super(errorMessage);
+        Error.captureStackTrace(this, ApiFetchError);
+        this.response = resp;
+        this.status = resp.status;
+    }
+}
+
+// Ensure that `path` starts with a `/`
+function _ensurePath(path) {
+    return path.starsWith("/") ? path : "/" + path;
```
reidka: should be 'path.startsWith' ?

reidka (commented):
Aside from the typo, the fetch calls seem to work.

frontend/src/libs/apiUtils.js
siefkenj: Sure should! Soon there'll actually be an api to query against ;-)

Issue #166, opened 2019-06-29
[Backend] Create seed data 


michellemtchai:
- seed the database using the API routes
- create mock data

PR #167, opened 2019-06-29
[Backend] Seed data

michellemtchai:
Please see issue #166 for more detail.

**Please merge this PR after PR #182**

To test this PR:
1) `docker-compose up --build`
2) `docker-compose exec api rake db:drop db:create db:migrate`
3) `docker-compose exec api rake db:seed`: There should be a log that tells you if the seeding succeeded or not
4) use browser to check db state via GET routes
6) change the content in `api/db/seed/mock_data.json`
5) `docker-compose exec api rake db:seed`: The db should update to match the changed `mock_data`
7) use browser to check db state via GET routes

**Note:** The last line in `seed.rb` generates fake data. You can comment it out if you only want to deal with the data from the `mock_data.json`.


siefkenj:
Can you rebase this branch now that PR #152 has been merged?

michellemtchai:
Okay. I'll do that.

api/app/helpers/seeds_handler.rb
```diff
@@ -0,0 +1,231 @@
...
+        rescue Errno::ENOENT => e
+            log.push(e)
+            return [], log
+        rescue JSON::ParserError 
+            log.push("#{file_dir} is not a valid JSON file.")
+            return [], log
+        end
+    end
+
+    def insert_data(seed)
```
siefkenj: I don't understand why you need all these functions...all your data should be seeded by making calls to the api with `request`. You should never need to peek inside any controllers...

api/db/seeds.rb
```diff
@@ -3,65 +3,150 @@
...
+            }
+        ]
+    },
+    {
+        table: ReportingTag,
+        create: '/wage_chunks/:wage_chunk_id/add_reporting_tag',
+        update: '/reporting_tags/:id',
+        file: 'reporting_tag.json',
+        unique: [:wage_chunk_id, :position_id],
+        foreign_keys: [
```
siefkenj: Why do you need to know anything about foreign keys?

api/db/seed/wage_chunk.json
```diff
@@ -0,0 +1,65 @@
+[
+    {
+        "hours": 54,
+        "rate": 42.25,
+        "start_date": "2018-9-10",
+        "end_date": "2018-12-10",
+        "assignment_id": 0
```
siefkenj: The ids shouldn't be hardcoded...they should be read from the return value of the api call and synthesized on the spot. Eventually the controller will generate randomized ids, and then this seeing will fail.

michellemtchai: Actually, this only looks like `id`s it's actually the index of entry from `assignment.json`.

siefkenj: Oh. That's pretty confusing then...If it's separate linking data, put it in a separate JSON object. That way the seed data looks like seed data.

michellemtchai: Do you mean like in a separate file? I'm having trouble imagining how that would work.

api/db/seeds.rb
michellemtchai: I actually used it to get the actually id value like in the discussion below.

api/app/helpers/seeds_handler.rb
michellemtchai: So, we're chaining request outcome to make more requests?

api/db/seed/wage_chunk.json
siefkenj: Yes. Maybe `wage_chunk_links.json`. You're essentially creating a DSL for batch-processing with the API right now...

michellemtchai: Is the DSL comparison what I'm doing right now or what I should/should not be attempting? Also, the seeding should be done via chaining API in order(?)

siefkenj: @michellemtchai Exactly! You're chaining the API calls. If you want to create a mini DSL to assist you with chaining a bunch of calls in a particular order, that's fine. If you want to hard code it, that's also fine.

Whatever sequence of calls you code should also be able to be ported more-or-less directly to Javascript for client-side unit tests.

michellemtchai: Okay. I think I get it now. Thanks.

api/app/controllers/api/v1/applicants_controller.rb
```diff
@@ -34,6 +39,17 @@ def update
                 render_error(applicant.errors)
             end
         end
+
+        # /applicants/delete or /applicants/:id/delete
+        def delete
+            params.require(:id)
+            applicant = Applicant.find(params[:id])
+            if session.destroy!
```
dingtaoliu: Should this be `applicant.destroy!`?

michellemtchai: Yes. This is a mistake that carried over from PR #183. Once, I fix that PR, I'll rebase this branch off that PR, again.

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -16,7 +16,14 @@ def index
...
         # POST /applications
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Application.exists?(params[:id])
+                update and return
+            end
+
+            params.require(:applicant_id)
             if invalid_id(Session, :session_id) then return end
+            if invalid_id(Applicant, :applicant_id) then return end
```
dingtaoliu: you can write this as `return if invalid_id(Applicant, :applicant_id)` so its a little easier to read.

michellemtchai: Okay. I'll do that. Though, do you mind commenting on PR #183 instead, if you're going to review the controller.

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -84,13 +107,21 @@ def matching_data_update_params
...
-            return Application.order(:id).select do |entry|
+            return all_applications.select do |entry|
                 entry[:session_id] == params[:session_id].to_i
             end
         end
 
+        def all_applications
+            exclusion = [:id, :created_at, :updated_at, :application_id]
+            return Application.order(:id).map do |entry|
+                matching = ApplicantDataForMatching.find_by(application_id: entry[:id])
```
dingtaoliu: it might be better to use a join here so we don't need to make a db call for each Application, but we don't need to worry about it for now.

api/app/controllers/api/v1/assignments_controller.rb
```diff
@@ -44,6 +49,17 @@ def update
             end
         end
 
+        # /assignments/delete or /assignments/:id/delete
+        def delete
+            params.require(:id)
+            assignment = Assignment.find(params[:id])
+            if session.destroy!
```
dingtaoliu: should this be `assignment.destroy!`?

api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -14,30 +14,44 @@ def index
...
             else
                 render_error(instructor.errors)
             end
         end
 
+        # /instructors/delete or /instructors/:id/delete
+        def delete
+            params.require(:id)
+            instructor = Instructor.find(params[:id])
+            if session.destroy!
```
dingtaoliu: this might be a type as well

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -38,6 +48,17 @@ def update
             end
         end
 
+        # /position_templates/delete or /position_templates/:id/delete
+        def delete
+            params.require(:id)
+            position_template = PostionTemplate.find(params[:id])
+            if session.destroy!
```
dingtaoliu: typo

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -129,11 +145,35 @@ def matching_update_params
...
+                matching = json(matching, except: exclusion)
+                ad = PositionDataForAd.find_by(position_id: entry[:id])
+                combined = json(ad, include: matching, except: exclusion)
+                combined = json(combined, include: {instructor_ids: entry.instructor_ids})
+                json(entry, include: combined, except: [:instructors])
+            end
+        end
+
+        def update_instructors_ids(position)
+            if params[:instructor_ids]
```
dingtaoliu: you should use a guard clause for these kinds of nested conditionals, `return if params[:instructor_ids].blank?`

api/app/controllers/api/v1/reporting_tags_controller.rb
```diff
@@ -29,6 +39,17 @@ def update
             end
         end
 
+        # /reporting_tags/delete or /reporting_tags/:id/delete
+        def delete
+            params.require(:id)
+            reporting_tag = ReportingTag.find(params[:id])
+            if session.destroy!
```
dingtaoliu: typo

api/app/controllers/api/v1/wage_chunks_controller.rb
```diff
@@ -36,6 +41,17 @@ def update
             end
         end
 
+        # /wage_chunks/delete or /wage_chunks/:id/delete
+        def delete
+            params.require(:id)
+            wage_chunk = WageChunk.find(params[:id])
+            if session.destroy!
```
dingtaoliu: typo

api/app/controllers/api/v1/sessions_controller.rb
```diff
@@ -39,12 +39,13 @@ def update
...
         def delete
-                session = Session.find(params[:id])
-                if session.destroy!
-                    render_success(session)
-                else
-                    render_error(session.errors.full_messages.join("; "))
-                end
+            params.require(:id)
+            session = Session.find(params[:id])
+            if session.destroy!
```
dingtaoliu: not typo!

api/app/helpers/application_helper.rb
```diff
@@ -2,4 +2,18 @@
 
 # Base class from which most applications use
 module ApplicationHelper
+	def json(data, include: nil, except: nil)
+		data = JSON.parse(data.to_json.to_s, symbolize_names: true)
```
dingtaoliu: if you just want the keys in data to be symbolized, you can do `data = data.with_indifferent_access` instead of converting it into a json string and then parsing the json.

api/db/seeds.rb
```diff
@@ -3,65 +3,73 @@
...
-
-# users
-data = File.read("#{seed_data_dir}/user.json")
-raise JSON::ParserError.new("the source file is empty") if data.strip.length == 0
-json_data = JSON.parse(data).fetch("user")
-users = json_data.map do |row|
-  User.create!(row.with_indifferent_access)
-end
+insert_data(chaining, file: 'mock_data.json')
+insert_data(chaining, entries: entries)
```
dingtaoliu: let's comment out the auto generated data for now.

api/app/helpers/seeds_handler.rb
```diff
@@ -0,0 +1,184 @@
...
+            end
+        end
+        return true
+    end
+
+    def array_result(route)
+        return route.scan(/(\/add_)/).length > 0
+    end
+
+    def get_route(route, data)
```
dingtaoliu, line 168: can we add the name of the route into chaining? This avoids potential parsing errors.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Done.

api/app/controllers/api/v1/assignments_controller.rb
michellemtchai: I've fixed it.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: Fixed.

api/app/controllers/api/v1/position_templates_controller.rb
michellemtchai: Fixed.

api/app/controllers/api/v1/reporting_tags_controller.rb
michellemtchai: Fixed.

api/app/controllers/api/v1/wage_chunks_controller.rb
michellemtchai: Fixed.

api/db/seeds.rb
michellemtchai: Actually, I plan to remove this so that it's a separate task.

api/app/helpers/seeds_handler.rb
michellemtchai: Sure.

michellemtchai: Actually, I can't because I don't know the value of `:session_id` or `:id`, so I can't include them.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: Fixed.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: Fixed.

dingtaoliu:
I'm good with these changes, I think it can be merged.  Is it okay that the unit tests failed?

siefkenj:
@dingtaoliu We don't merge PRs with broken tests. This PR also needs to wait for  #183 to be ready.

api/db/seeds.rb
```diff
@@ -3,65 +3,72 @@
 # Seed data is contained in the json files under seed folder. This file serves as the script of 
 # pasing json file and create tables from the parsed data. The data can then be loaded with the rails db:seed command (or created alongside the
 # database with db:setup). For testing purpose, use db:reset to reload all the table.
+include SeedsHandler
 
-seed_data_dir = Rails.root.join('db', 'seed')
+chaining = [
```
siefkenj: variables like this should be named as nouns, not verbs. Something like `seed_data_sequence` would be better.

api/db/seeds.rb
```diff
@@ -3,65 +3,72 @@
...
+    {
+        get: '/sessions',
+        create: '/sessions',
+        unique: [:name]
```
siefkenj: `unique` is a confusing name. What do you think of `index_on`?

api/app/helpers/seeds_handler.rb
```diff
@@ -0,0 +1,259 @@
...
+                json, log = read_json(file, log)
+                valid_templates, log = check_templates(json, log)
+                if valid_templates
+                    log = post_data(chaining, json, log)
+                end
+            else
+                log.push('No JSON file was given.')
+            end
+        end
+        puts log
```
siefkenj: Isn't there a more offical way to log something?

api/app/helpers/seeds_handler.rb
```diff
@@ -0,0 +1,259 @@
...
+            }
+            if not templates.include?(data)
+                log.push("Invalid position_template: #{entry[:offer_template]}")
+                return false, log
+            end
+        end
+        return true, log
+    end
+
+    def post_data(chaining, json, log)
```
siefkenj: You're passing around `log` between several functions. This is a pattern `class`es were invented to solve. Would it make sense to refactor some of this into a class?

api/db/seeds.rb
michellemtchai: Fixed.

api/db/seeds.rb
michellemtchai: Fixed.

api/app/helpers/seeds_handler.rb
michellemtchai: There is. However, the problem with that it is the logging goes immediately into the log of the rest of the stuff Rack is logging at the same time, so any message that's relevant to the user kind of gets lost in between other really long messages. This is why I chose to accumulate the log messages because I print them.

api/app/helpers/seeds_handler.rb
michellemtchai: I'm not sure which pattern class you're referring to. For now, I made use of `@log` to make it so that there's no passing around `log` through several functions.

PR #175, opened 2019-07-06
[Frontend] Add components to dashboard

jacobljh:
Added assignments, instructors, position templates and positions


frontend/src/components/postition-templates-list.js
```diff
@@ -0,0 +1,38 @@
...
+    };
+    render() {
+        const { position_templates } = this.props;
+        let templateList = <div>No Templates...</div>;
+        if (position_templates.length > 0) {
+            templateList = (
+                <ul>
+                    {position_templates.map(template => (
+                        <li key={template.offer_template}>
+                            <font size="3" color="blue">
```
siefkenj: It would be better to use a `span` with a `className` here so that it can be styled with CSS

siefkenj (commented):
This is looking really good!

frontend/src/components/postition-templates-list.js
```diff
@@ -0,0 +1,40 @@
...
+            color: "blue"
+        };
+        const { position_templates } = this.props;
+        let templateList = <div>No Templates...</div>;
+        if (position_templates.length > 0) {
+            templateList = (
+                <ul>
+                    {position_templates.map(template => (
+                        <li key={template.offer_template}>
+                            <span style={positionTypeStyle}>
```
siefkenj: You should almost never use inline styles... (basically you only use them when drafting components and then you remove them later). Use an actual `className` and put some css into a corresponding css file.

frontend/src/components/postition-templates-list.js
```diff
@@ -0,0 +1,38 @@
...
+    };
+    render() {
+        const { position_templates } = this.props;
+        let templateList = <div>No Templates...</div>;
+        if (position_templates.length > 0) {
+            templateList = (
+                <ul>
+                    {position_templates.map(template => (
+                        <li key={template.offer_template}>
+                            <span className="positionTypeStyle">
```
siefkenj: CSS names usually use `-` instead of camel case. Also, there is no need for `Style` because it is CSS, it's already clear that it's a style.  This whole thing should be namespaced. Add a `className="template-list"` to the surrounding `div`. Then, your CSS selector can be

```css
.template-list .position-type {
   ...
}
```

frontend/src/components/postition-templates-list.js
```diff
@@ -0,0 +1,40 @@
...
+                                </span>
+                                , {template.offer_template}
+                            </li>
+                        ))}
+                    </ul>
+                </div>
+            );
+        }
+        return (
+            <div>
```
siefkenj: Your surrounding `div` is here ;-)

jacobljh: Hahaha, my bad lol.

frontend/src/components/postition-templates-list.js
```diff
@@ -0,0 +1,40 @@
...
+                offer_template: PropTypes.string
+            })
+        ).isRequired
+    };
+    render() {
+        const { position_templates } = this.props;
+        let templateList = <div>No Templates...</div>;
+        if (position_templates.length > 0) {
+            templateList = (
+                <div>
```
siefkenj: Donn't leave this `div` around if it isn't needed.

Issue #182, opened 2019-07-16
[Backend] change all PUT routes to POST


michellemtchai:
- change all put routes to post
- create routes for deleting tables

PR #183, opened 2019-07-16
[Backend]: Change all PUT routes to POST

michellemtchai:
Please see issue #182 for more details.

### DIFF
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -16,11 +16,6 @@ module Api::V1
 
         # POST /applicants
         def create
-            # if we passed in an id that exists, we want to update
-            if params[:id] && Applicant.exists?(params[:id])
-                update and return
-            end
-
             applicant = Applicant.new(applicant_params)
             if applicant.save # passes Applicant model validation
                 render_success(applicant)
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -34,6 +34,17 @@ module Api::V1
                 render_error(applicant.errors)
             end
         end
+
+        # /applicants/delete or /applicants/:id/delete
+        def delete
+            params.require(:id)
+            applicant = Applicant.find(params[:id])
+            if session.destroy!
+                render_success(applicant)
+            else
+                render_error(applicant.errors.full_messages.join("; "))
+            end
+        end
  
         private
         def applicant_params
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -16,6 +16,11 @@ module Api::V1
 
         # POST /applicants
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Applicant.exists?(params[:id])
+                update and return
+            end
+
             applicant = Applicant.new(applicant_params)
             if applicant.save # passes Applicant model validation
                 render_success(applicant)
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -16,6 +16,10 @@ module Api::V1
 
         # POST /applicants
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Applicant.exists?(params[:id])
+                update and return
+            end
             applicant = Applicant.new(applicant_params)
             if applicant.save # passes Applicant model validation
                 render_success(applicant)
@@ -25,7 +29,6 @@ module Api::V1
             end
         end
 
-        # POST /applicants/:id
         def update
             applicant = Applicant.find(params[:id])
             if applicant.update_attributes!(applicant_params)
@@ -35,7 +38,7 @@ module Api::V1
             end
         end
 
-        # POST /applicants/:id/delete
+        # POST /applicants/delete
         def delete
             params.require(:id)
             applicant = Applicant.find(params[:id])
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -35,7 +35,7 @@ module Api::V1
             end
         end
 
-        # /applicants/delete or /applicants/:id/delete
+        # POST /applicants/:id/delete
         def delete
             params.require(:id)
             applicant = Applicant.find(params[:id])
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -17,7 +17,7 @@ module Api::V1
         # POST /applicants
         def create
             # if we passed in an id that exists, we want to update
-            if params[:id] && Applicant.exists?(params[:id])
+            if params.has_key?(:id) and Applicant.exists?(params[:id])
                 update and return
             end
             applicant = Applicant.new(applicant_params)
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -30,7 +30,7 @@ module Api::V1
             end
         end
 
-        # PUT/PATCH /applicants/:id
+        # POST /applicants/:id
         def update
             applicant = Applicant.find(params[:id])
             if applicant.update_attributes!(applicant_params)
@@ -44,7 +44,7 @@ module Api::V1
         def delete
             params.require(:id)
             applicant = Applicant.find(params[:id])
-            if session.destroy!
+            if applicant.destroy!
                 render_success(applicant)
             else
                 render_error(applicant.errors.full_messages.join("; "))
--- a/api/app/controllers/api/v1/applicants_controller.rb
+++ b/api/app/controllers/api/v1/applicants_controller.rb
@@ -10,7 +10,7 @@ module Api::V1
                 render_success(Applicant.order(:id))
                 return
             end
-            if invalid_id(Session, :session_id) then return end
+            return if invalid_id(Session, :session_id)
             render_success(applicants_by_session)
         end
 
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -17,7 +17,7 @@ module Api::V1
         # POST /applications
         def create
             # if we passed in an id that exists, we want to update
-            if params[:id] && Application.exists?(params[:id])
+            if params.has_key?(:id) and Application.exists?(params[:id])
                 update and return
             end
             params.require(:applicant_id)
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -108,7 +108,7 @@ module Api::V1
 
         def all_applications
             return Application.order(:id).map do |entry|
-                application_data(application)
+                application_data(entry)
             end
         end
 
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -16,6 +16,10 @@ module Api::V1
 
         # POST /applications
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Application.exists?(params[:id])
+                update and return
+            end
             params.require(:applicant_id)
             return if invalid_id(Session, :session_id)
             return if invalid_id(Applicant, :applicant_id)
@@ -35,7 +39,6 @@ module Api::V1
             end
         end
 
-        # POST /applications/:id
         def update
             application = Application.find(params[:id])
             matching = application.applicant_data_for_matching
@@ -49,8 +52,9 @@ module Api::V1
             end
         end
 
-        # POST /applications/:id/delete
+        # POST /applications/delete
         def delete
+            params.require(:id)
             application = Application.find(params[:id])
             entry = application_data(application)
             if application
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -58,6 +58,11 @@ module Api::V1
         def delete
             params.require(:id)
             application = Application.find(params[:id])
+            if application
+                applicant_data_for_matching = ApplicantDataForMatching.find_by(
+                    application_id: params[:id])
+                applicant_data_for_matching.destroy!
+            end
             if application.destroy!
                 render_success(application)
             else
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -40,7 +40,7 @@ module Api::V1
             end
         end
 
-        # PUT/PATCH /applications/:id
+        # POST /applications/:id
         def update
             application = Application.find(params[:id])
             matching = ApplicantDataForMatching.find_by!(application_id: application[:id])
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -49,6 +49,17 @@ module Api::V1
             end
         end
 
+        # /applications/delete or /applications/:id/delete
+        def delete
+            params.require(:id)
+            application = Application.find(params[:id])
+            if application.destroy!
+                render_success(application)
+            else
+                render_error(application.errors.full_messages.join("; "))
+            end
+        end
+
         private
         def application_params
             params.permit(
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -28,7 +28,7 @@ module Api::V1
             params[:application_id] = application[:id]
             message = valid_applicant_matching_data(application.errors.messages)
             if not message
-                render_success(application)
+                render_success(application_data(application))
             else
                 application.destroy!
                 render_error(message)
@@ -38,12 +38,12 @@ module Api::V1
         # POST /applications/:id
         def update
             application = Application.find(params[:id])
-            matching = ApplicantDataForMatching.find_by!(application_id: application[:id])
+            matching = application.applicant_data_for_matching
             application_res = application.update_attributes!(application_update_params)
             matching_res = matching.update_attributes!(matching_data_update_params)
             errors = application.errors.messages.deep_merge(matching.errors.messages)
             if application_res and matching_res
-                render_success(application)
+                render_success(application_data(application))
             else
                 render_error(errors)
             end
@@ -51,15 +51,14 @@ module Api::V1
 
         # POST /applications/:id/delete
         def delete
-            params.require(:id)
             application = Application.find(params[:id])
+            entry = application_data(application)
             if application
-                applicant_data_for_matching = ApplicantDataForMatching.find_by(
-                    application_id: params[:id])
-                applicant_data_for_matching.destroy!
+                matching = application.applicant_data_for_matching
+                matching.destroy!
             end
             if application.destroy!
-                render_success(application)
+                render_success(entry)
             else
                 render_error(application.errors.full_messages.join("; "))
             end
@@ -108,12 +107,16 @@ module Api::V1
         end
 
         def all_applications
-            exclusion = [:id, :created_at, :updated_at, :application_id]
             return Application.order(:id).map do |entry|
-                matching = ApplicantDataForMatching.find_by(application_id: entry[:id])
-                matching = json(matching, except: exclusion)
-                json(entry, include: matching)
+                application_data(application)
             end
+        end
+
+        def application_data(application)
+            exclusion = [:id, :created_at, :updated_at, :application_id]
+            matching = application.applicant_data_for_matching
+            matching = json(matching, except: exclusion)
+            return json(application, include: matching)
         end
 
         def valid_applicant_matching_data(errors)
--- a/api/app/controllers/api/v1/applications_controller.rb
+++ b/api/app/controllers/api/v1/applications_controller.rb
@@ -16,7 +16,9 @@ module Api::V1
 
         # POST /applications
         def create
+            params.require(:applicant_id)
             if invalid_id(Session, :session_id) then return end
+            if invalid_id(Applicant, :applicant_id) then return end
             application = Application.new(application_params)
             if not application.save # does not pass Application model validation
                 application.destroy!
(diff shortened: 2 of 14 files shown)


api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -54,10 +68,18 @@ def instructors_by_position
...
+                result = error = instructors_by_position
+            else
+                result = instructor
+                error = {}
+            end
+            if instructor.save # passes Instructor model validation
+                render_success(result)
+            else
+                instructor.destroy!
+                render_error(instructor.errors, error)
```
siefkenj: Why not `render_error(instructor.errors.full_messages.join("; "))`?

michellemtchai: Good catch. I'll fix that now. Thanks.

api/app/controllers/api/v1/applicants_controller.rb
```diff
@@ -16,6 +16,11 @@ def index
...
         # POST /applicants
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Applicant.exists?(params[:id])
+                update and return
+            end
+
             applicant = Applicant.new(applicant_params)
             if applicant.save # passes Applicant model validation
                 render_success(applicant)
```
George334456, line 25: For line 34 in this file, couldn't we techinically hit PUT/PATCH on applicants/:id?

api/app/controllers/api/v1/applicants_controller.rb
```diff
@@ -16,6 +16,11 @@ def index
 
         # POST /applicants
         def create
+            # if we passed in an id that exists, we want to update
```
George334456, line 19: To clarify this defines two routes:
`/applicants` which will create an applicant based on the things passed in.
`/applicants/:id` which will update an existing applicant based on ID.

Side note: if you pass in an `id` that does not exist, this will behave like `/applicants` and ignore the id parameter. Is this intended or should we give a `id not found` error?

api/app/controllers/api/v1/applicants_controller.rb
```diff
@@ -34,6 +39,17 @@ def update
                 render_error(applicant.errors)
             end
         end
+
+        # /applicants/delete or /applicants/:id/delete
```
George334456: `/applicants/delete` can't happen right? Because you require the id.

api/app/controllers/api/v1/applicants_controller.rb
```diff
@@ -34,6 +39,17 @@ def update
...
+        def delete
+            params.require(:id)
+            applicant = Applicant.find(params[:id])
+            if session.destroy!
```
George334456: Why destroy the `session`? Aren't we destroying the `applicant`?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -84,11 +107,20 @@ def matching_data_update_params
...
-            return Application.order(:id).select do |entry|
+            return all_applications.select do |entry|
                 entry[:session_id] == params[:session_id].to_i
             end
         end
 
+        def all_applications
+            exclusion = [:id, :created_at, :updated_at, :application_id]
+            return Application.order(:id).map do |entry|
+                matching = ApplicantDataForMatching.find_by(application_id: entry[:id])
```
George334456: `Application` has one `ApplicantDataForMatching`.

Couldn't we do `matching = entry.ApplicantDataForMatching`?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -16,7 +16,14 @@ def index
 
         # POST /applications
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Application.exists?(params[:id])
```
George334456: Same comment for `POST applicant`

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -16,7 +16,14 @@ def index
...
+                update and return
+            end
+
+            params.require(:applicant_id)
             if invalid_id(Session, :session_id) then return end
+            if invalid_id(Applicant, :applicant_id) then return end
```
George334456: Is `invalid_id` defined somewhere?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -16,7 +16,14 @@ def index
 
         # POST /applications
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Application.exists?(params[:id])
+                update and return
+            end
+
+            params.require(:applicant_id)
             if invalid_id(Session, :session_id) then return end
```
George334456: I'm assuming that something like `:session_id` is passed in the JSON body right? If not, where did we store it?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -36,7 +43,7 @@ def create
         # PUT/PATCH /applications/:id
         def update
             application = Application.find(params[:id])
-            matching = ApplicantDataForMatching.find(application[:id])
+            matching = ApplicantDataForMatching.find_by!(application_id: application[:id])
```
George334456: Again, `Application` has a `has one` relationship to `ApplicantDataForMatching`. Couldn't we just do:
`application.ApplicantDataForMatching`?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -47,6 +54,22 @@ def update
             end
         end
 
+        # /applications/delete or /applications/:id/delete
```
George334456: Again, if `id` is required, we should be never able to access `applications/delete` right?

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -47,6 +54,22 @@ def update
...
+        def delete
+            params.require(:id)
+            application = Application.find(params[:id])
+            if application
+                applicant_data_for_matching = ApplicantDataForMatching.find_by(
```
George334456: Another place to potentially change to `application.applicant_data_for_matching`

api/app/controllers/api/v1/applications_controller.rb
```diff
@@ -47,6 +54,22 @@ def update
...
+                    application_id: params[:id])
+                applicant_data_for_matching.destroy!
```
George334456: Also are we guaranteed applicant_data_for_matching? If there isn't one, won't `destroy` fail and muck up everything down stream? Should we gate this with an `applicant_data_for_matching.destroy! if applicant_data_for_matching`?

api/app/controllers/api/v1/assignments_controller.rb
```diff
@@ -16,6 +16,11 @@ def index
 
         # POST /assignments
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Assignment.exists?(params[:id])
```
George334456: Again, if an invalid id is passed in, won't this just do a regular create?

api/app/controllers/api/v1/assignments_controller.rb
```diff
@@ -44,6 +49,17 @@ def update
             end
         end
 
+        # /assignments/delete or /assignments/:id/delete
+        def delete
+            params.require(:id)
+            assignment = Assignment.find(params[:id])
+            if session.destroy!
```
George334456: Why are we destroying the session? Shouldn't we be destroying the assignment?

api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -14,30 +14,44 @@ def index
...
-            position = Position.find(params[:position_id])
-            instructor = position.instructors.new(instructor_params)
-            if instructor.save # passes Instructor model validation
-                render_success(instructors_by_position)
+            # if we passed in an id that exists, we want to update
+            if params[:id] && Instructor.exists?(params[:id])
+                update and return
+            end
+
+            if not params.include?(:position_id)
```
George334456: To get this straight, if there isn't a position_id in the parameters, we would create a new instructor.

If there is a position_id in the parameters, then we are adding an instructor to that position.

Is that correct?

api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -14,30 +14,44 @@ def index
...
             else
                 render_error(instructor.errors)
             end
         end
 
+        # /instructors/delete or /instructors/:id/delete
+        def delete
+            params.require(:id)
+            instructor = Instructor.find(params[:id])
+            if session.destroy!
```
George334456: Same as previous.

api/app/controllers/api/v1/instructors_controller.rb
```diff
@@ -14,30 +14,44 @@ def index
...
+            end
+
+            if not params.include?(:position_id)
+                instructor = Instructor.new(instructor_params)
+                instructor_create(instructor)
             else
-                instructor.destroy!
-                render_error(instructor.errors, instructors_by_position)
+                position = Position.find(params[:position_id])
+                instructor = position.instructors.new(instructor_params)
```
George334456: Does this syntax create a new instructor from the instructor_params? If my previous comment was correct, then I thought you could only do:
```
instructor = Instructor.new(instructor_params)
instructor_create(instructor, true)
position.instructors.append(instructor)
position.save!
```
To create a new instructor, then add a new instructor to the position.

api/app/controllers/api/v1/position_preferences_controller.rb
```diff
@@ -4,8 +4,18 @@ module Api::V1
...
-        # POST /add_preference
+        # GET /position_preferences
+        def index
+            render_success(PositionPreference.order(:id))
+        end
+
+        # POST /add_preference and /position_preferences
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && PositionPreference.exists?(params[:id])
```
George334456: Same as previous

api/app/controllers/api/v1/position_preferences_controller.rb
```diff
@@ -28,6 +38,17 @@ def update
             end
         end
 
+        # /position_preferences/delete or /position_preferences/:id/delete
+        def delete
```
George334456, line 40: Same as previous

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -6,11 +6,21 @@ class PositionTemplatesController < ApplicationController
...
+            end
+            if invalid_id(Session, :session_id) then return end
             render_success(position_templates_by_session)
         end
 
-        # POST /add_position_template
+        # POST /add_position_template and /position_templates
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && PositionTemplate.exists?(params[:id])
```
George334456: Same as previous

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -38,6 +48,17 @@ def update
             end
         end
 
+        # /position_templates/delete or /position_templates/:id/delete
+        def delete
```
George334456, line 50: Same as previous

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -38,6 +48,17 @@ def update
...
+            params.require(:id)
+            position_template = PostionTemplate.find(params[:id])
+            if session.destroy!
```
George334456: Same as previous.

api/app/controllers/api/v1/positions_controller.rb
```diff
@@ -24,6 +29,7 @@ def create
                 render_error(position.errors.full_messages.join("; "))
                 return
             end
+            update_instructors_ids(position)
```
George334456, line 31: Don't you have to save once you've updated the instructor ids for the position?

api/app/controllers/api/v1/reporting_tags_controller.rb
```diff
@@ -4,8 +4,18 @@ module Api::V1
...
-        # POST /add_reporting_tag
+        # GET /reporting_tags
+        def index
+            render_success(ReportingTag.order(:id))
+        end
+
+        # POST /add_reporting_tag and /reporting_tags
         def create
+            # if we passed in an id that exists, we want to update
+            if params[:id] && ReportingTag.exists?(params[:id])
```
George334456: Same as above.

api/app/controllers/api/v1/reporting_tags_controller.rb
```diff
@@ -29,6 +39,17 @@ def update
             end
         end
 
+        # /reporting_tags/delete or /reporting_tags/:id/delete
+        def delete
+            params.require(:id)
+            reporting_tag = ReportingTag.find(params[:id])
+            if session.destroy!
```
George334456: Same as above.

api/app/controllers/api/v1/wage_chunks_controller.rb
```diff
@@ -36,6 +41,17 @@ def update
             end
         end
 
+        # /wage_chunks/delete or /wage_chunks/:id/delete
+        def delete
+            params.require(:id)
+            wage_chunk = WageChunk.find(params[:id])
+            if session.destroy!
```
George334456: Same as above.

api/app/controllers/api/v1/wage_chunks_controller.rb
```diff
@@ -14,8 +14,13 @@ def index
             render_success(wage_chunks_by_assignment)
         end
 
-        # POST /add_wage_chunk
+        # POST /add_wage_chunk and /wage_chunks
         def create
+            # if we passed in an id that exists, we want to update
```
George334456, line 19: Same as above

api/app/controllers/api/v1/sessions_controller.rb
```diff
@@ -39,12 +39,13 @@ def update
 
         # /sessions/delete or /sessions/:id/delete
         def delete
```
George334456, line 39: Again, would `delete` work without an `id` in the url?

George334456 (changes requested):
Left some comments. Let's discuss them whenever you have time.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: Good catch. This was a mistake. I'll fix that. Thanks.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: Actually, it can thanks to the `config/routes.rb`. Though, I have no idea why there are two ways to delete. @siefkenj was the one who set up this structure. You would have to ask him if you want to know why it was setup this way.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Yes, the `session_id` is stored in the JSON body.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: It is in `app/controllers/concerns/response.rb`.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Good point. Thanks for catching that. I'll change it.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: I'll fix this as well. Thanks.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: It should be guaranteed because the controller is assuming that our tables are always created via the API routes. If there was an error in the `applicant_data_for_matching` creation, it should have shown in the error message and the corresponding `application` should have been deleted.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: I'll fix this. Thanks for catching this.

api/app/controllers/api/v1/applications_controller.rb
George334456: I meant it in the sense that couldn't we have an application without an applicant data for matching?

api/app/controllers/api/v1/assignments_controller.rb
michellemtchai: Yes. I'm assuming that what we want(?) The initial structure for this setup was made by @siefkenj, so you have to ask him about the specifics on why it was designed this way.

api/app/controllers/api/v1/assignments_controller.rb
michellemtchai: Good catch. I'll fix this. Thanks.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: It's more like if there isn't a `position_id`, we are creating a new `instructor` and adding that new `instructor` into `position`.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: This syntax does create a new `instructor` from `instructor_params`. I'll check out if it works differently from the code you wrote, though.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: That is actually a very interesting point. I'll check if save affects anything.

api/app/controllers/api/v1/instructors_controller.rb
George334456: If there is no position_id, how can you add an instructor to the position since you don't know the `position_id`?

api/app/controllers/api/v1/instructors_controller.rb
George334456: I'm just wondering if the position with a new instructor is really being saved. i know you save the instructor, but I don't know if you save the fact that this position now has a new instrucotr.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: I don't know if the former or latter is the intended behaviour. Usually, I would think it is the latter, which will make this implementation. However, when I was writing this section, @siefkenj said I should follow the format he used for this `update` vs. `create` separation. You would have to ask him for more details.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: Yes, we could do that. I guess I'll just remove that if statement.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: I think I'll delete the former. It's kind of redundant anyways.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: Fixed.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: Nevermind, I just realized the `/applicants/delete` exist because without it, this route is taken as `/applicants/:id`, where `id=delete`...

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Not really. @siefkenj's explanation for this was that the `application` and `applicant_data_for_matching` are really meant to be the same table and that they are only separated for semantics.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: Sorry. I just realized that I made a mistake in my reply. I was referring to when there **is** a `position_id`.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: I ended up creating `/applicants/:id` sort of `update` routes instead of rerouting via `create`.

api/app/controllers/api/v1/applicants_controller.rb
michellemtchai: I deleted the `/applicants/delete` sort of route and kept only the `/applicants/:id/delete` type of routes.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Done.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Done.

api/app/controllers/api/v1/applications_controller.rb
michellemtchai: Done.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: I think the `save` is actually a misnomer. It's not really saving to the db, `save` actually just checks if the data matching the table constraints. `new` already inserted the data into the db.

api/app/controllers/api/v1/assignments_controller.rb
George334456: Hmm, this will have to be tested and kept consistent on live. If this is the supposition, we need to make sure that this is confirmed.

api/app/controllers/api/v1/instructors_controller.rb
George334456: https://guides.rubyonrails.org/active_record_basics.html#create

According to the docs though new will not insert the data to the database? Did you try an `object.reload` to reload it from database and check that the changes have propagated?

api/app/controllers/api/v1/assignments_controller.rb
michellemtchai: I got rid of the `create` rerouting to `update` and made the routes directly call `POST /tables/:id` instead.

api/app/controllers/api/v1/instructors_controller.rb
michellemtchai: Okay. I'll check that.

michellemtchai: I tried the `reload`, but it seems like `save` is only really relevant when creating a table. If you update, say, `position.instructors` after creation, it does persist.

api/app/controllers/api/v1/positions_controller.rb
michellemtchai: While it doesn't seem like saving changes the persistence in the db here, should I `save` anyways?

George334456: I guess we couuuuuuuuuuuuuuuuuuuuuuuld. We'll have to test it somehow. I feel that something will pop up in prod when we try LUL

PR #186, opened 2019-07-19
[Frontend] front end API test 

jacobljh:
To run the test, run `docker-compose run frontend npm test`

Some fixing with backend 
- ~~`session/:session id/instructors` routes are not supported right now.~~
	- added `session/:session id/instructors`
- ~~does not check whether the `offer_template` or `position_type` is empty when creating/updating `position_templates`~~
	- fixed
- ~~can't delete an instructor from a session~~
    - added `session/:session id/instructors/delete`

### DIFF
--- a/api/app/helpers/application_helper.rb
+++ b/api/app/helpers/application_helper.rb
@@ -2,4 +2,18 @@
 
 # Base class from which most applications use
 module ApplicationHelper
+	def json(data, include: nil, except: nil)
+		data = JSON.parse(data.to_json.to_s, symbolize_names: true)
+		if include
+			include.keys.each do |key|
+				data[key] = include[key]
+			end
+		end
+		if except
+			except.each do |key|
+				data.delete(key)
+			end
+		end
+		return data
+	end
 end
--- a/api/app/helpers/application_helper.rb
+++ b/api/app/helpers/application_helper.rb
@@ -2,15 +2,15 @@
 
 # Base class from which most applications use
 module ApplicationHelper
-	def json(data, include: nil, except: nil)
-		if not data.is_a?(ActiveSupport::HashWithIndifferentAccess)
-			data = data.as_json(except: except).with_indifferent_access
-		end
-		if include
-			include.keys.each do |key|
-				data[key] = include[key]
-			end
-		end
-		return data
-	end
+    def json(data, include: nil, except: nil)
+        if not data.is_a?(ActiveSupport::HashWithIndifferentAccess)
+            data = data.as_json(except: except).with_indifferent_access
+        end
+        if include
+            include.keys.each do |key|
+                data[key] = include[key]
+            end
+        end
+        return data
+    end
 end
--- a/api/app/helpers/application_helper.rb
+++ b/api/app/helpers/application_helper.rb
@@ -3,15 +3,12 @@
 # Base class from which most applications use
 module ApplicationHelper
 	def json(data, include: nil, except: nil)
-		data = JSON.parse(data.to_json.to_s, symbolize_names: true)
+		if not data.is_a?(ActiveSupport::HashWithIndifferentAccess)
+			data = data.as_json(except: except).with_indifferent_access
+		end
 		if include
 			include.keys.each do |key|
 				data[key] = include[key]
-			end
-		end
-		if except
-			except.each do |key|
-				data.delete(key)
 			end
 		end
 		return data
--- a/frontend/src/__tests__/api.test.js
+++ b/frontend/src/__tests__/api.test.js
@@ -107,8 +107,7 @@ function sessionsTests(api = { apiGET, apiPOST }) {
     });
 
     it("update a session", async () => {
-        const id = session.id;
-        const newData = { id, rate2: 57.75 };
+        const newData = { ...session, rate1: 57.75 };
         const resp1 = await apiPOST("/sessions", newData);
         expect(resp1).toMatchObject({ status: "success" });
         expect(resp1.payload).toMatchObject(newData);
@@ -116,22 +115,27 @@ function sessionsTests(api = { apiGET, apiPOST }) {
         // get the sessions list and make sure we're updated there as well
         const resp2 = await apiGET("/sessions");
         // filter session list to get the updated session obj
-        const updatedSession = resp2.payload.filter(s => s.id == id);
+        const updatedSession = resp2.payload.filter(s => s.id == session.id);
         expect(updatedSession).toContainObject(newData);
     });
 
     it("throw error when `name` is empty", async () => {
         // create new session with empty name
-        const newData = { name: "" };
-        const resp1 = await apiPOST("/sessions", newData);
-        // expected an error
+        const newData1 = { ...newSessionData, name: "" };
+        const newData2 = { ...newSessionData, name: undefined };
+
+        const resp1 = await apiPOST("/sessions", newData1);
         expect(resp1).toMatchObject({ status: "error" });
         checkPropTypes(errorPropTypes, resp1);
+
+        const resp2 = await apiPOST("/sessions", newData2);
+        expect(resp2).toMatchObject({ status: "error" });
+        checkPropTypes(errorPropTypes, resp2);
     });
 
     it("throw error when `name` is not unique", async () => {
         // name identical to the exisiting session
-        const newData = { name: session.name };
+        const newData = { ...newSessionData, name: session.name };
         // POST to create new session
         const resp1 = await apiPOST("/sessions", newData);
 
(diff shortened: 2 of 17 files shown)


frontend/src/api/mockAPI.js
```diff
@@ -57,6 +57,19 @@ export class MockAPI {
                 return Object.assign(matchingSessions[0], body);
             }
             // if we're here, we need to create a new session
+            // but check if the session name is empty or duplicate
+            if (body.name === null || body.name === "") {
```
siefkenj: Use `==` to compare with `null` (`undefined == null` is true, but `undefined === null` is false)

frontend/src/api/mockAPI.js
```diff
@@ -57,6 +57,19 @@ export class MockAPI {
...
+                throw new Error("Session name cannot be empty!");
+            }
+            const duplicateSessions = data.sessions.filter(
+                s => s.name === body.name
+            );
+            if (duplicateSessions.length > 0) {
```
siefkenj: You can change this to
```
if (data.sessions.some( s => s.name === body.name )) {...}
```

frontend/src/__tests__/utils.js
```diff
@@ -147,6 +147,46 @@ export async function deleteSession(api, session) {
...
+ * Delete a position
+ *
+ * @export
+ * @param {{apiGET: Function, apiPOST: Function}} api api object containing `apiGET` and `apiPOST`
+ * @param {{id: number}} position position to be deleted (must have an `id` attribute)
+ * @returns
+ */
+export async function deletePosition(api, position) {
+    let resp = {};
+    resp = await api.apiPOST(`/positions/delete`, position);
```
siefkenj: No need to delcare the variable and give it a default value...
```
const resp = await api.apiPOST("/positions/delete", position);
```

If you're not interpolating, use " instead of `

frontend/src/__tests__/utils.js
```diff
@@ -147,6 +147,46 @@ export async function deleteSession(api, session) {
...
+        est_start_date: "2018/05/09",
+        est_end_date: "2018/09/09",
+        position_type: "Standard"
+    };
+    let resp = {};
+    resp = await api.apiPOST(
+        `/sessions/${session.id}/positions`,
+        newPositionData
+    );
+    const position = resp.payload;
```
siefkenj: Why are you creating unneeded variables? See comment below

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+    it("fetch sessions", async () => {
+        const resp = await apiGET("/sessions");
+
+        expect(resp).toMatchObject({ status: "success" });
+        checkPropTypes(PropTypes.arrayOf(sessionPropTypes), resp.payload);
+    });
+
+    it("update a session", async () => {
+        const id = session.id;
+        const newData = { id, rate2: 57.75 };
```
siefkenj: How about ` { ...session, rate2: 57.75 }`

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+        // get the sessions list and make sure we're updated there as well
+        const resp2 = await apiGET("/sessions");
+        // filter session list to get the updated session obj
+        const updatedSession = resp2.payload.filter(s => s.id == id);
+        expect(updatedSession).toContainObject(newData);
+    });
+
+    it("throw error when `name` is empty", async () => {
+        // create new session with empty name
+        const newData = { name: "" };
```
siefkenj: test with `{ ...newSessionData, name: undefined}` and  `{ ...newSessionData, name: ""}`

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+        const resp1 = await apiPOST("/sessions", newData);
+        // expected an error
+        expect(resp1).toMatchObject({ status: "error" });
+        checkPropTypes(errorPropTypes, resp1);
+    });
+
+    it("throw error when `name` is not unique", async () => {
+        // name identical to the exisiting session
+        const newData = { name: session.name };
```
siefkenj: `{...newSessionData, name: session.name}`

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+        const resp1 = await apiPOST("/sessions", newData);
+
+        // expected an error as name not unique
+        expect(resp1).toMatchObject({ status: "error" });
+        checkPropTypes(errorPropTypes, resp1);
+    });
+
+    it("throw error when deleting item with invalid id", async () => {
+        const resp1 = await apiPOST("/sessions/delete", {
+            id: session.id + Math.floor(Math.random() * 100) + 1 // returns a random integer from 1 to 100
```
siefkenj: I don't like this....get all the sessions, find the largest ID and then add 1 to it to ensure you found an ID that doesn't exist. Also, try to delete something with `id=null`

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+        const resp1 = await apiPOST(
+            `/sessions/${session.id}/add_position_template`,
+            newTemplateDataToUpdate
+        );
+        expect(resp1).toMatchObject({ status: "success" });
+        checkPropTypes(
+            PropTypes.arrayOf(offerTemplatePropTypes),
+            resp1.payload
+        );
+        expect(resp1.payload).toContainObject(newTemplateDataToUpdate);
```
siefkenj: We already checked this stuff in the previous test. We don't need to do it again. (add a comment though, that `add_position_template` was already tested.)

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+            resp2.payload
+        );
+        expect(resp2.payload).toContainObject(newTemplateDataToUpdate);
 
-        // clean up by deleting the session
-        await apiPOST("/sessions/delete", {
-            id: sessionId
+        const templateToUpdate = resp2.payload.filter(t => {
+            return (
+                t.offer_template == newTemplateDataToUpdate.offer_template &&
```
siefkenj: `==`  -> `===`

frontend/src/__tests__/api.test.js
```diff
@@ -96,95 +95,230 @@ function sessionsTests(api = { apiGET, apiPOST }) {
...
+            id: templateToUpdate[0].id,
+            offer_template: "this_is_a_updated_test_template.html",
+            position_type: "Standard"
+        };
+        const resp3 = await apiPOST(
+            `/sessions/${session.id}/add_position_template`,
+            updateData
+        );
+        expect(resp3).toMatchObject({ status: "success" });
+        expect(resp3.payload).toMatchObject(updateData);
```
siefkenj: Shouldn't this be an array just like all the other calls to this API?

frontend/src/__tests__/api.test.js
```diff
@@ -253,9 +387,41 @@ function positionsTests(api = { apiGET, apiPOST }) {
...
+            `/sessions/${sessionId}/positions`,
+            newPositionData2
+        );
+        expect(resp2).toMatchObject({ status: "success" });
+        // make sure we get back what we put in
+        expect(resp2.payload).toMatchObject(newPositionData2);
+
+        // make sure the position is created
+        const resp3 = await apiGET(`/sessions/${sessionId}/positions`);
+        expect(resp3.payload).toContainObject(resp2.payload);
```
siefkenj, line 417: Do we also have a test that ensures that creating two positions with the same code in the same session indeed fails?

frontend/src/__tests__/api.test.js
```diff
@@ -269,11 +435,81 @@ function positionsTests(api = { apiGET, apiPOST }) {
...
+        checkPropTypes(PropTypes.arrayOf(instructorPropTypes), resp.payload);
+    });
+
+    it("get instructors for session", async () => {
+        const resp = await apiGET(`/sessions/${session.id}/instructors`);
+        expect(resp).toMatchObject({ status: "success" });
+        checkPropTypes(PropTypes.arrayOf(instructorPropTypes), resp.payload);
+    });
+
+    it("create instructor", async () => {
```
siefkenj, line 447: Make this the first test so that we know something gets returned in the `get` tests.

frontend/src/__tests__/api.test.js
```diff
@@ -269,11 +435,81 @@ function positionsTests(api = { apiGET, apiPOST }) {
...
+        // update the instructor
+        const resp = await apiPOST(`/instructors`, updateInstructorData);
+        expect(resp).toMatchObject({ status: "success" });
+        expect(resp.payload).toMatchObject(updateInstructorData);
+    });
+
+    // delete an instructor
+    it("delete instructor", async () => {
+        const resp = await apiPOST(`/instructors/delete`, instructor);
+        expect(resp).toMatchObject({ status: "success" });
```
siefkenj: Get the instructor list and make sure they are not there. We also need to delete an instructor from a session.

api/app/helpers/application_helper.rb
```diff
@@ -2,4 +2,15 @@
...
+		if not data.is_a?(ActiveSupport::HashWithIndifferentAccess)
+			data = data.as_json(except: except).with_indifferent_access
+		end
+		if include
+			include.keys.each do |key|
+				data[key] = include[key]
+			end
+		end
+		return data
+	end
```
siefkenj: Spaces, not tabs

PR #187, opened 2019-07-20
fix(api): added error handling for undefined api routes

dingtaoliu:
Added a 404 route that matches urls for undefined '/api' routes for the rails backend.  Previously, undefined api routes are handled as frontend react router routes and rails would return a 200 status.

### DIFF
--- a/api/app/controllers/application_controller.rb
+++ b/api/app/controllers/application_controller.rb
@@ -7,10 +7,9 @@ class ApplicationController < ActionController::Base
 
   	def index
     	render :file => 'public/index.html'
-		end  
+    end  
 		
 		def not_found
-			raise ActionController::RoutingError.new('Unknown Route')
+			render json: {status: 'not_found'}, status: :not_found
 		end
-
 end
--- a/api/app/controllers/application_controller.rb
+++ b/api/app/controllers/application_controller.rb
@@ -7,6 +7,10 @@ class ApplicationController < ActionController::Base
 
   	def index
     	render :file => 'public/index.html'
-  	end  
+		end  
+		
+		def not_found
+			raise ActionController::RoutingError.new('Unknown Route')
+		end
 
 end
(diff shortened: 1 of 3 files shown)


api/app/controllers/application_controller.rb
```diff
@@ -7,6 +7,10 @@ class ApplicationController < ActionController::Base
 
   	def index
     	render :file => 'public/index.html'
-  	end  
+		end
```
siefkenj: spaces, not tabs

siefkenj (commented):
Can you add an javascript API test for this?

api/app/controllers/application_controller.rb
```diff
@@ -7,6 +7,9 @@ class ApplicationController < ActionController::Base
 
   	def index
     	render :file => 'public/index.html'
-  	end  
-
+    end  
+		
+		def not_found
```
siefkenj: Spaces, not tabs

frontend/src/__tests__/api.test.js
```diff
@@ -323,6 +323,20 @@ function reportingTagsTests({ apiGET, apiPOST }) {
 // eslint-disable-next-line
 function applicationsTests({ apiGET, apiPOST }) {}
 
+function unknownRouteTests(api = { apiGet, apiPost }) {
+    const { apiGet, apiPost } = api;
+
+    it("should succeed GET request with known '/api' routes", async () => {
```
siefkenj: This test is not needed. It is already done elsewhere

frontend/src/__tests__/api.test.js
```diff
@@ -352,6 +366,9 @@ describe("API tests", () => {
     describe("`/applications` tests", () => {
         applicationsTests({ apiGET, apiPOST });
     });
+    describe("`/404` tests", () => {
```
siefkenj: Describe these tests better

PR #188, opened 2019-07-21
[WIP] [FRONTEND] added offers view on /tapp, and on /dashboard.

SimonAayani:
Added a view for the offers table on /tapp and on /dashboard.

### /tapp
- Added 'Offers' tab to the header
- Currently uses local data, working on using API calls to get the data instead

### /dashboard

-  Added OffersList component to dashboard page.

- Added empty actions.js, reducer.js, constants.js files for offers module to be filled in later.

### DIFF
--- a/api/config/environments/development.rb
+++ b/api/config/environments/development.rb
@@ -9,7 +9,7 @@ Rails.application.configure do
   config.cache_classes = false
 
   # Do not eager load code on boot.
-  config.eager_load = false
+  config.eager_load = true
 
   # Show full error reports.
   config.consider_all_requests_local = true


siefkenj:
Run the code formatting tests and make sure to fix all the errors.

frontend/src/modules/offers/components/Offers.jsx
```diff
@@ -0,0 +1,93 @@
...
+                        showPagination={false}
+                        pageSize={1}
+                        columns={COLUMNS}
+                        data={data}
+                        className={"positions-table"}
+                        noDataText={"No Offers Found"}
+                        SubComponent={({ original }) => {
+                            return (
+                                <div
+                                    style={{
```
siefkenj: This should be done with `className` and corresponding CSS

frontend/src/api/actions/offers.js
```diff
@@ -0,0 +1,45 @@
...
+import { FETCH_OFFERS_SUCCESS } from "../constants";
+import { fetchError } from "./errors";
+//import { fetchError, upsertError, deleteError } from "./errors";
+import {
+    actionFactory,
+    runOnActiveSessionChange,
+    validatedApiDispatcher
+} from "./utils";
+import { apiGET } from "../../libs/apiUtils";
+//import { apiGET, apiPOST } from "../../libs/apiUtils";
```
siefkenj: Don't leave this comment around

SimonAayani:
Noticed that when I was trying to run my API calls (Specifically in the session routes) the server would hang and it would not populate the data properly. Forcing shut down and doing docker-compose up again would yield different results without having changed any code. I.e. sometimes the Positions view on the dashboard would populate, sometimes multiple views, and sometimes none. After searching through the internet, it said to set

 `config.eager_load = false` 

(but our development config file already had it set to false), so I tried setting it to true and it fixed it (Temporary fix I think). This is in 

`/api/config/environments/development.rb`

Issue #189, opened 2019-07-24
[API] update available_position_templates


michellemtchai:
- show only relative path
- use `DEPARTMENT` rails environment variable to specify subfolder

PR #190, opened 2019-07-24
feat(api): update available_position_templates

michellemtchai:
See issue #189 for more details


siefkenj:
Let's change this so we only have `oto.html` `standard.html` and `invigilate.html`. Put them in a `defaultdepartment` directory and set the environment variable to `DEPARTMENT=defaultdepartment`

api/app/controllers/api/v1/position_templates_controller.rb
```diff
@@ -24,7 +24,12 @@ def create
 
         # GET /available_position_templates
         def available
-            files = Dir.glob("#{Rails.root}/app/views/position_templates/*")
+            dir = "#{Rails.root}/app/views/position_templates/"
+            files = Dir.glob("#{dir}/#{ENV['DEPARTMENT']}/*").map do |entry|
+                {
+                    offer_template: entry.gsub(dir, '')
```
siefkenj: `sub` not `gsub`

michellemtchai: Fixed.

Issue #192, opened 2019-07-28
[Backend] Add some model unit tests


George334456:
**Acceptance Criteria** 
- New factories for all the models
- Start some sample unit tests on the models

dingtaoliu:
done in #195

PR #195, opened 2019-07-30
TAPP-192: Added in factories and starter specs

George334456:
See issue #192 for details. Upgraded the factorybots

### DIFF
--- a/api/spec/factories/applicant_data_for_matching.rb
+++ b/api/spec/factories/applicant_data_for_matching.rb
@@ -2,8 +2,8 @@
 
 FactoryBot.define do
   factory :applicant_data_for_matching do
-    association :application
-    association :applicant
+    association :application, factory: :application
+    association :applicant, factory: :applicant
 
     program { "Computer Science" }
     department { "Arts and Science" }
--- a/api/spec/factories/applicant_data_for_matching.rb
+++ b/api/spec/factories/applicant_data_for_matching.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :applicant_data_for_matching do
+    association :application
+    association :applicant
+
+    program { "Computer Science" }
+    department { "Arts and Science" }
+    previous_uoft_ta_experience { "I have none" }
+    yip { 3 }
+    annotation { "This is an annotation" }
+  end
+end
--- a/api/spec/factories/application.rb
+++ b/api/spec/factories/application.rb
@@ -2,7 +2,7 @@
 
 FactoryBot.define do
   factory :application do
-    association :session
+    association :session, factory: :session
     comments {"Hello there"}
   end
 end
--- a/api/spec/factories/application.rb
+++ b/api/spec/factories/application.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :application do
+    association :session
+    comments {"Hello there"}
+  end
+end
+
--- a/api/spec/factories/instructor.rb
+++ b/api/spec/factories/instructor.rb
@@ -1,10 +1,19 @@
 # frozen_string_literal: true
 
+# Note that since instructor is in a has many relationship with positions, you should manually create
+# them in conjunction in any spec
 FactoryBot.define do
   factory :instructor do
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
     email { Faker::Internet.email("#{first_name} #{last_name}", '') }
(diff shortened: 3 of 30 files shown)


api/spec/factories/applicant_data_for_matching.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :applicant_data_for_matching do
+    association :application, factory: :application
+    association :applicant, factory: :applicant
+
+    program { "Computer Science" }
+    department { "Arts and Science" }
+    previous_uoft_ta_experience { "I have none" }
```
siefkenj, line 10: Should be a bool

dingtaoliu: @George334456 can you add a migration that changes this field to bool in the db?

api/spec/factories/application.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :application do
+    association :session, factory: :session
```
dingtaoliu, line 5: are associations defined only for belongs_to relations?

api/spec/factories/instructor.rb
```diff
@@ -1,10 +1,19 @@
...
+# them in conjunction in any spec
 FactoryBot.define do
   factory :instructor do
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
     email { Faker::Internet.email("#{first_name} #{last_name}", '') }
     utorid { Faker::Internet.slug("#{last_name} #{first_name} #{Faker::Number.number(2)}", '') }
+
+    factory :instructor_with_positions do
+      after(:build) do |instructor, evaluatlor|
```
dingtaoliu, line 13: Should we use `after(:create)` instead of `after(:build)` for database validations as well as model validations? 
 Remove evaluatlor or replace it with '_' if it is not being used.

api/spec/factories/position.rb
```diff
@@ -3,14 +3,19 @@
...
+    
+    position_code { 'CSC148' }
+    position_title { 'Introduction to Computer Science' }
+    est_hours_per_assignment { 60.00 }
+    est_start_date { Time.now }
+    est_end_date { Time.now + 120.days }
+    position_type { "Some role" }
+    
+    factory :position_with_instructors do
+      after(:build) do |position, evaluator|
```
dingtaoliu, line 15: Remove evaluator if it is not used.

api/spec/factories/session.rb
```diff
@@ -5,25 +5,32 @@
   sequence(:year, Time.now.year)
```
dingtaoliu, line 5: If we generate a session with the summer trait, the name would be "Summer #{current_year}", but if we then want to generate a fall session in the same year, would the sequence force the year to be current_year + 1?

api/spec/models/applicant_spec.rb
```diff
@@ -2,45 +2,52 @@
 
 describe Applicant do
   it 'should have a valid factory' do
-    FactoryBot.create(:applicant)
+    a = FactoryBot.create(:applicant)
   end
 
   it 'should not be valid without a first name' do
+    skip "To do. Old tests are commented so fix for new"
```
dingtaoliu, line 9: Are these cases work for another issue/PR?

michellemtchai:
I noticed that only some of the mode_spec files have those schema model information in the comment.   Do you mind running `docker-compose run api bundle exec annotate -p bottom` to make everything consistent? I only noticed the bool thing both @siefkenj  and @dingtaoliu have only mentioned. Otherwise, everything looks fine to me.

api/spec/factories/application.rb
George334456: It is defined for basically "parent" associations. Simply put, whenever we want to create an object after this object is created, we use `association`.

api/spec/factories/instructor.rb
George334456: I use after build here to build other objects, and then save those changes when create finishes.

I'll remove evaluator.

api/spec/factories/session.rb
George334456: Using the sequence, yes it will generate to curr_year + 1.

However, we can override the attributes in specs.

api/spec/models/applicant_spec.rb
George334456: Yeah I'll add an issue to do.

siefkenj:
Lot's of merge conflicts now. I propose closing unless someone wants to fix this.

PR #202, opened 2019-08-06
TAPP-191: Added ControlPanel view to CP 

dingtaoliu:
This PR adds a Control Panel view to the Contract Presentment part of TAPP.  The Control Panel view contains a filterable list of offers/assignments, and serves as a wrapper for additional connected components that perform actions on offers/assignments.  

Currently, the Control Panel uses a position-list as a placeholder until the offer/assignment list is completed.  Future work includes the list renderer for the assignment/offers, api actions for emailing offers to applicants, importing and exporting offers, and withdrawing offers.


siefkenj:
The title of the PR should be short and it should include a description which explains what it is.

frontend/src/components/filtered-list.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export function FilteredList(props) {
```
siefkenj, line 17: Add a jsdoc explaining this function.

This component is also sufficiently high-level that you should include propTypes. Look at https://blog.logrocket.com/validating-react-component-props-with-prop-types-ef14b29963fc/

frontend/src/components/positions-list.js
```diff
@@ -3,15 +3,15 @@ import PropTypes from "prop-types";
 
 export class PositionsList extends React.Component {
     static propTypes = {
-        positions: PropTypes.arrayOf(
+        data: PropTypes.arrayOf(
             PropTypes.shape({
                 position_code: PropTypes.string,
                 position_title: PropTypes.string
             })
```
siefkenj, line 10: I think this is overly restrictive.

frontend/src/modules/navigation/components/Header.jsx
```diff
@@ -31,7 +31,11 @@ const TappNavItems = () => (
     </Nav>
 );
 
-const CpNavItems = () => <Nav className="mr-auto" />;
+const CpNavItems = () => (
```
siefkenj: We have `CP` and `Control Panel`. This is a really confusing naming convention. If it's not exported, don't give it an overly specific name.

frontend/src/views/cp_control_panel/ControlPanel.jsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import { PositionsList } from "../../components/positions-list";
+import { FilteredList } from "../../components/filtered-list";
+import { connect } from "react-redux";
+import { positionsSelector } from "../../api/actions";
+import { ConnectedSessionSelect } from "../sessions/ConnectedSessionSelector";
+
+function ControlPanel() {
+    const ConnectedFilteredList = connect((state, ownProps) => ({
```
siefkenj: Connect outside this function. Otherwise, every time the component re-renders it will be re`connect`ed...

frontend/src/components/positions-list.js
dingtaoliu: We need to validate that the positions have the relevant fields before they are rendered in the table, do you think this is not the right place to do this validation?  Should I change it to `data: PropTypes.arrayOf( PropTypes.any)`?

frontend/src/modules/navigation/components/Header.jsx
dingtaoliu: Would it make sense if I changed `Control Panel` to `Offers` since its main purpose is to handle creation/management of offers?

siefkenj: Remove the `Cp` from `CpNavItems`

frontend/src/components/positions-list.js
siefkenj: I guess this is fine here. I mistook `shape` for `exact`.

frontend/src/components/filtered-list.js
```diff
@@ -0,0 +1,39 @@
...
+                onChange={e => setQuery(e.target.value)}
+            />
+            <Renderer data={filteredData} {...rest} />
+        </div>
+    );
+}
+
+FilteredList.propTypes = {
+    data: PropTypes.arrayOf(PropTypes.any),
+    listRenderer: PropTypes.func
```
siefkenj: `PropTypes.elementType`

Issue #204, opened 2019-08-11
[Backend] Fix Server hang


SimonAayani:
Currently, when making multiple concurrent API calls, the server gets stuck and the calls do not get carried out. 

Need to find a solution which eliminates the sever hang.

siefkenj:
This is resolved, right?

PR #205, opened 2019-08-11
Issue #204 Added fix for server hang when making concurrent API calls

SimonAayani:
Noticed that when I was trying to run my API calls (Specifically in the session routes) the server would hang and it would not populate the data properly. Forcing shut down and doing docker-compose up again would yield different results without having changed any code. I.e. sometimes the Positions view on the dashboard would populate, sometimes multiple views, and sometimes none. 

I fixed this problem by setting: 

`config.eager_load = true`

in /api/config/environments/development.rb

### DIFF
--- a/api/config/environments/development.rb
+++ b/api/config/environments/development.rb
@@ -9,7 +9,7 @@ Rails.application.configure do
   config.cache_classes = false
 
   # Do not eager load code on boot.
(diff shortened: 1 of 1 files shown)


George334456 (commented):
Hmm, if this is a problem in development, wouldn't there be issues in production for this too?

PR #214, opened 2019-08-21
TAPP-207: added Offer controller actions

dingtaoliu:
This PR adds active_offer_id to the assignment table, as well as several callback functions to maintain this field.  The active_offer_id of the assignment is modified by two actions: 

- when a new offer is created from an assignment, the assignment's active_offer_id is set to the id     of the offer 

- when the offer of an assignment is withdrawn, the active_offer_id is set to nil

To maintain the invariant that an assignment's active_offer only points to the latest offer that is insync with the assignment, we need to be careful about when and how we perform actions such as generating new offers, modifying assignments, withdrawing offers, and so on.  

This PR will remain WIP until the details are ironed out and implemented.


siefkenj:
I looks like you might need to rebase. A lot of michelle's changes are in there...

dingtaoliu:
In addition to adding an `active_offer_id` field to `Assignment`, this PR also adds controller actions for `Offer`, which includes the following actions:

- `create`: creates an `Offer` from an `Assignment`

- `show`: find and render an `Offer` in json format using its `url_token` 

- `active_offer`: render the `active_offer` of an `Assignment` if it exists

- `email_offer`: enqueue a `contract_email` job to be performed by an ActiveJob worker

- `withdraw_offer`: withdraw the `active_offer` of an `Assignment` 

- `reject_offer`: reject an `Offer` 

- `accept_offer`: accept an `Offer`

- `nag`: enqueue a `nag_email` job to be performed by an ActiveJob worker

The ActiveJob engine is currently set to be `:async`, which uses the in-process thread pool.  This means that if the server crashes or is restarted, the queued job data does not persist.  If we continue to use the ActiveJob interface for queueing and sending emails, we should look into other possible engines such as Sidekiq or DelayedJob.  

This PR also adds a light `MailCatcher` container to the development environment.  In dev mode, the emails are configured to be sent to the `MailCatcher` port, which provides a gui interface to examining the emails (which can include pdf attachments).  Once the docker containers spin up, simply visit `localhost:1080` to access `MailCatcher`.

George334456 (approved):
Looks good

Issue #203, opened 2019-08-07
[Frontend] Refactor New Position Form into Component and View


thezane:
Split up new position form into component for defining page structure and view for accessing redux store.

PR #227, opened 2019-09-20
Tapp 203

thezane:
Split up new position form into component for defining page structure and view for accessing redux store.  https://github.com/uoft-tapp/tapp/issues/203


frontend/src/views/new_position_form/reducer.js
```diff
@@ -0,0 +1,30 @@
+import { createReducer } from "redux-create-reducer";
+import {
+    CREATE_NEW_POSITION_SUCCESS,
+    IMPORT_NEW_POSITION_FAILURE,
+    IMPORT_NEW_POSITION_SUCCESS
+} from "../../api/constants/new_position";
+
+const initialState = {
+    list: [],
```
siefkenj: `list` is not a good name...what is it a list of?

frontend/src/components/new-position-form.js
```diff
@@ -0,0 +1,151 @@
+import React from "react";
+import { moment } from "moment";
+
+const toNames = ({ first_name, last_name }) => `${first_name} ${last_name}`;
+
+const validNumber = val => (val && isNaN(val) ? "Not a valid number" : false);
+const validDate = val =>
+    (val && moment(val, "YYYY-MM-DD", true).isValid()) || val === ""
+        ? false
+        : "Not a valid date";
```
siefkenj: What?? `validDate === false` means that the date is valid? Don't mix variables that store booleans and strings. If it is supposed to be an error message, make it an error message.

frontend/src/components/new-position-form.js
```diff
@@ -0,0 +1,151 @@
...
+// check for format XXX123H1
+const validCourseCode = val => {
```
siefkenj: No. Don't do this check. The backend will say whether a course code is valid. If it exists in the database, it's valid.

frontend/src/components/new-position-form.js
```diff
@@ -0,0 +1,151 @@
...
+        !isNaN(val.slice(3, 6)) &&
+        (val[6] === "H" || val[6] === "Y") &&
+        !isNaN(val[7])
+    ) {
+        return false;
+    } else {
+        return "Not a valid course code";
+    }
+};
+const validInstructor = (val, props) => {
```
siefkenj: You're creating needless datastructures...This function should also be called `existingInstructor`. Any of these functions that might be used more than once should be in the `utils` directory.

frontend/src/components/new-position-form.js
```diff
@@ -0,0 +1,151 @@
...
+    const instructors = props.instructors.map(toNames);
+    return instructors.indexOf(val) !== -1 ? false : "Not a valid instructor";
+};
+
+export const newPositionFields = [
+    {
+        label: "Course Code",
+        value: "course_code",
+        validate: [validCourseCode],
```
siefkenj: Why are the validators an array? Do we need to chain validators?

frontend/src/components/new-position-form.js
```diff
@@ -0,0 +1,151 @@
...
+        <label style={{ width: "100px" }}>
+            {required ? label + "*" : label}
+        </label>
+        <input value={curValue} onChange={onChange} required={required} />
+    </div>
+);
+
+export const TextboxInput = ({ label, curValue, onChange, required }) => (
+    <div>
+        <label style={{ width: "100px" }}>
```
siefkenj, line 117: Don't do inline styles.

frontend/src/components/new-position-form.js
```diff
@@ -0,0 +1,151 @@
...
+    </div>
+);
+
+export const InstructorList = ({ curValue, instructors, setInstructor }) => {
+    const valid = validInstructor(curValue, { instructors });
+    const matchingInstructors = instructors
+        .map(({ first_name, last_name }) => `${first_name} ${last_name}`)
+        .filter(
+            name => name.toLowerCase().indexOf(curValue.toLowerCase()) !== -1
+        );
```
siefkenj: This confuses me...how could there me more than one matching instructor?

frontend/src/api/actions/new_position.js
```diff
@@ -0,0 +1,47 @@
...
+    IMPORT_NEW_POSITION_FAILURE,
+    IMPORT_NEW_POSITION_SUCCESS
+} from "../constants/new_position";
+import { apiPOST } from "../../libs/apiUtils";
+
+export const createNewPosition = payload => async dispatch => {
+    try {
+        const data = await apiPOST("/positions", payload);
+        dispatch(createNewPositionSuccess(data));
+        window.location = "/tapp/positions";
```
siefkenj: Never, never, never set `window.location` yourself. API actions should be side-effect free. That is, they can change the model state, but nothing else (no changing of routes, no changing of UI state)

siefkenj: When you dispatch something in the UI, you can `await` the dispatch. That is the appropriate place to change the route.

siefkenj:
Put the details in this PR. It's no fun to have to click around links to see what is supposed to be happening...

frontend/src/components/new-position-form.js
thezane: The styling for this page will be changed in a different issue.  When that happens, inline styles will be removed.

siefkenj:
Merge conflict now...

frontend/src/views/new_position_form/reducer.js
thezane: Renamed

frontend/src/components/new-position-form.js
thezane: A validator now returns true iff a field is valid.  If a field is invalid, an err msg is accumulated in `fillablePosition.jsx`.

frontend/src/components/new-position-form.js
thezane: Removed check

frontend/src/components/new-position-form.js
thezane: Replaced array of validators with single validator

frontend/src/components/new-position-form.js
thezane: Removed constructed list via map.  Renamed function.

frontend/src/api/actions/new_position.js
thezane: Moved route changing to view

frontend/src/components/new-position-form.js
thezane: The map & filter code is part of an autocompletion feature that lists instructors whose name is a superstring of what the user typed.  So yes, more than 1 instructor can survive the filter.  I rewrote it so we only have 1 loop.

Screenshot from 2019-10-10 21-28-19

siefkenj:
Looks good! I'll merge and any remaining issues can be fixed later.

PR #265, opened 2019-11-10
adjustment on mockAPI

CarolineYao:
* added `positionPreference` (this name is from our schema table, might want to change this to `applicantPreference` or `applicationPreference`, not sure which one is better) to `position` for applicants' preference level on that position
* added `instructorPreference` to `position` in order to indicate instructors' preference level for any applicant
* reformatted `applicationSelector` to include `applicant` and `position` objects

### DIFF
--- a/frontend/src/api/actions/applications.js
+++ b/frontend/src/api/actions/applications.js
@@ -8,13 +8,18 @@ import {
 import { fetchError, upsertError, deleteError } from "./errors";
 import {
     actionFactory,
+    arrayToHash,
     runOnActiveSessionChange,
-    validatedApiDispatcher
+    validatedApiDispatcher,
+    flattenIdFactory,
+    splitObjByProps
 } from "./utils";
 import { apiGET, apiPOST } from "../../libs/apiUtils";
 import { applicationsReducer } from "../reducers/applications";
 import { createSelector } from "reselect";
+import { applicantsSelector } from "./applicants";
 import { activeRoleSelector } from "./users";
+import { positionsSelector } from "./positions";
 
 // actions
 const fetchApplicationsSuccess = actionFactory(FETCH_APPLICATIONS_SUCCESS);
@@ -25,6 +30,20 @@ const upsertOneApplicationSuccess = actionFactory(
 const deleteOneApplicationSuccess = actionFactory(
     DELETE_ONE_APPLICATION_SUCCESS
 );
+
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+const positionToPositionId = flattenIdFactory("position", "position_id");
+function prepForApi(data) {
+    const [ret, filtered] = splitObjByProps(data, ["position_preferences"]);
+
+    if (filtered["position_preferences"]) {
+        ret["position_preferences"] = filtered[
+            "position_preferences"
+        ].map(preference => positionToPositionId(preference));
+    }
+
+    return applicantToApplicantId(ret);
+}
 
 // dispatchers
 export const fetchApplications = validatedApiDispatcher({
@@ -66,7 +85,7 @@ export const upsertApplication = validatedApiDispatcher({
         const { id: activeSessionId } = getState().model.sessions.activeSession;
         const data = await apiPOST(
             `/${role}/sessions/${activeSessionId}/applications`,
-            payload
+            prepForApi(payload)
         );
         dispatch(upsertOneApplicationSuccess(data));
     }
@@ -82,7 +101,7 @@ export const deleteApplication = validatedApiDispatcher({
         const { id: activeSessionId } = getState().model.sessions.activeSession;
         const data = await apiPOST(
             `/${role}/sessions/${activeSessionId}/applications/delete`,
-            payload
+            prepForApi(payload)
         );
         dispatch(deleteOneApplicationSuccess(data));
     }
@@ -95,9 +114,39 @@ export const deleteApplication = validatedApiDispatcher({
 // search for and return the isolated state associated with `reducer`. This is not
 // a standard redux function.
 export const localStoreSelector = applicationsReducer._localStoreSelector;
-export const applicationsSelector = createSelector(
+export const _applicationsSelector = createSelector(
     localStoreSelector,
     state => state._modelData
+);
+
+// Get the current list of applications and recompute `applicant_id` and `position_id`
+// to have corresponding `applicant` and `position` objects
+export const applicationsSelector = createSelector(
+    [_applicationsSelector, applicantsSelector, positionsSelector],
+    (applications, applicants, positions) => {
+        if (applications.length === 0) {
+            return [];
+        }
+
+        const applicantsById = arrayToHash(applicants);
+        const positionsById = arrayToHash(positions);
+
+        // Change `applicant_id` to the corresponding `applicant` object
+        // and similarly, change each `position_id` in each entry of
+        // `position_preferences` to corresponding `position` object.
+        return applications.map(
+            ({ position_preferences, applicant_id, ...rest }) => ({
+                ...rest,
+                applicant: applicantsById[applicant_id] || {},
+                position_preferences: (position_preferences || []).map(
+                    ({ position_id, ...rest }) => ({
+                        position: positionsById[position_id],
+                        ...rest
+                    })
+                )
+            })
+        );
+    }
 );
 
 // Any time the active session changes, we want to refetch
--- a/frontend/src/api/actions/positions.js
+++ b/frontend/src/api/actions/positions.js
@@ -11,7 +11,8 @@ import {
     runOnActiveSessionChange,
     validatedApiDispatcher,
     arrayToHash,
-    flattenIdFactory
+    flattenIdFactory,
+    splitObjByProps
 } from "./utils";
 import { apiGET, apiPOST } from "../../libs/apiUtils";
 import { positionsReducer } from "../reducers/positions";
@@ -19,6 +20,7 @@ import { createSelector } from "reselect";
 import { instructorsSelector } from "./instructors";
 import { contractTemplatesSelector } from "./contract_templates";
 import { activeRoleSelector } from "./users";
+import { applicantsSelector } from "./applicants";
 
 // actions
 const fetchPositionsSuccess = actionFactory(FETCH_POSITIONS_SUCCESS);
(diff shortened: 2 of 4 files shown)


frontend/src/api/defs/prop-types.js
```diff
@@ -59,7 +59,21 @@ function generatePropTypes(PropTypes) {
...
+                    preference_level: PropTypes.number,
+                    applicant_id: id,
+                    instructor_id: id
+                })
+            ),
+            positionPreference: PropTypes.arrayOf(
+                PropTypes.shape({
+                    preference_level: PropTypes.number,
+                    application_id: id,
+                    instructor_id: id
```
siefkenj: Why is there an `instructor_id` here?

frontend/src/api/mockAPI/data.js
```diff
@@ -98,7 +98,48 @@ export const mockData = {
             desired_num_assignments: 15,
             current_enrollment: 1200,
             current_waitlisted: 200,
-            instructor_ids: [1000, 1001]
+            instructor_ids: [1000, 1001],
+            instructorPreference: [
```
siefkenj: We use snake case in the api data. This should also be plural since it's a list.

frontend/src/api/mockAPI/data.js
```diff
@@ -98,7 +98,48 @@ export const mockData = {
...
+                    applicant_id: 2002,
+                    instructor_id: 1001
+                },
+                {
+                    preference_level: 1,
+                    applicant_id: 2006,
+                    instructor_id: 1001
+                }
+            ],
+            positionPreference: [
```
siefkenj: This should be `applicant_preferences` or `application_preferences` (depending on what Karen thinks)

frontend/src/api/defs/prop-types.js
CarolineYao: Opps, sorry, totally misread the schema

frontend/src/api/actions/applications.js
```diff
@@ -88,11 +91,36 @@ export const deleteApplication = validatedApiDispatcher({
...
+    (applications, applicants, positions) => {
+        if (applications.length === 0) {
+            return [];
+        }
+
+        const applicantsById = arrayToHash(applicants);
+        const positionsById = arrayToHash(positions);
+
+        return applications.map(
+            ({ application_preferences, applicant_id, ...rest }) => ({
```
siefkenj: These are `position_preferences`. In general, I don't like replacing a field with a recomputed one in a selector. I think it's better that they have different names. For example, replacing `applicant_id` with `applicant` is fine because the fields have different names. This case kinda walks the line, since you're modifying what's inside the preferences...

siefkenj (commented):
I like the way this is looking.

frontend/src/api/actions/applications.js
CarolineYao: Oh, when you said "it's an application's preferences for positions", I thought you meant it is ok to keep it as `application_preferences`. Not sure if you meant we should only name it as `position_preferences` in here but keep it as `application_preferences` in the API and mock data.

siefkenj: If it's in the application, it should be `position_preferences`

CarolineYao: Also, not sure how we feel about this, but I think it might be a good idea if we recompute each object key from snake case to camel case (probably write a helper function and call that function in `createSelector`):
1. Better coding style as it will mostly have only camel cases in each component
2. As we are centralizing the model data processing, when each individual component receive its' `props`, it should not concern/know whether or not the data is coming from API.
3. It will also help with the situation here. The selector will then have `applicationPreferences ` instead of `application_preferences` so it is easier to separate them.

Of course, that also mean we need to recompute it to use snake case before we send it to API, but I think it is a good practice in general unless there is some other reason why we chose do it the way it is right now

siefkenj: I don't like this idea, but we can talk about it Friday.

frontend/src/api/actions/positions.js
```diff
@@ -58,12 +59,28 @@ const instructorsToInstructorIds = flattenIdFactory(
...
+const instructorToInstructorId = flattenIdFactory(
+    "instructor",
+    "instructor_id"
+);
+
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+
 function prepForApi(data) {
-    return contractTemplateToInstructorIds(instructorsToInstructorIds(data));
+    data.instructor_preferences = (
```
siefkenj: Don't mutate the incoming data...

If `instructor_preferences` is null, we want to leave it as null, lest we accidentally delete the instructor preferences by sending a blank array.

frontend/src/api/actions/applications.js
```diff
@@ -88,11 +102,36 @@ export const deleteApplication = validatedApiDispatcher({
 // search for and return the isolated state associated with `reducer`. This is not
 // a standard redux function.
 export const localStoreSelector = applicationsReducer._localStoreSelector;
-export const applicationsSelector = createSelector(
+export const _applicationsSelector = createSelector(
     localStoreSelector,
     state => state._modelData
 );
 
+export const applicationsSelector = createSelector(
```
siefkenj, line 124: Add a comment explaining the special stuff this selector does

frontend/src/api/actions/applications.js
```diff
@@ -25,6 +29,16 @@ const deleteOneApplicationSuccess = actionFactory(
     DELETE_ONE_APPLICATION_SUCCESS
 );
 
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+const positionToPositionId = flattenIdFactory("position", "position_id");
+function prepForApi(data) {
+    data.position_preferences = (
+        data.position_preferences || []
+    ).map(preference => positionToPositionId(preference));
```
siefkenj: If `position_preferences` is null, we want to keep it as null

frontend/src/api/actions/applications.js
```diff
@@ -25,6 +30,23 @@ const deleteOneApplicationSuccess = actionFactory(
...
 );
 
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+const positionToPositionId = flattenIdFactory("position", "position_id");
+function prepForApi(data) {
+    const propName = "position_preferences";
+    const [ret, filtered] = splitObjByProps(data, [propName]);
+
+    if (filtered.length <= 0) {
+        return data;
```
siefkenj: This doesn't make sense to me...we should always have an `applicant` that needs to be flattened.

frontend/src/api/actions/applications.js
```diff
@@ -25,6 +30,23 @@ const deleteOneApplicationSuccess = actionFactory(
...
+    }
+
+    ret[propName] = filtered[propName].map(preference =>
+        positionToPositionId(preference)
+    );
```
siefkenj: There is no harm mapping on an empty array.

frontend/src/api/actions/positions.js
```diff
@@ -58,12 +60,33 @@ const instructorsToInstructorIds = flattenIdFactory(
...
+
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+
 function prepForApi(data) {
-    return contractTemplateToInstructorIds(instructorsToInstructorIds(data));
+    const propName = "instructor_preferences";
+    const [ret, filtered] = splitObjByProps(data, [propName]);
+
+    if (filtered.length <= 0) {
+        return data;
```
siefkenj: Again, I don't think this is right

frontend/src/api/actions/applications.js
CarolineYao: Good point, I'll fix this

CarolineYao: So, I changed the if condition to be the opposite: `if (filtered[propName].length > 0)`. Put the mapping of `position_preference` inside the `if` scope so that it won't try to change anything when `position_preference` is `null`. Same for the `instructor_preference`.

I am assuming that we do not need to consider the case where somehow `position_preference` is not returned from the API?

frontend/src/api/actions/applications.js
```diff
@@ -25,6 +30,21 @@ const deleteOneApplicationSuccess = actionFactory(
     DELETE_ONE_APPLICATION_SUCCESS
 );
 
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+const positionToPositionId = flattenIdFactory("position", "position_id");
+function prepForApi(data) {
+    const propName = "position_preferences";
+    const [ret, filtered] = splitObjByProps(data, [propName]);
+
+    if (filtered[propName].length > 0) {
```
CarolineYao: If it is `null`, will still error out, change to if it exist

frontend/src/api/actions/applications.js
```diff
@@ -25,6 +30,21 @@ const deleteOneApplicationSuccess = actionFactory(
     DELETE_ONE_APPLICATION_SUCCESS
 );
 
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+const positionToPositionId = flattenIdFactory("position", "position_id");
+function prepForApi(data) {
+    const propName = "position_preferences";
```
CarolineYao: Just use "position _preference"

frontend/src/views/applicants_by_position/index.js
```diff
@@ -0,0 +1,53 @@
...
+                <Col xs={2}>
+                    <ConnectedSelectPosition
+                        selectedPositions={selectedPositions}
+                        onSelectPosition={position => {
+                            setSelectedPositionsFromChild(stateProps, position);
+                            debugger;
+                        }}
+                    />
+                </Col>
+                {/* {props.selectedPositionIds.map(positionId => (
```
siefkenj: Don't keep commented out code around

frontend/src/views/applicants_by_position/index.js
```diff
@@ -0,0 +1,53 @@
...
+        if (newList.includes(newlySelectedPostion)) {
+            result = newList.splice(newList.indexOf(newlySelectedPostion), 1);
+        } else {
+            newList.push(newlySelectedPostion);
+            result = newList.slice(-2);
+        }
+        result = newList.splice(-2, 0);
+
+        setSelectedPositions(result);
+        debugger;
```
siefkenj: Remove the debugger statements

frontend/src/api/actions/applications.js
```diff
@@ -26,6 +31,20 @@ const deleteOneApplicationSuccess = actionFactory(
...
 );
 
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+const positionToPositionId = flattenIdFactory("position", "position_id");
+function prepForApi(data) {
+    const [ret, filtered] = splitObjByProps(data, ["position_preferences"]);
+
+    if (filtered["position_preferences"]) {
+        ret["position_preferences"] = (
+            filtered["position_preferences"] || []
```
siefkenj: The `|| []` is redundant. You already checked that it is truthy, so it will never fall back to the `[]`

frontend/src/api/actions/positions.js
```diff
@@ -63,12 +65,32 @@ const instructorsToInstructorIds = flattenIdFactory(
...
+
+const applicantToApplicantId = flattenIdFactory("applicant", "applicant_id");
+
 function prepForApi(data) {
-    return contractTemplateToInstructorIds(instructorsToInstructorIds(data));
+    const [ret, filtered] = splitObjByProps(data, ["instructor_preferences"]);
+
+    if (filtered["instructor_preferences"]) {
+        ret["instructor_preferences"] = (
+            filtered["instructor_preferences"] || []
```
siefkenj: Same comment again

frontend/src/api/actions/applications.js
CarolineYao: Right, I didn't remove it because I think it wouldn't hurt to keep it here. But I guess it makes sense to remove code that will never reach.

Issue #228, opened 2019-09-22
[Backend] Refactoring to best practices


tsimica:
As discussed with @reidka, the Rails codebase needs to be refactored to adhere to best Rails practices. Some of the work would include (but not limited to):

**General**
- [ ] Bringing the codebase to use Rails 6. It's already being used in production at GitHub and Basecamp -- might as well upgrade to it. 
- [ ] Gemfile can be refactored. Lots of unused gems can be removed. Gems that are being used should be versioned. 
- [ ] Puma should be upgraded to 4.1. Perf testing should be done to know how many threads should be used in production. 
- [ ] Styling should use 2 spaces, not sure why it was changed to 4. Controller classes should be Api::V1::ControllerName < ApplicationController. Follow Airbnb or Shopify conventions. 
- [ ] Authorization is non-existent. Pundit (gem) can help here. This can be a separate issue. 
- [ ] Background processing for emails & PDF generation. Consider Sidekiq or an alternative. Redis looks to be used in production but not development? Use it also in development to be consistent. You can use foreman to manage the processes running.  

**Rails Specific**
`config/routes.rb`
- [ ] Refactoring routes to match Rails conventions (e.g. "delete" is not a Rails verb, it's "destroy")
`app/controllers/concerns/Response.rb` 
- [ ] params are never needed to be casted to Integer, Rails does this automatically
- [ ] There are no status codes besides 200 returned by the api. `render json: {...}, status: :bad_request` should be used in `render_error`

`app/controllers/api/v1/applications_controller.rb`
- [ ] No use of `before_action` anywhere. `invalid_id(Session, :session_id)`, `Application.find(params[:id])` appear twice. These should be refactored.
- [ ] Using `model.destroy!` after a save is invalid raises serious concerns -- is this something that is explicitly needed? The database and model validations should enforce constraints. 
- [ ] Update route should be a transaction -- either both succeed or both fail. If one succeeds but the other fails, it's treated as an error in the front-end but not in the back-end. This will lead to stale and inconsistent data. 
- [ ] `all_applications` and `application_data` should be refactored into the model. If you're trying to include an associated model, consider using `.includes` then `.to_json(includes: X)`
- [ ] `applications_by_session` should also use ActiveRecord methods instead of Ruby methods
- [ ] Why use `.exists?` at all when `.find` can be used -- you get the object for free and can use the methods on it 

Controllers not mentioned fall into some issue mentioned above, or they are yet to be posted here. 

[...]

PR #273, opened 2019-11-29
[WIP] #228 - New Rails Backend

tsimica:


siefkenj:
Modify the rubocop rules in a suitable way for your code to pass. Also, it says `db:create` fails...Does the .travis.yml file need to be changed?

api/app/services/assignment_wage_chunk_create_service.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def upsert_and_return_wage_chunks
+        @assignment.wage_chunks
+                   .upsert_all(@wage_chunks,
+                               returning: RETURNING_VALUES)
+    end
+
+    class WageChunkMissing < StandardError
+        def message
+            I18n.t('assignment_wage_chunk_service.wage_chunk_missing.message')
```
siefkenj, line 50: We're internationalizing our messages?

api/config/cable.yml
```diff
@@ -2,9 +2,9 @@ development:
   adapter: async
 
 test:
-  adapter: async
+  adapter: test
 
 production:
   adapter: redis
-  url: redis://localhost:6379/1
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
```
siefkenj: What's with the hard-coded URL and port?

api/config/routes.rb
```diff
@@ -1,102 +1,91 @@
...
+
+                # Application
+                resources :applications, only: :create
+
+                # Assignments
+                resources :assignments, only: %i[show create] do
+                    resources :wage_chunks, controller: :assignment_wage_chunks, only: %i[index create]
+                    resources :offers, path: :active_offer, controller: :active_offers, only: :index do
+                        collection do
+                            post 'create'
```
siefkenj, line 21: Why is `'create'` a string while the rest are identifiers?

siefkenj (commented):
Let's disable the failing tests for now, and we'll reactivate them in a new PR. (Theoretically, they should work if we prefix all the routes with `/admin`)

api/config/cable.yml
tsimica: This was autogenerated. We don't need ActionCable for our use cases. I will remove it in a future commit.

api/config/routes.rb
tsimica: `create` has special meaning in Rails and will give you a different route depending on whether it's a symbol or string. Defining `create` as a symbol will give you `POST /api/v1/offers/` vs. a string `POST api/v1/offers/create`

PR #276, opened 2019-12-10
Move session selector to where active user display is, and add dropdown to change both fields

CarolineYao:


frontend/src/api/reducers/sessions.js
```diff
@@ -9,7 +9,7 @@ import { createBasicReducerObject, createReducer } from "./utils";
 
 const initialState = {
     _modelData: [],
-    activeSession: { id: null }
+    activeSession: { id: null, name: "" }
```
siefkenj: Is there not an active session selector? I don't think the name should be stored in multiple places...

frontend/src/components/active-user.js
```diff
@@ -1,20 +1,61 @@
...
+                    setDropdownVisible(desiredVisibility)
+                }
+                show={dropdownVisible}
+            >
+                <Dropdown.Toggle split variant="light">
+                    {label}
+                </Dropdown.Toggle>
+                <Dropdown.Menu>
+                    {(roles || []).map((role, index) => {
+                        return (
```
siefkenj: You can use an implicit return here instead of a return statement.

frontend/src/components/header.js
```diff
@@ -95,7 +95,9 @@ export function Header(props) {
                     </Route>
                 ))}
             </Nav>
-            {infoComponent}
+            {infoComponents.map((component, index) => {
+                return <div key={index}>{component}</div>;
```
siefkenj: Use an implicit return unless an actual return helps readability.

frontend/src/views/admin/header.js
```diff
@@ -71,14 +78,40 @@ function AdminHeader() {
...
+
+const mapSessionsDispatchToProps = { setActiveSession };
+
+const ConnectedActiveSessionDisplay = connect(
+    mapSessionsStateToProps,
+    mapSessionsDispatchToProps
+)(ActiveSessionDisplay);
+
+const mapActiveUserStateToProps = state => {
+    return {
```
siefkenj: Implicit return

frontend/src/api/reducers/sessions.js
CarolineYao: I agree that that it should not be stored in multiple places. The reason why I added this here is because there was a warning complaining about missing `activeSession.name` when active session is unset. That is due to the fact that `name` is a required attribute for session.  IMHO, it make sense to have the initial value contain all the required fields (maybe we can have a util function where it automatically create all the required fields, it can even have some flexibility where you can specify the value for some particular fields). 

I think instead of removing `name` here, it makes more sense to remove the ` = { id: null, name: "" }` of 
```
 const {
        sessions = [],
        activeSession = { id: null, name: "" },
        setActiveSession
    } = props;
```
in active session selector component as `activeSession` would not be empty; it would at least be the default value, and it has the name field as well.

CarolineYao: The proptype should null or session

PR #282, opened 2020-01-18
removed the comment in ./package.json

PeterSenpai:
removed the comment in ./package.json


siefkenj:
The build is broke for other reasons, so I will merge this PR.

PR #283, opened 2020-01-19
remove node_sass dependency that causes the build failure

PeterSenpai:


siefkenj:
This was superseded by PR #284

PR #287, opened 2020-01-24
Refactor `componentDidMount` to init function

CarolineYao:


frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+    // setActiveUserRole,
+    fetchActiveUser,
+    fetchSessions,
+    fetchApplicants,
+    fetchApplications,
+    fetchAssignments,
+    fetchContractTemplates,
+    fetchInstructors,
+    fetchPositions
+} from "./";
```
siefkenj: Don't do this kind of directory import. Import from the files instead.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+import { setGlobals, globalsSelector } from "../../views/globals/actions";
```
siefkenj: I don't like the idea of importing from views into `api/`. `api/` should be self-contained.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+
+const toggleMockApi = (getState, newMockAPIValue) => {
```
siefkenj: `newMockAPIValue` -> `enableMockAPI`

siefkenj: We shouldn't use arrow functions where they aren't needed/don't make the code more concise. Do `function toggleMockApi()` instead

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+const fetchAdditionalData = [
+    fetchApplicants,
+    fetchApplications,
+    fetchAssignments,
+    fetchContractTemplates,
+    fetchInstructors,
+    fetchPositions
+];
+
+export function initFromStage(options, stage = "toggleMockAPI") {
```
siefkenj: `stage` before `options`. Stage shouldn't have a default value. Options should (since it's optional). Also, this function needs a comment to explain what it does and what it's for.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+        {
+            stage: "setActiveUser",
+            action: fetchActiveUser,
+            variables: []
+        },
+        // {
+        //     stage: "setActiveUserRole",
+        //     action: setActiveUserRole,
+        //     variables: [getState, activeUserRole]
+        // },
```
siefkenj: No commented out code in commits

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+        globals = {
+            ...globals,
+            activeSession: newActiveSession.id
+        };
+    }
+
+    return setGlobals(globals);
+};
+
+const generateListOfActions = (getState, stage, options) => {
```
siefkenj: Make this a `function` function. This needs some comments.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+        });
+    }
+
+    onInitialization.push({
+        stage: "updateActiveSessionGlobal",
+        action: updateGlobals,
+        variables: [getState, mockAPI, activeSession]
+    });
+
+    while (onInitialization.length > 0) {
```
siefkenj: I don't like while loops (especially when for loops will do)

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+        if (newMockAPIValue === true) {
+            mockAPI.replaceGlobalFetch();
+        } else {
+            mockAPI.restoreGlobalFetch();
+        }
+    }
+    /* eslint-enable */
+};
+
+const updateGlobals = (getState, newMockAPIValue, newActiveSession) => {
```
siefkenj: Having an `updateGlobals` and a `setGlobals` confuses me. Let's think of better names.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+            stage: "setActiveSession",
+            action: setActiveSession,
+            variables: [activeSession]
+        });
+    }
+
+    onInitialization.push({
+        stage: "updateActiveSessionGlobal",
+        action: updateGlobals,
+        variables: [getState, mockAPI, activeSession]
```
siefkenj: This is too much abstraction. I think closures would be clearer. (e.g, pass `() => updateGlobals(getState, mockAPI, activeSession)` as the `action`).

frontend/src/api/actions/init.js
siefkenj: The order of arguments is strange. Look at where `getState` appears in other actions.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+        if (onInitialization[0].stage === stage) {
+            break;
+        }
+
+        onInitialization.shift();
+    }
+    return onInitialization;
+};
+
+const fetchAdditionalData = [
```
siefkenj: Why is this defined top-level?

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,150 @@
...
+        const promises = [];
+        for (const step of onInitialization) {
+            if (step.action) {
+                promises.push(dispatch(step.action(...step.variables)));
+            } else {
+                step.function(...step.variables);
+            }
+        }
+
+        for (const fetchAction of fetchAdditionalData) {
```
siefkenj: You're aiming at something too abstract. Make a big, long, ugly function that literally dispatches the actions in order.

frontend/src/api/actions/init.js
CarolineYao: Don't delete it. Make it a empty function. Also check if an action fails, does it only get error response or if it crashes.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,119 @@
...
+        globals = {
+            ...globals,
+            activeSession: newActiveSession.id
+        };
+    }
+
+    return setGlobals(globals);
+}
+
+export function initFromStage(stage, options) {
```
siefkenj, line 90: Function needs a docstring

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,119 @@
...
+    return async (dispatch, getState) => {
+        const onInitialization = [
```
siefkenj: `onIntialization` -> `initOrder`

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,119 @@
...
+
+        const {
+            activeSession = null,
+            activeUserRole = "admin",
+            mockAPI = null
+        } = options;
+
+        const startIndex = onInitialization.indexOf(stage);
+
+        if (startIndex <= onInitialization.indexOf("toggleMockAPI")) {
```
siefkenj: Instead of `startIndex` and testing it each time, make a `shouldRunStage(<string>) -><bool>` function that does the test. That should make the code intention a bit clearer.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,119 @@
...
+            activeSession !== null
+        ) {
+            // after sessions are fetched, we compare with the active session.
+            // The active session might need to be "updated" if the ID matches but
+            // the data doesn't
+            const state = getState();
+            const sessions = sessionsSelector(state);
+            const matchingSession = sessions.filter(
+                s => s.id === activeSession.id
+            )[0];
```
siefkenj: `matchingSessions` could be undefined...

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,119 @@
...
+
+            await dispatch(setActiveSession(matchingSession));
+        }
+
+        if (startIndex <= onInitialization.indexOf("updateGlobals")) {
+            await dispatch(updateGlobals(mockAPI, activeSession, getState));
+        }
+
+        const promises = [];
+        const fetchData = [
```
siefkenj: `fetchData` -> `fetchActions`

siefkenj (commented):
Getting close!

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * A helper function to determine if the `currentStage`
+ * should be run
+ *
+ * @param {string} fromStage
+ * @param {boolean} currentStage
+ * @returns
+ */
+function shouldRunStage(fromStage, currentStage) {
```
siefkenj: Why don't you move this to an inline function in `initFromStage` so that `fromStage` is defined via a closure?

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,154 @@
...
+    ];
+
+    const startIndex = initOrder.indexOf(fromStage);
+
+    return startIndex <= initOrder.indexOf(currentStage);
+}
+
+/**
+ * Perform different API actions that is necessary
+ * after the specified `stage`
```
siefkenj, line 83: Let's explain this a bit more.

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,154 @@
...
+
+            await dispatch(setActiveSession(matchingSession));
+        }
+
+        if (shouldRunStage(stage, "updateGlobals")) {
+            await dispatch(updateGlobals(mockAPI, activeSession, getState));
+        }
+
+        const promises = [];
+        const fetchActions = [
```
siefkenj, line 159: I think this array needs a comment

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,154 @@
...
+            fetchApplicants,
+            fetchApplications,
+            fetchAssignments,
+            fetchContractTemplates,
+            fetchInstructors,
+            fetchPositions
+        ];
+
+        for (const fetchAction of fetchActions) {
```
siefkenj: I think `fetchActions.map(action => dispatch(action()))` do.

siefkenj (commented):
We're soooooo close!

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,170 @@
...
+            activeSession = null,
+            activeUserRole = "admin",
+            mockAPI = null
+        } = options;
+
+        /**
+         * A helper function to determine if the `currentStage`
+         * should be run
+         *
+         * @param {boolean} currentStage
```
siefkenj: It's a string...the return is a bool

frontend/src/api/actions/init.js
```diff
@@ -0,0 +1,170 @@
...
+ * API to this level instead of leaving it to the mockAPI component. Whenever
+ * a related events occur (for example, changing active session), the action
+ * creator will call this `initFromStage` action with the matching stage and
+ * this function will perform different API actions that is necessary
+ * after the specified `stage`
+ *
+ * @export
+ * @param {string} stage
+ * @param {object} options
+ * @returns
```
siefkenj: Returns a promise? Explain what it returns. And, should this function be async?

CarolineYao: The function returns an async function similar to `validatedApiDispatcher`. It is used this way because we need the `dispatch` and `getState` and I have used `validataedApiDispatcher` as an example.

siefkenj: Great. Put that in the `@returns` part of the docstring

PR #295, opened 2020-02-17
Refactored the init system

siefkenj:
This PR adds an initial "pageLoad" init value to the init system as well as makes sure that session-dependent data is only fetched when there is an active session.

### DIFF
--- a/frontend/src/api/actions/init.js
+++ b/frontend/src/api/actions/init.js
@@ -1,5 +1,14 @@
-import { setActiveSession, fetchSessions, sessionsSelector } from "./sessions";
-import { fetchActiveUser, setActiveUserRole } from "./users";
+import {
+    setActiveSession,
+    fetchSessions,
+    sessionsSelector,
+    activeSessionSelector
+} from "./sessions";
+import {
+    fetchActiveUser,
+    setActiveUserRole,
+    activeRoleSelector
+} from "./users";
 import { fetchApplicants } from "./applicants";
 import { fetchApplications } from "./applications";
 import { fetchAssignments } from "./assignments";
@@ -7,6 +16,7 @@ import { fetchContractTemplates } from "./contract_templates";
 import { fetchInstructors } from "./instructors";
 import { fetchPositions } from "./positions";
 import { setGlobals, globalsSelector } from "./globals";
+import { parseURLSearchString } from "../../libs/urlUtils";
 
 /**
  * A helper function to replace all API actions to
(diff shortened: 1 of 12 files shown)


frontend/src/api/actions/init.js
```diff
@@ -32,140 +42,167 @@ function toggleMockApi(enableMockAPI) {
...
  * @export
- * @param {string} stage
- * @param {object} options
+ * @param {string} stage - What stage to start the init procedure at
+ * @param {{ startAfterStage: boolean }} options - if true, start from the stage following the specified stage; if false, start from the specified stage
  * @returns {function} an async function that handles all the API calls.
  */
-export function initFromStage(stage, options) {
+export function initFromStage(stage, options = { startAfterStage: false }) {
+    const startAfterStage = !!options.startAfterStage;
```
CarolineYao, line 77: Is there any occasion that `options.startAfterStage` is not a boolean?

frontend/src/api/actions/init.js
```diff
@@ -32,140 +42,167 @@ function toggleMockApi(enableMockAPI) {
...
- * after the specified `stage`
+ * This function allows you to specify "stage" to start the init procedure at.
+ * It will handle re-fetching any dependent data and dispatching any
+ * required actions depending on the stage specified.
  *
  * @export
- * @param {string} stage
- * @param {object} options
+ * @param {string} stage - What stage to start the init procedure at
+ * @param {{ startAfterStage: boolean }} options - if true, start from the stage following the specified stage; if false, start from the specified stage
```
CarolineYao, line 73: Out of curiosity, why do we need this parameter? Doesn't every `initFromStage` call just start from the specified stage?

frontend/src/api/actions/sessions.js
```diff
@@ -79,23 +80,30 @@ export const deleteSession = payload =>
...
+        }
+        if ((currentActiveSession || { id: null }).id === payload.id) {
             return;
         }
         // If we made it here, the activeSession is changing.
         await dispatch(setActiveSessionAction(payload));
+        // Make sure all tasks we depend on get run
+        if (!skipInit) {
+            await dispatch(
+                initFromStage("setActiveSession", { startAfterStage: true })
```
CarolineYao, line 104: Instead of having `startAfterStage`, maybe we can just call `initFromStage("setActiveSession", { activeSession: payload })`?

frontend/src/api/actions/sessions.js
```diff
@@ -79,23 +80,30 @@ export const deleteSession = payload =>
...
+            await dispatch(setActiveSessionAction(null));
+            return;
+        }
+        if ((currentActiveSession || { id: null }).id === payload.id) {
             return;
         }
         // If we made it here, the activeSession is changing.
         await dispatch(setActiveSessionAction(payload));
+        // Make sure all tasks we depend on get run
+        if (!skipInit) {
```
CarolineYao, line 102: I am not too sure about the `skipInit` here: if the `activeSession` gets changed, all the following stage should be performed right?

frontend/src/api/actions/users.js
```diff
@@ -56,8 +57,13 @@ export const setActiveUserRole = validatedApiDispatcher({
     name: "setActiveUserRole",
     description: "Sets the role of the active user",
     onErrorDispatch: e => deleteError(e.toString()),
-    dispatcher: payload => async dispatch => {
-        dispatch(setActiveUserRoleSuccess(payload));
+    dispatcher: (payload, options = {}) => async dispatch => {
+        await dispatch(setActiveUserRoleSuccess(payload));
+        if (!options.skipInit) {
+            await dispatch(
+                initFromStage("setActiveUserRole", { startAfterStage: true })
```
CarolineYao, line 64: Same thing here, should we just have `initFromStage("setActiveUserRole", { activeUserRole: payload })` ?

frontend/src/api/actions/users.js
```diff
@@ -56,8 +57,13 @@ export const setActiveUserRole = validatedApiDispatcher({
     name: "setActiveUserRole",
     description: "Sets the role of the active user",
     onErrorDispatch: e => deleteError(e.toString()),
-    dispatcher: payload => async dispatch => {
-        dispatch(setActiveUserRoleSuccess(payload));
+    dispatcher: (payload, options = {}) => async dispatch => {
+        await dispatch(setActiveUserRoleSuccess(payload));
+        if (!options.skipInit) {
```
CarolineYao, line 62: Same thing here, if the activeUserRole gets changed, all the following stage should be performed right?

frontend/src/api/actions/init.js
```diff
@@ -32,140 +42,167 @@ function toggleMockApi(enableMockAPI) {
...
-            fetchApplications,
-            fetchAssignments,
-            fetchContractTemplates,
-            fetchInstructors,
-            fetchPositions
-        ];
-
-        fetchActions.map(action => dispatch(action()));
-        await Promise.all(promises);
+        console.log(stage, options);
```
CarolineYao: Is this `console.log` left here intentionally?

CarolineYao (approved):

frontend/src/api/actions/init.js
siefkenj: If someone passes in options but omits `startAfterStage`. If it is undefined, it will cause trouble when it is added.

frontend/src/api/actions/init.js
siefkenj: It says whether you do everything after and including the passed in `stage` or just everything after. It prevents you from needing to know the order of the init process. For example: If you just set the `activeSession`, you don't want the init to set the active session again, but you do want it to do everything afterwards (and if you change the order of stuff, you don't want to have to go back through code elsewhere in the system to change it).

frontend/src/api/actions/sessions.js
siefkenj: I don't see how that would help..?

frontend/src/api/actions/sessions.js
siefkenj: That's right. But there'd be an infinite loop if `activeSession` was set from withing `initFromStage`.

frontend/src/api/actions/init.js
siefkenj: Nope. That was a mistake!

PR #297, opened 2020-02-19
Made `positions` render joined with data for ad and for matching

siefkenj:
The data stored in `position`, `position_data_for_matching` and `position_data_for_ad` should all be joined when rendered. This PR creates a new `PositionService` to help create and get all needed data associated with a position.

**TODO** This PR does not yet handle the `instructor_ids` field correctly.

### DIFF
--- a/backend/app/controllers/api/v1/admin/positions_controller.rb
+++ b/backend/app/controllers/api/v1/admin/positions_controller.rb
@@ -24,21 +24,16 @@ class Api::V1::Admin::PositionsController < ApplicationController
                 service = @position.as_position_service
                 service.update(params: position_params)
             end
-            render_success @position if @position
-            return
-        end
-
         # create a new position if one doesn't currently exist
-        start_transaction_and_rollback_on_exception do
-            service = PositionService.new(params: position_params.except(:id))
-            service.perform
-            @position = service.position
+        else
+            start_transaction_and_rollback_on_exception do
+                service = PositionService.new(params: position_params.except(:id))
+                service.perform
+                @position = service.position
+            end
         end
 
-        # if @position == nil at this point, there was an error which
-        # is being currently rendered. Therefore, we'd better not try to render
-        # a success.
-        render_success @position if @position
+        render_success @position
     end
 
     def find_position
--- a/backend/app/controllers/api/v1/admin/positions_controller.rb
+++ b/backend/app/controllers/api/v1/admin/positions_controller.rb
@@ -10,7 +10,7 @@ class Api::V1::Admin::PositionsController < ApplicationController
 
     # POST /positions/delete
     def delete
-        render_success(object: @position, condition: proc { @position.destroy! })
+        render_on_condition(object: @position, condition: proc { @position.destroy! })
     end
 
     # This method may be manually called from other controllers. Because
--- a/backend/app/controllers/api/v1/admin/positions_controller.rb
+++ b/backend/app/controllers/api/v1/admin/positions_controller.rb
@@ -1,18 +1,39 @@
 # frozen_string_literal: true
 
 class Api::V1::Admin::PositionsController < ApplicationController
-    before_action :find_position
+    before_action :find_position, :store_params
 
     # POST /positions
-    def update
-        render_on_condition(object: @position, condition: proc {
-            @position.update!(position_update_params)
-        })
+    def create
+        render_success upsert(params: @params, position: @position)
     end
 
     # POST /positions/delete
     def delete
         render_success(object: @position, condition: proc { @position.destroy! })
+    end
+
+    # This method may be manually called from other controllers. Because
+    # of that, it doesn't render, instead leaving rendering up to the caller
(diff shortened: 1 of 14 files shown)


siefkenj:
Merging for now so we can continue with test development. Still need a review, though!

backend/app/models/assignment.rb
```diff
@@ -15,6 +15,11 @@ class Assignment < ApplicationRecord
     validates_uniqueness_of :applicant_id, scope: [:position_id]
 
     scope :by_position, ->(position_id) { where(position_id: position_id).order(:id) }
+    scope(:by_session, lambda do |session_id|
```
tsimica, line 18: Keep one style or the other (line 17 vs. 18).

backend/app/controllers/concerns/transaction_handler.rb
```diff
@@ -7,7 +7,15 @@ def start_transaction_and_rollback_on_exception
         ActiveRecord::Base.transaction do
             yield
         rescue StandardError => e
-            render_error(message: e.message)
+            begin
+                render_error(message: e.message)
+            rescue NoMethodError
```
tsimica, line 12: Is this the only error that can happen? I would leave it as rescue from StandardError and let it bubble up.

backend/app/controllers/api/v1/admin/positions_controller.rb
```diff
@@ -1,31 +1,55 @@
...
 
     # POST /positions/delete
     def delete
-        render_success(object: @position, condition: proc { @position.destroy! })
+        render_on_condition(object: @position, condition: proc { @position.destroy! })
     end
 
     private
 
+    # This method may be manually called from other controllers. Because
```
tsimica, line 18: I like the refactoring 😄

backend/app/services/position_service.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def perform
+        @position = Position.new(position_params)
+        @position.save!
+        @position.instructor_ids = instructor_id_params if instructor_id_params
+        @position_data_for_ad = @position.create_position_data_for_ad(position_data_for_ad_params)
+        @position_data_for_matching = @position.create_position_data_for_matching(position_data_for_matching_params)
+        @position_data_for_matching.save!
+        @position_data_for_ad.save!
```
tsimica, line 21: Transaction might be necessary here? I see multiple saves going on.

backend/app/services/position_service.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def update(params:)
+        @params = params
+        @position.update!(position_params)
+        @position.instructor_ids = instructor_ids if instructor_ids
+        @position_data_for_ad.update!(position_data_for_ad_params)
+        @position_data_for_matching.update!(position_data_for_matching_params)
```
tsimica, line 29: Transaction might be necessary here? Same concern as above.

backend/app/controllers/api/v1/admin/session_positions_controller.rb
```diff
@@ -10,31 +10,17 @@ def index
...
-
-    def position_find_or_create_params
-        params.permit(:id, :position_code, :position_title, :hours_per_assignment,
-                      :start_date, :end_date, :duties, :qualifications,
-                      :ad_hours_per_assignment, :ad_num_assignments,
-                      :ad_open_date, :ad_close_date, :desired_num_assignments,
-                      :current_enrollment, :current_waitlisted, :instructor_ids)
+    def find_position
+        find_session
+        @position = @session.positions.find_by_id(params[:id])
```
tsimica, line 20: This should be `find_by(id: params[:id])`

backend/app/models/assignment.rb
siefkenj: Rubocop doesn't like multiline lambdas...

backend/app/services/position_service.rb
siefkenj: The whole action is called from within a transaction when it is used. But...maybe someone will misuse it?

backend/app/controllers/api/v1/admin/session_positions_controller.rb
siefkenj: Why is that?

tsimica: I'm surprised that it works in Rails 6. `find_by_<attribute>` was changed in Rails 5 to `find_by(attribute: x)`.

backend/app/services/position_service.rb
tsimica: Makes sense. No need for it then.

reidka: I was expecting @tsimica to take issue with this class.   I don't understand enough about how ruby or rails works, but I was wondering if Position should have position_data_for_ad and position_data_for_matching as nested attributes or delegates.

I find this approach pretty readable though.

PR #298, opened 2020-02-20
Seeding database

CarolineYao:


frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,246 @@
...
+            return {
+                message: () =>
+                    `expected ${this.utils.printReceived(
+                        received
+                    )} to contain object ${this.utils.printExpected(argument)}`,
+                pass: false
+            };
+        }
+    }
+});
```
siefkenj: This is needed in each of our test files. Let's move it to `utils`

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+    let createdSessionId;
+    let createdPositionId;
+    let createdApplicantId;
+
+    it("create a session", async () => {
+        // get all the sessions so that we can check that our newly inserted session has
+        // a unique id.
+        const { payload: initialSessions } = await apiGET("/admin/sessions");
```
siefkenj: The database is blank when we start. No need to do this.

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,246 @@
...
+        // make sure the data we get back is the same we put in
+        expect(createdSession).toMatchObject(newSessionData);
+        // make sure we have an id
+        expect(createdSession.id).not.toBeNull();
+
+        // fetch all sessions and make sure the data is in there
+        const { payload: withNewSessions } = await apiGET("/admin/sessions");
+        expect(withNewSessions.length).toBe(initialSessions.length + 1);
+        // make sure the id of our new session came back
+        expect(withNewSessions.map(x => x.id)).toContain(createdSession.id);
```
siefkenj: We're just assuming all of this works, so these checks aren't needed.

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,246 @@
...
+        expect(createdAssignment.id).not.toBeNull();
+
+        // fetch the new assignment we just created
+        const newAssignment = await apiGET(
+            `/admin/assignments/${createdAssignment.id}`
+        );
+
+        expect(newAssignment.status).toBe("success");
+        expect(newAssignment.payload).toMatchObject(newAssignmentData);
+    });
```
siefkenj, line 156: Group all of these tests into a `"seed database"` test. Seed everything and assume it works. Then, at the end of the test, retrieve everything and verify that it is our seed data.

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,246 @@
...
+        // make sure the id of our new session came back
+        expect(withNewSessions.map(x => x.id)).toContain(createdSession.id);
+
+        // save the id for later use
+        createdSessionId = createdSession.id;
+    });
+
+    it("create a position", async () => {
+        const { payload: initialPositions } = await apiGET(
+            `/sessions/${createdSessionId}/positions`
```
siefkenj: Routes need to be prefixed with `/admin`. We might consider making a wrapper in `utils` that prefixes routes with a specific user so that it will be easier to test other users in the future.

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,163 @@
...
+ * @param {Function} api.apiPOST A function that when passed a route and data, will return the post response
+ */
+function minimalSetupTest(api = { apiGET, apiPOST }) {
+    const { apiGET, apiPOST } = api;
+    let session = null;
+
+    it("create a session", async () => {
+        await apiPOST("/admin/debug/clear_data");
+
+        // create session
```
siefkenj, line 36: We need comments explaining what's going on. I.e., that we're seeding, assuming everything is going to work. Then we're testing that things got seeded correctly.

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,163 @@
...
+        const newSessionData = {
+            start_date: new Date("2019/09/09").toISOString(),
+            end_date: new Date("2019/12/31").toISOString(),
+            name: "First Session",
+            rate1: 56.54
+        };
```
siefkenj, line 42: These objects should be moved to the top level so that they can be exported.

frontend/src/__tests__/basic.test.js
```diff
@@ -0,0 +1,163 @@
...
+            end_date: new Date("2019/12/31").toISOString(),
+            hours: 90
+        };
+
+        const { payload: createdAssignment } = await apiPOST(
+            `/admin/assignments`,
+            newAssignmentData
+        );
+
+        // fetch all the data and ensure the newly created data comes back
```
siefkenj: Make this comment stand out more.

CarolineYao: Not sure if I understand correctly, but I have use the comment block for this part, and at line 29  since they are comments about the general idea

frontend/src/__tests__/basic.test.js
CarolineYao: I assume that we need the payload of the response since it has the generated id. So I exported an object `seededData` to contain all the session, contract template, etc, and update `seededData` at the end of the test (line 149).

PR #299, opened 2020-02-22
Removed some required fields from the schema

siefkenj:
This PR
  1. Removes some required attributes from the schema
  2. Reogranizes the `tests` directory
  3. Sets up an object to manage the seed data

Point 2: Every js file inside a `__tests__` folder gets run. This folder was renamed to `tests` so that only files ending in `*.test.js` get run.

Point 3: A `databaseSeeder` object has been created which is exported from `setup.js`. This object will contain the IDs and attributes of all the seeded data that exists at the start of every test accessible via `databaseSeeder.seeded`. This will make future tests simpler and more reliable since these "basic" objects will always exist in the database.


frontend/src/tests/setup.js
```diff
@@ -0,0 +1,167 @@
+/**
+ * @jest-environment node
+ */
+import { expect } from "./utils";
+
+class DatabaseSeeder {
```
PeterSenpai, line 6: dumb question: why did you use class? Can't we use pure functions?

siefkenj: We need to store the `seeded` data so that we can export it. It could be a global variable instead...but a class also works. In this case, it's just a way to group related functions.

PeterSenpai: I see, ty.

frontend/src/api/mockAPI/applicants.js
```diff
@@ -33,12 +35,24 @@ export class Applicant extends MockAPIController {
...
+        // The applicants for this session are those who have submitted an
+        // application for this session or those who have an assignment in this
+        // session.
+        const applications = new Application(this.data).findAllBySession(
+            matchingSession
+        );
+        const assignments = new Assignment(this.data).findAllBySession(
+            matchingSession
+        );
+        console.log("the both", applications, assignments);
```
PeterSenpai, line 47: console.log()

frontend/src/tests/setup.js
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Seeds the database by clearing it first, then seeding,
+     * then taking a snapshot.
+     *
+     * @param {*} api
+     * @memberof DatabaseSeeder
+     */
+    async seed(api) {
+        return await seedDatabase(api, this.seededData);
```
PeterSenpai, line 52: Why can't we just put the logic(whatever inside `seedDatabase`) inside `seed` instead of putting extra wrapper?

frontend/src/tests/setup.js
```diff
@@ -0,0 +1,170 @@
...
+    //
+
+    let resp = null;
+
+    // Session
+    resp = await apiPOST("/admin/sessions", seeded.session);
+    // Save the data that was returned to us. What's most important
+    // is the session id. Use Object.assign so that all memory
+    // references to the original seeded.session are kept intact.
+    Object.assign(seeded.session, resp.payload);
```
CarolineYao, line 99: This probably does not happen a lot and sounds like nit picking, but should we change the second argument from `resp.payload` to `{id: respect.payload.id}` just to be extra cautious? With `Object.assign`, the second argument will overwrite the existing field in the first argument. I know that this rarely happens, but for the sake of testing, should we not assume that every field in the response is correct?

PeterSenpai: Personally, I don't feel comfortable to assign certain fields to a data manually since there might be other fields more than just id. Then we have to keep track of all the fields that got created by the backend which makes the tests more complex while checking an incorrect response is relatively easier in my opinion.

siefkenj: I don't think so. We should prefer information from the backend. It might, for instance, reformat our dates in a particular way, and that would be fine.

frontend/src/tests/utils.js
```diff
@@ -2,10 +2,41 @@ import axios from "axios";
...
-const { expect, test } = global;
+const { expect, test, it, describe, beforeAll, afterAll } = global;
 
-// Jest demands there be a test in every file, so we put a no-op test here.
-test("no-op", () => {});
+// add a custom `.toContainObject` method to `expect()` to see if an array contains
+// an object with matching props. Taken from
+// https://medium.com/@andrei.pfeiffer/jest-matching-objects-in-array-50fe2f4d6b98
+expect.extend({
+    toContainObject(received, argument) {
```
colezemel, line 11: I think this is unnecessary because the `toContainEqual()` method accomplishes the same thing:

[https://jestjs.io/docs/en/expect#tocontainequalitem](url)

PeterSenpai: Good point but this requires a package called `jest-extended` which I am not sure if we have it or not.

colezemel: No I don't think you do. I've been playing around with it and I was able to use it without that package

PeterSenpai: Then, in this case, we have the package somewhere (probably with the latest create react app or jest). But according to the link you attached and the GitHub page it leads to, it's a separate package of jest

colezemel: I wasn't using it in our app, just doing some simple jest testing separately. Are you talking about where it says _"For additional Jest matchers maintained by the Jest Community check out jest-extended"_ ? If you follow that link it looks like there are more matchers in addition to the ones included in the link I provided that aren't maintained by jest but by the jest community. The link I provided is the official jest documentation so I believe `toContainEqual()` and all the other ones there are provided without having to install any other packages.

colezemel: [https://jestjs.io/docs/en/expect#tocontainequalitem](url) Try copying and pasting this link into your browser. When I click it directly it leads to a 404 page not found on Github. But if you copy and paste it leads you to the page I was talking about (official jest docs)

PeterSenpai: Yeah, I opened the correct link. It doesn't matter since it works. (i was talking about the README on this page [jest-exended](https://github.com/jest-community/jest-extended) )

PeterSenpai: since it works, the next question would be: are they actually do the same thing.(I haven't read it carefully, will do that later)

PeterSenpai: they are slightly different tho:
```javascript
const myBeverage = { delicious: true };
const l = [{ delicious: true, sour: false }];

expect(l).toContainEqual(myBeverage); // fail
expect(l).toContainObject(myBeverage); // pass
```
but not sure if they serve the same thing in our cases

PR #304, opened 2020-02-25
fix the cancel buttons

### LINKED ISSUE #301: Pressing "cancel" on "Add Assignment" dialog doesn't close it.

PeterSenpai:


CarolineYao:
Here is my theory about why close button worked but cancel button didn't.

When close button is clicked, no argument is passed in to `onHide` while cancel button is clicked, the click event (an object) is passed in to `onHide`. Therefore, the attribute `..dialogVisible` is set to `undefined` (which has the same effect as `false`) after close button is clicked and attribute is set to an object (which has the same effect as `true`) after cancel button is clicked. 

I setup a temp function to be passed in as `onHide` so we can see it more clearly:
Screen Shot 2020-02-27 at 9 03 28 PM

As for exactly why 2 different values are passed in, referring to react-bootstrap source code: 
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/SafeAnchor.js line 55: 
```js
if (onClick) {
    onClick(event);
}
```
we can see that the event is usually passed in to `onClick` for buttons and that is why the cancel button has the click event object (for cancel button, `<Button onClick={onHide}>`); however, as for cancel button, `onHide` is explicitly called without any argument referring to the source code:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ModalHeader.js line 57 
```js
const handleClick = useEventCallback(() => {
      if (context) context.onHide();
      if (onHide) onHide();
});
```

PeterSenpai:
So we have `addDialogVisible` and `setAddDialogVisible` passed in Modal as `show` and `onHide`. 
First one is a `bool` value and its value is controlled by `setAddDialogVisible`.
  
  
 this is the close button of the modal:

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ModalHeader.js

```jsx
 return (
      <div ref={ref} {...props} className={classNames(className, bsPrefix)}>
        {children}

        {closeButton && (
          <CloseButton label={closeLabel} onClick={handleClick} />
        )}
      </div>
    );
```
and the `onClick` is define in the same file:

```javascript
const handleClick = useEventCallback(() => {
    if (context) context.onHide();
    if (onHide) onHide();
});
```
It calls the `context.onHide()` first and then calls our `onHide()`

`context.onHide()` is declared in this file:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ModalContext.js

Literally 3 lines:
`ModalContext.js`:

```javascript
import React from 'react';

const ModalContext = React.createContext({ onHide() {} });

export default ModalContext;
```

But it is modified when  `Modal` Component got created.
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js

`Modal.js`:
```javascript
modalContext = {
    onHide: () => this.props.onHide(),
  };
```
So now `Modal.props.onHide` now is hooked up with `ModalHeader`'s `closeButton`'s `onClick` witch changes `Modal.props.show`


And `Modal.props.show` also got passed into a component called `BaseModal` import from `react-overlays`

`Modal.js`:
```jsx
render() {
    const {
      
      /* BaseModal props */
      show,
      // some code
      ...props
    } = this.props;

```

```jsx
<BaseModal
    {...{
            show,
            // some code
            ...
          }}
        >
          <Dialog
            {...props}
            onMouseDown={this.handleDialogMouseDown}
            className={dialogClassName}
          >
            {children}
          </Dialog>
        </BaseModal>
```



Below is `react-overlays`'s `Modal`, the state `exited` is controlled by the `show` property we passed in. If `show === undefined/false/null` the Modal would exit.

https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.js

```jsx
  static defaultProps = {
    show: false,
    ...
    onHide: () => {},
    renderBackdrop: props => <div {...props} />,
  };

  state = { exited: !this.props.show };
```

And this our code of creating `Modal`:

in `index.js`:

```jsx
<ConnectedAddAssignmentDialog
    show={addDialogVisible}
    onHide={setAddDialogVisible}
/>
```
In `add-assignment-dialog.js`:
```jsx
<Modal show={show} onHide={onHide}>
```

So when the `closeButton` got clicked `setAddDialogVisible` is triggered without given arguement. It set `addDialogVisible` to `undefined` and pass all the way down to `react-ovelays/Modal.js` and make the modal exited.

But when the cancel button is clicked:
```jsx
<Button onClick={onHide} variant="light">
    Cancel
</Button>
```
We passed the `onHide` derectly without using a arrow function wrapper or `bind` in the parent component. It actually lost the context so it won't work.

If we do:
```jsx
<Button onClick={() => {onHide}} variant="light">
    Cancel
</Button>
```
The cancel button works but `show` is still `undefined`.

So I propose the fix should be the one on my PR

yarn.lock
```diff
@@ -9,6 +9,46 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
+"@babel/code-frame@^7.8.3":
```
siefkenj: Why are all these changes here? They are unrelated to your fix...

PeterSenpai: The formatting tests were failing due to the outdated yarn.lock file. So I updated it.
Do you think this should be a separate PR?

siefkenj: Yep. PRs should be focused on one thing.

PeterSenpai: Fixed!

yarn.lock
```diff
@@ -628,7 +628,7 @@ glob@^7.1.3:
     once "^1.3.0"
     path-is-absolute "^1.0.0"
 
-globals@^11.1.0, globals@^11.7.0:
+globals@^11.7.0:
```
siefkenj, line 532: No changes to `yarn.lock` in this PR!

PeterSenpai: yeah sorry, I had some problem with my rebase steps. I am learning + resolving it now.

siefkenj:
This has gotten out of hand. I am closing this PR. Make a fresh branch and make the changes again.

PR #315, opened 2020-03-06
Reorganize Tests

CarolineYao:


frontend/src/tests/wage-chunk-tests.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable */
```
siefkenj, line 1: Why is eslint disabled?

CarolineYao: it's complaining about the 
```js
export function wageChunksTests({ apiGET, apiPOST })
```
we need to either use eslint disabled, or not including the passed in arguments for all the tests only have `todo`. For example:
```js
export function wageChunksTests() {
    it.todo("get wage_chunks for position");
}
```

PR #321, opened 2020-03-19
Refactor Session tests

PeterSenpai:


frontend/src/tests/session-tests.js
```diff
@@ -27,6 +27,7 @@ export function sessionsTests(api) {
 
     beforeAll(async () => {
         await apiPOST("/debug/restore_snapshot");
+        session = databaseSeeder.seededData.session;
```
CarolineYao, line 30: Out of curiosity, did it still not work when you assign it before the `beforeAll` and used it inside of each test?

frontend/src/tests/session-tests.js
```diff
@@ -76,12 +78,12 @@ export function sessionsTests(api) {
         expect(withNewSessions.map(x => x.id)).toContain(createdSession.id);
 
         // save this session for use in later tests
```
CarolineYao: Probably don't need this command anymore?

frontend/src/tests/session-tests.js
```diff
@@ -76,12 +78,12 @@ export function sessionsTests(api) {
         expect(withNewSessions.map(x => x.id)).toContain(createdSession.id);
```
CarolineYao: Should also check that `expect(withNewSessions).toContainObject(createdSession)`? Like what we did in `verifySeededDatabase`

frontend/src/tests/session-tests.js
```diff
@@ -92,7 +94,7 @@ export function sessionsTests(api) {
         expect(updatedSession).toContainObject(newData);
```
CarolineYao, line 93: Don't know for sure, but should we call `apiGET` again and see if it match `updatedSession`? Maybe redundant, but just to be extra safe?

frontend/src/tests/session-tests.js
```diff
@@ -92,7 +94,7 @@ export function sessionsTests(api) {
...
     });
```
CarolineYao, line 95: General question for error checking, should we also check the response code?

CarolineYao (approved):

frontend/src/tests/session-tests.js
PeterSenpai: yeah it fails some of the tests

frontend/src/tests/session-tests.js
PeterSenpai: since we only have 200 or 500. I think it's enough to just check status.

frontend/src/tests/session-tests.js
PeterSenpai: sure

frontend/src/tests/session-tests.js
PeterSenpai: sure

frontend/src/tests/session-tests.js
PeterSenpai: I think I already did here, or am I wrong?

CarolineYao: Oops, did not see the folded part

PR #323, opened 2020-03-20
Refactor Application tests

colezemel:
First basic tests for application. Works for mock API, but adding a new application to the real API gives an error


PeterSenpai:
Hmm, why is the name of PR `seeding database` lol?

Issue #317, opened 2020-03-07
[MockAPI/Backend] session gets updated with an invalid value of `name`


PeterSenpai:
Come across this issue when refactoring session test ``throw error when ` name` is empty``

__The correct__ behavior:
- Throw an error on creating a new session with invalid `name` (`""` or `undefined`)
- Throw an error on updating a new session with invalid `name` (`""` or `undefined`)

__Current__ behavior:
- Throw an error on creating a new session with invalid `name` (`""` or `undefined`)
-  Session gets updated with an invalid value of `name`

PR #325, opened 2020-03-26
[MockAPI] fix the issue when session gets updated with an invalid value of `name` 

PeterSenpai:
[issue#317](https://github.com/uoft-tapp/tapp/issues/317)

### DIFF
--- a/frontend/src/api/mockAPI/sessions.js
+++ b/frontend/src/api/mockAPI/sessions.js
@@ -27,19 +27,24 @@ export class Session extends MockAPIController {
         }
     }
 
-    validateProp(obj, prop) {
+    validateProp(prop, value, id) {
         if (prop === "name") {
-            if (obj[prop] === undefined || obj[prop].length === 0) {
+            if (value === undefined || value.length === 0) {
                 throw new Error(
                     `Property ${prop} cannot be empty or undefined.`
                 );
             } else {
                 const filteredData = this.findAll().filter(
-                    item => item.id !== obj.id
+                    item => item.id !== id
                 );
-                const message = getAttributesCheckMessage(obj, filteredData, {
-                    name: { unique: true }
-                });
+
+                const message = getAttributesCheckMessage(
+                    { name: value },
+                    filteredData,
+                    {
+                        name: { unique: true }
+                    }
+                );
                 if (message) {
                     throw new Error(message);
                 }
--- a/frontend/src/api/mockAPI/sessions.js
+++ b/frontend/src/api/mockAPI/sessions.js
@@ -27,13 +27,28 @@ export class Session extends MockAPIController {
         }
     }
     validateUpdate(session) {
-        const message = getAttributesCheckMessage(session, this.ownData, {
-            name: { required: true, unique: true },
-        });
-        if (message) {
-            throw new Error(message);
+        if ("name" in session) {
+            if (session.name === undefined || session.name.length === 0) {
+                // console.log("Invalid!!!!!!!!!!!!!!!!!!");
+                throw new Error("Invalid property: name.");
+            } else {
+                const filteredData = this.findAll().filter(
+                    (item) => item.id !== session.id
+                );
+                const message = getAttributesCheckMessage(
+                    session,
+                    filteredData,
+                    {
+                        name: { unique: true },
+                    }
+                );
+                if (message) {
+                    throw new Error(message);
+                }
+            }
         }
     }
+
     upsert(obj) {
         if (this.rawFind(obj)) {
             this.validateUpdate(obj);
--- a/frontend/src/api/mockAPI/sessions.js
+++ b/frontend/src/api/mockAPI/sessions.js
@@ -2,7 +2,7 @@ import { getAttributesCheckMessage, MockAPIController } from "./utils";
 import {
     documentCallback,
     wrappedPropTypes,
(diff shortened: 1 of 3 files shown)


colezemel (approved):
Looks good!

frontend/src/api/mockAPI/sessions.js
```diff
@@ -26,6 +26,36 @@ export class Session extends MockAPIController {
             throw new Error(message);
         }
     }
+    validateUpdate(session) {
+        if ("name" in session) {
+            if (session.name === undefined || session.name.length === 0) {
```
CarolineYao, line 31: If `session.name` is `undefined`, doesn't `session.name.length` give a TypeError?

CarolineYao (approved):

frontend/src/api/mockAPI/sessions.js
```diff
@@ -26,6 +26,36 @@ export class Session extends MockAPIController {
             throw new Error(message);
         }
     }
+    validateUpdate(session) {
```
PeterSenpai: don't create a new method here

frontend/src/api/mockAPI/utils.js
```diff
@@ -134,7 +134,10 @@ export function getAttributesCheckMessage(
 ) {
     for (const [prop, requirements] of Object.entries(props)) {
         // Required attributes cannot be null or the empty string
-        if (requirements.required && (obj[prop] == null || obj[prop] === "")) {
+        if (
```
PeterSenpai: undo the change and make a comment

frontend/src/api/mockAPI/utils.js
```diff
@@ -245,6 +246,18 @@ export class MockAPIController {
...
+    /**
+     * Validates a property of an item instance. Throws an error
+     * if this property are invalid/incompatable.
+     *
+     * @param {object} obj
+     * @param {string} prop
+     * @memberof MockAPIController
+     */
+    validateProp(obj, prop) {
+        if (obj[prop]) return true;
```
siefkenj: Never, ever, ever use an if statement without `{...}` around it's following block. Also, this code is wrong. You probably meant `prop in obj` or `obj[prop] != null`.

Also, why the if statement? I think the if statement should not be there.

PeterSenpai: I see, if I don't use obj or prop anywhere inside this function, prettier would complain(fail the formatting tests). but both obj and prop are checked before even call this function. what do you think I should do here

siefkenj: You can remove `obj` and `prop` from the function signature. Or you can add a `// eslint-ignore-next-line` (or whatever it is..) comment.

frontend/src/api/mockAPI/utils.js
```diff
@@ -245,6 +246,19 @@ export class MockAPIController {
...
+    /**
+     * Validates a property of an item instance. Throws an error
+     * if this property are invalid/incompatable.
+     *
+     * @param {object} obj
+     * @param {string} prop
+     * @memberof MockAPIController
+     */
+    // eslint-disable-next-line no-unused-vars
+    validateProp(obj, prop) {
```
siefkenj: Why are you passing in an object? I don't get it...

frontend/src/api/mockAPI/sessions.js
```diff
@@ -26,6 +26,28 @@ export class Session extends MockAPIController {
...
         }
     }
+
+    validateProp(obj, prop) {
+        if (prop === "name") {
+            if (obj[prop] === undefined || obj[prop].length === 0) {
+                throw new Error(
+                    `Property ${prop} cannot be empty or undefined.`
+                );
+            } else {
```
siefkenj: I don't understand what you're doing. I suspect it's wrong (i.e., it doesn't belong in this function). You need comments to explain.

PeterSenpai: say we want to update a session. `obj` here is the request body `{ id: 1, name: xxx}`. `prop` is the key/field. if the field is `name` , we need to make sure `{ ...,name: ""}` and `{..., name: undefined}` throws an error. And if `name` is not empty, we still need to check if it's updatable, ie, if it is unique to the rest of the sessions (all the session that id is not 1)

frontend/src/api/mockAPI/utils.js
PeterSenpai: say we want to update a session. `obj` here is the request body `{ id: xxx, name: xxx}`. `prop` is the key/field. I need to get the value of the field.

frontend/src/api/mockAPI/sessions.js
siefkenj: This function should only check the `prop` that is passed in and nothing else.

frontend/src/api/mockAPI/utils.js
siefkenj: You should be passing in the prop and the value, not the prop and the object.

PeterSenpai: right... That sounds way better

frontend/src/api/mockAPI/sessions.js
PeterSenpai: Hmm... You mean it shouldn't throw errors but return false or a message?

PR #329, opened 2020-03-26
adding tests for assignment section

CarolineYao:


frontend/src/tests/assignment-tests.js
```diff
@@ -1,10 +1,125 @@
...
+    const session = databaseSeeder.seededData.session;
+    const position = databaseSeeder.seededData.position;
+    const assignment = databaseSeeder.seededData.assignment;
+
+    beforeAll(async () => {
+        await apiPOST("/debug/restore_snapshot");
+    }, 30000);
+
+    it("get assignments for session", async () => {
+        const resp = await apiGET(`/admin/sessions/${session.id}/assignments`);
```
PeterSenpai, line 22: maybe we can compare the response with the data in `databaseSeeder.seededData`

frontend/src/tests/assignment-tests.js
```diff
@@ -1,10 +1,125 @@
...
+import { databaseSeeder } from "./setup";
 
-// TODO: Remove eslint disable
-// eslint-disable-next-line
 export function assignmentsTests({ apiGET, apiPOST }) {
-    it.todo("get assignments for session");
-    it.todo("get assignments for position");
-    it.todo("create and delete assignment for position");
-    it.todo("update assignment dates/note/offer_override_pdf");
+    const session = databaseSeeder.seededData.session;
```
PeterSenpai, line 12: for some reason, if I assigned the initial data here, I got the incorrect data ( without ids) not sure if that happened to you or not

frontend/src/tests/assignment-tests.js
```diff
@@ -1,10 +1,125 @@
...
+        expect(withUpdatedAssignment.map(x => x.id)).toContain(
+            updatedAssignment.id
+        );
+        expect(
+            withUpdatedAssignment.filter(s => s.id === updatedAssignment.id)
+        ).toContainObject(updatedAssignment);
+
+        expect(withUpdatedAssignment.length).toEqual(1);
+    });
+
```
PeterSenpai: maybe we can have some error checking later when the backend is fully functional?

frontend/src/tests/assignment-tests.js
colezemel: What do you mean by "assigned the initial data"?

frontend/src/tests/assignment-tests.js
colezemel: I believe that is what she's doing. In like 25 she compares the response payload to seededData.assignment.

Or are you suggesting we also compare the response to the seededData response and not the explicit value of ``status: success``?

colezemel (approved):

frontend/src/tests/assignment-tests.js
CarolineYao: What is happening here after the get response is: 

1. check the response status and make sure the api call succeed. (line 23)
2. compare the payload with seeded data (line 25)
3. check if payload matches the proptype (line 27)

frontend/src/tests/assignment-tests.js
CarolineYao: Maybe we can do it in another PR? I was just finishing all the test cases that was originally here. Needs to investigate for error checking.

frontend/src/tests/assignment-tests.js
CarolineYao: If you look at line 23, it is using `session.id` and the test is passing. (try doing `console.log` inside of a test instead of outside of the test)

frontend/src/tests/assignment-tests.js
PeterSenpai: oops, I was blind, my bad

frontend/src/tests/assignment-tests.js
```diff
@@ -1,10 +1,125 @@
...
+    it("create assignment for position", async () => {
+        // create a new applicant to assign
+        const newApplicantData = {
```
CarolineYao: Create a new position instead

PR #332, opened 2020-04-01
(README) update 'api' to 'backend'

colezemel:
Change the ``mkdir`` command to use backend rather than api


frontend/src/tests/api.test.js
```diff
@@ -76,4 +76,7 @@ describe("Mock API tests", () => {
     describe.skip("`/instructors` tests", () => {
         instructorsTests(mockAPI);
     });
+    describe("`/applications` tests", () => {
+        applicationsTests(mockAPI);
```
PeterSenpai, line 80: the frontend files shouldn't be changed, I suspect that you created this branch out of your applicant tests branch instead of master, so it has the previous commit.

PR #333, opened 2020-04-02
Refactor Application tests

colezemel:
Adding basic tests for application.

Tests pass for Mock API, but not for Real API (I have added an issue for this)

### DIFF
--- a/frontend/src/tests/application-tests.js
+++ b/frontend/src/tests/application-tests.js
@@ -22,53 +22,41 @@ import { databaseSeeder } from "./setup";
 export function applicationsTests(api) {
     const { apiGET, apiPOST } = api;
     let session = null,
-        applicant = null;
-
-    const newApplicationData = {
-        comments: "",
-        program: "Phd",
-        department: "Computer Science",
-        previous_uoft_ta_experience: "Last year I TAed a bunch",
-        yip: 2,
-        annotation: "",
-        position_preferences: [
-            {
-                preference_level: 2,
-                position_id: 10
-            },
-            {
-                preference_level: 3,
-                position_id: 15
-            }
-        ]
-    };
+        applicant = null,
+        application = null;
 
     // set up a session to be available before tests run
     beforeAll(async () => {
         await apiPOST("/debug/restore_snapshot");
         session = databaseSeeder.seededData.session;
         applicant = databaseSeeder.seededData.applicant;
-        newApplicationData.session_id = session.id;
-        newApplicationData.applicant_id = applicant.id;
+        application = databaseSeeder.seededData.application;
+        application = {
+            ...application,
+            session_id: session.id,
+            applicant_id: applicant.id
+        };
     }, 30000);
 
     it("fetch applications", async () => {
         // grab the applications of the new session
-        const resp1 = await apiGET(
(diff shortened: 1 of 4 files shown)


frontend/src/tests/application-tests.js
```diff
@@ -1,3 +1,74 @@
...
+        comments: "",
+        program: "Phd",
+        department: "Computer Science",
+        previous_uoft_ta_experience: "Last year I TAed a bunch",
+        yip: 2,
+        annotation: "",
+        position_preferences: [
+            {
+                preference_level: 2,
+                position_id: 10
```
PeterSenpai: is it better if we can use the real position data for this?

frontend/src/tests/application-tests.js
```diff
@@ -1,3 +1,74 @@
...
+    it("fetch applications", async () => {
+        // grab the applications of the new session
+        const resp1 = await apiGET(
+            `/admin/sessions/${session.id}/applications`
+        );
+
+        expect(resp1).toMatchObject({ status: "success" });
+
+        // check the type of payload
+        checkPropTypes(PropTypes.arrayOf(applicationPropTypes), resp1.payload);
```
PeterSenpai: can we check the `resp1` with the seeded data?

frontend/src/tests/application-tests.js
```diff
@@ -1,3 +1,74 @@
...
+            }
+        ]
+    };
+
+    // set up a session to be available before tests run
+    beforeAll(async () => {
+        await apiPOST("/debug/restore_snapshot");
+        session = databaseSeeder.seededData.session;
+        applicant = databaseSeeder.seededData.applicant;
+        newApplicationData.session_id = session.id;
```
CarolineYao: We don't have a style guide for this, but we have been using 
```
newApplicationData = {
    ...newApplicationData,
    session_id: session.id,
    applicant_id: applicant.id
}
```
Personally, I prefer doing it this way since `session_id` and `applicant_id` fields do not exist in `newApplicationData`. But it is just a suggestion, maybe we should come up with a consistent way of doing this.

frontend/src/tests/application-tests.js
```diff
@@ -1,3 +1,74 @@
...
+        // check the type of payload
+        checkPropTypes(PropTypes.arrayOf(applicationPropTypes), resp1.payload);
+    });
+
+    it("add an application to a session", async () => {
+        const resp2 = await apiPOST(`/admin/applications`, newApplicationData);
+
+        expect(resp2).toMatchObject({ status: "success" });
+        checkPropTypes(applicationPropTypes, resp2.payload);
+        expect(resp2.payload).toMatchObject(newApplicationData);
```
CarolineYao: We should call `apiGET` again and make sure that newly added data is in the array as well

CarolineYao (approved):

frontend/src/tests/application-tests.js
siefkenj: We absolutely should not be hardcoding this...you have access to the seeded data through databaseSeeder.

siefkenj:
This PR is stale and we're not uploading applications ATM. We'll revisit this when we're ready to do applications.

PR #338, opened 2020-04-20
Add `active_user` to the backend

siefkenj:
This PR adds debug routes to get/set users and the active user. When we enable user-based permissions, we will need debug routes to fake "user credentials", etc.


backend/app/controllers/api/v1/admin/users_controller.rb
```diff
@@ -28,7 +26,7 @@ def create
     private
 
     def user_params
-        params.permit(:id, :utorid, :email, roles: [])
+        params.permit(:id, :utorid, roles: [])
```
colezemel, line 29: Are we not storing email in user anymore? Why?

siefkenj: Good question! We have the email of TAs through their applicant info, we have the email of instructors in the instructor table. I don't think it makes sense to duplicate it. The only one left out is `admin`s, but they don't get stuff emailed to them.

Issue #339, opened 2020-04-21
Nicer debug checks


siefkenj:
Right now, we run `expect(resp).toMatchObject({ status: "success" })` to check if an API call was successful. However, when it is not successful, we get an unhelpful message saying that the object doesn't match a `success` object. Then we have to manually add a print statement to see the contents of `resp` to start debugging.

There should be a standard function that is called to check if a call was successful that prints the contents of `resp` when there is an error.

PR #340, opened 2020-04-21
[Issue #339] Nicer debug checks

colezemel:
Response is logged to console because error message isn't working properly.

**Unit test fails on purpose:**
Temporarily tests with status = failed for debugging (to see the functionality)

Is this the kind of function you were looking for?

### DIFF
--- a/frontend/src/tests/instructor-tests.js
+++ b/frontend/src/tests/instructor-tests.js
@@ -10,7 +10,7 @@ import {
     it,
     beforeAll,
     afterAll,
-    toMatchSuccessDebug
+    checkResponseSuccess
 } from "./utils";
 
 export function instructorsTests({ apiGET, apiPOST }) {
@@ -39,12 +39,12 @@ export function instructorsTests({ apiGET, apiPOST }) {
 
         // create a new instructor
         const resp1 = await apiPOST(`/instructors`, newInstructorData);
-        toMatchSuccessDebug(resp1);
+        checkResponseSuccess(resp1);
         expect(resp1.payload).toMatchObject(newInstructorData);
 
         // make sure instructor is on instructor list
         const resp2 = await apiGET(`/instructors`);
-        toMatchSuccessDebug(resp2);
+        checkResponseSuccess(resp2);
         expect(resp2.payload).toContainObject(newInstructorData);
 
         // set instructor to used by later test
@@ -53,7 +53,7 @@ export function instructorsTests({ apiGET, apiPOST }) {
 
     it("get instructors", async () => {
         const resp = await apiGET("/instructors");
-        toMatchSuccessDebug(resp);
+        checkResponseSuccess(resp);
         checkPropTypes(PropTypes.arrayOf(instructorPropTypes), resp.payload);
     });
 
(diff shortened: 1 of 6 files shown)


frontend/src/tests/utils.js
```diff
@@ -36,6 +36,25 @@ expect.extend({
...
+ * @param {object} resp The recieved array
+ * @param {object} obj  The json object to compare resp to
+ * @returns
+ */
+export function toMatchSuccessDebug(resp) {
+    try {
+        expect(resp).toMatchObject({ status: "failed" });
+    } catch (err) {
+        // err.message = "resp: \n", resp`;
+        console.log("resp: \n", resp);
```
siefkenj: You shouldn't use console.log. Look at how the proptypes check was done.

colezemel: Changed to console.warn

frontend/src/tests/utils.js
```diff
@@ -36,6 +36,23 @@ expect.extend({
...
+ * @export
+ * @param {object} resp The recieved array
+ * @returns
+ */
+export function toMatchSuccessDebug(resp) {
+    try {
+        expect(resp).toMatchObject({ status: "success" });
+    } catch (err) {
+        console.warn("resp: \n", resp);
+        throw err;
```
siefkenj, line 52: This is still not right. The error message needs to be descriptive, and you shouldn't use console.warn.

frontend/src/tests/instructor-tests.js
```diff
@@ -9,7 +9,8 @@ import {
     expect,
     it,
     beforeAll,
-    afterAll
+    afterAll,
```
siefkenj, line 12: Where is this used? It probably shouldn't be...

frontend/src/tests/utils.js
```diff
@@ -36,6 +36,23 @@ expect.extend({
...
 
+/**
+ * Customize the `.toMatchObject` method to print the contents of the recieved
+ * object if the response call fails
+ *
+ * @export
+ * @param {object} resp The recieved array
+ * @returns
+ */
+export function toMatchSuccessDebug(resp) {
```
siefkenj: This isn't a good name. Come up with a better one :-)

frontend/src/tests/instructor-tests.js
colezemel: I didn't change this

frontend/src/tests/utils.js
colezemel: Ok think I finally got it. I was missing `JSON.stringify`

frontend/src/tests/instructor-tests.js
siefkenj: Please look into it and file an issue/PR if needed.

frontend/src/tests/utils.js
siefkenj: This is improved! But I think you can do better. Imagine you, in January, trying to work on your first bug. What error message would you hope to see? Write that message.

Also, the description of this function is not good. In particular, it doesn't explain why anyone should use this function.

CarolineYao (approved):

siefkenj:
I merged the other refactoring, so there are a few conflicts now.

siefkenj:
I've addressed part of this in #349

colezemel:
#349 addressed this. I created #354 for the `afterAll` issue in Instructor

colezemel:
#349 addressed this. I created #354 for the afterAll issue in Instructor

PR #349, opened 2020-04-26
Removed foreign key requirements & other fixes

siefkenj:
Reporting tags previously required both a wage chunk and a position. However, reporting tags can belong to just one of those. The requirement they belong to both is removed.

Tests have been refactored to reseed the database every time instead of restoring data. This is a workaround because `restore_snapshot` behaves in unexpected ways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Where
 **[change region]** is a name of a logical region of code in the project, i.e.
 
 -   docker
--   api
+-   backend
 -   db
 -   frontend
 -   documentation
@@ -83,7 +83,7 @@ This patch introduces a new front end view for adding positions to the system.
 ```
 
 ```
-fix(api): Fix api endpoint for adding positions
+fix(api): Fix backend endpoint for adding positions
 
 This patch fixes a bug in the API whereby attempts to hit the endpoint meant for
 adding a new position would always fail.
@@ -174,7 +174,7 @@ is served on a different port. To prevent confusion if you try to access TAPP th
 port, set up a static asset to be served by rails.
 
 ```
-mkdir api/public && echo "Please go to localhost:8000 for frontend" > api/public/index.html
+mkdir backend/public && echo "Please go to localhost:8000 for frontend" > backend/public/index.html
 ```
 
 Finally, we can build the docker images and migrate the database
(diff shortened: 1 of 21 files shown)


frontend/src/tests/utils.js
```diff
@@ -33,6 +33,32 @@ expect.extend({
...
+        }
+        if (received.status === argument) {
+            return {
+                pass: true,
+                message: () =>
+                    `expected API response to not have ${this.utils.printExpected(
+                        {
+                            status: argument
+                        }
+                    )} but received ${this.utils.printReceived(received)}`
```
colezemel, line 51: This looks good. Much better way to do it. But I'm not understanding why we'd want this message

colezemel: Never mind, this message appears when we run `toHaveStatus` with `expect.not`. I don't think we'd ever want to run `expect(...).not.toHaveStatus(...)` but it is useful for our other custom matcher. Should we keep it in for consistency or remove it since it's a little confusing in this case (maybe just for me)?

README.md
```diff
@@ -213,9 +213,38 @@ To view the STDOUT from a docker container of a running server, you can invoke
...
+which will display the output to the command line (it won't actually insert into a database.)
+
+##### Running tests from a ram-drive
+
+Since the test suite clears the database several times, it can be slow! To speed things
+up by a factor of ~10x, you can run the database from a ram drive instead. In linux, you can
+set up your ram drive with
+
+```
+cd <tapp home director>
```
colezemel, line 233: directory*

README.md
```diff
@@ -213,9 +213,38 @@ To view the STDOUT from a docker container of a running server, you can invoke
...
+mkdir /dev/shm/pg_data
```
colezemel, line 234: I get an error when running this.
`mkdir: /dev/shm: No such file or directory`

frontend/src/tests/api.test.js
```diff
@@ -26,13 +26,14 @@ describe("API tests", () => {
         await databaseSeeder.verifySeed(api);
     }, 30000);
 
-    describe.skip("`/admin/sessions` tests", () => {
+    describe("`/admin/sessions` tests", () => {
         sessionsTests(api);
-    });
+    }, 30000);
```
colezemel, line 31: Why do we need this here?

frontend/src/tests/utils.js
siefkenj: The documentation for `jest` says to include both cases. It confused me at first as well...

README.md
siefkenj: Good catch! Can you fix this?

README.md
siefkenj: Are you on linux? Or are you on mac? Whatever the case, figure out how to get a ram drive set up and put instructions that work into the readme.

frontend/src/tests/api.test.js
siefkenj: We probably don't need it if the `beforeAll` action has a long timeout.

PR #355, opened 2020-04-29
Refactor Instructor Tests

colezemel:


frontend/src/tests/instructor-tests.js
```diff
@@ -18,67 +18,83 @@ import { databaseSeeder } from "./setup";
...
+        const seededInstructor = databaseSeeder.seededData.instructor;
+        expect(responsedInstructor).toEqual(seededInstructor);
     });
 
-    it("get instructors", async () => {
-        const resp = await apiGET("/instructors");
-        expect(resp).toMatchObject({ status: "success" });
-        checkPropTypes(PropTypes.arrayOf(instructorPropTypes), resp.payload);
-    });
+    // it("create instructor", async () => {
```
CarolineYao: Out of curiosity, why are these test commented out? If the purpose is to not run this test, we should probably go for `it.skip`

frontend/src/tests/setup.js
```diff
@@ -42,6 +42,12 @@ class DatabaseSeeder {
             active_user: {
                 utorid: "smithh",
                 roles: ["admin", "instructor", "ta"]
+            },
```
CarolineYao: Correct me if I am wrong, but I was under the impression that these seeded data are the minimal data that we need for all the test. For example, because sessions and positions are commonly used in the the whole app, if we don't have these seeded data, we will have to create the data repeatedly and that is why we have them inside setup. But instructor is not a commonly used data for it to be inside seeded data. We should do it the old approach: test apiPOST and then apiGET.

PR #362, opened 2020-05-03
Finish position test

CarolineYao:


frontend/src/tests/position-tests.js
```diff
@@ -102,15 +102,33 @@ export function positionsTests(api = { apiGET, apiPOST }) {
...
 
         const resp2 = await apiGET(`/admin/sessions/${session.id}/positions`);
         expect(resp2.payload).not.toContainObject(newPosition);
     });
 
-    it.todo("error when updating a position to have an empty position_code");
+    it("error when updating a position to have an empty position_code", async () => {
+        const updatedPositionData = {
+            ...newPosition,
+            position_code: null
```
siefkenj: This should be `""`, not `null`. In theory, null attributes are ignored...

frontend/src/tests/position-tests.js
```diff
@@ -102,15 +102,33 @@ export function positionsTests(api = { apiGET, apiPOST }) {
...
+
+        const resp = await apiPOST(`/admin/positions`, updatedPositionData);
+        expect(resp).toHaveStatus("error");
+        checkPropTypes(errorPropTypes, resp);
+    });
+
+    it("error when creating a position for a session with an invalid id", async () => {
+        const updatedPositionData = {
+            ...newPosition,
+            id: 500
```
siefkenj: This may or may not be invalid. Grab all positions first and pick an ID that isn't in the list.

PR #365, opened 2020-05-06
downgrade swagger-ui-react version.

PeterSenpai:
API docs is currently broken because of the swagger-ui-react library.
Therefore downgrade to ver.3.24.3. After upgrading, it works correctly.

Screenshot of the error on ver.3.25.x
image


siefkenj:
You need to update the description of this PR to explain why we are downgrading.

PeterSenpai:
> You need to update the description of this PR to explain why we are downgrading.

Sure, I just added a description.

PR #367, opened 2020-05-15
add offer tests

PeterSenpai:


frontend/src/api/mockAPI/assignments.js
```diff
@@ -316,12 +317,21 @@ class ActiveOffer extends MockAPIController {
...
-                status: "pending"
+                status: "provisional"
             })
         );
+
+        if (!this.data.offer) {
+            this.data.offers = [newOffer];
+        } else {
+            this.data.offer.push(newOffer);
+        }
```
wisemandenny: I'm not sure how you feel about boolean short circuit operations / oneliners with regards to readability, but you could write this as:
`this.data.offer ? this.data.offer.push(newOffer) : this.data.offers = [newOffer]`

frontend/src/api/mockAPI/positions.js
```diff
@@ -59,6 +59,9 @@ export class Position extends MockAPIController {
         });
     }
     getInstructors(position) {
+        if (!position.instructor_ids) {
+            return [];
+        }
         return position.instructor_ids.map(id =>
             new Instructor(this.data).find({ id })
         );
```
wisemandenny, line 67: `return !position.instructor_ids ? [] : position.instructor_ids.map(id => new Instructor(this.data).find({ id }))`

I write lots of one liners but I don't know if that's the preferred syntax for this project, so it's just an alternative.

frontend/src/api/mockAPI/utils.js
```diff
@@ -346,6 +346,9 @@ function joinWithConjunction(items) {
  * @returns {string}
  */
 export function formatInstructorsContact(instructors) {
+    if (!instructors) {
+        return [];
+    }
```
ilikeballoons, line 351: `!instructors && return []`

frontend/src/tests/api.test.js
```diff
@@ -83,6 +83,9 @@ describe("Mock API tests", () => {
     describe("`/admin/assignments` tests", () => {
         assignmentsTests(mockAPI);
     });
+    describe("offers tests", () => {
```
wisemandenny: `"offers tests"` should most likely match the `"/foo/bar"` syntax of other tests

frontend/src/tests/offer-tests.js
```diff
@@ -1,14 +1,121 @@
...
+        applicant = databaseSeeder.seededData.applicant;
+        position = databaseSeeder.seededData.position;
+    }, 30000);
+
+    it("create an active offer", async () => {
+        // checking not active offer before create
+        const resp = await apiGET(
+            `/admin/assignments/${assignment.id}/active_offer`
+        );
+        expect(resp.payload).toEqual(null);
```
wisemandenny: `expect(resp.payload).toBeNull();`

wisemandenny (commented):
I don't see any problems, just a few places that the code could (optionally) be shortened using boolean short circuit operators and/or ternary operators.

frontend/src/api/mockAPI/assignments.js
```diff
@@ -316,12 +317,21 @@ class ActiveOffer extends MockAPIController {
...
+        const newOffer = this.find(
             this.create({
                 assignment_id: matchingAssignment.id,
-                status: "pending"
+                status: "provisional"
             })
         );
+
+        if (!this.data.offer) {
+            this.data.offers = [newOffer];
```
siefkenj: Make sure `.offers` is in the original data-structure and then we can remove this.

frontend/src/api/mockAPI/assignments.js
siefkenj: I'm very against that. It makes it *much* harder to read. The ternary operator can be used for setting variables (e.g. `x = foo ? "yes" : "no"`), but shouldn't be used for code that has side effects.

frontend/src/api/mockAPI/positions.js
PeterSenpai: I actually don't have any preferences regarding the ternary operator(I go with everyone else's preferences lol). But for the same level of performance, I usually write something more straightforward and more readable (mostly for me).

frontend/src/tests/api.test.js
PeterSenpai: good point, I ll change that

frontend/src/tests/offer-tests.js
PeterSenpai: that's better.

frontend/src/api/mockAPI/assignments.js
PeterSenpai: sure, I 'll have to modify the constructor then?

siefkenj: No...it should just be there. In `data.js`

PeterSenpai: I thought it was but if I remove the check it throws an error and if I console log it, it says undefined. I ll take a closer look later.

frontend/src/tests/api.test.js
PeterSenpai: so all the routes that related to offers are  under assignments for example: `/admin/assignments/${assignment_id}/active_offer`
I don't know if I should just leave it as `offer tests` or change to `/xxx/yyy` since it's technically part of assignments. What do you think, @ilikeballoons ? and what do you suggest, @siefkenj ?

frontend/src/api/mockAPI/assignments.js
PeterSenpai: ok, I found a more proper way to update this.

frontend/src/tests/api.test.js
siefkenj: Put the full route in there. The API is interconnected, and that's okay. It's just how it is.

PeterSenpai: sure, np

PR #369, opened 2020-05-29
Use toHaveStatus where applicable

colezemel:
Use our new toHaveStatus extension of expect in all places where we want to check the status of a message


siefkenj:
Does this fix all of them? Can we close #339 ?   (PS: always refer to the issue number in the body-text of your PR so that github will automatically link the issue to this PR)

PR #378, opened 2020-06-11
Add test for wage chunk

CarolineYao:


backend/app/controllers/api/v1/admin/assignment_wage_chunks_controller.rb
```diff
@@ -10,24 +10,39 @@ def index
...
-            AssignmentWageChunkCreateService.new(
-                assignment: @assignment,
-                wage_chunk_params: wage_chunks_create_params
-            )
         start_transaction_and_rollback_on_exception do
-            service.perform
-            render_success service.values
+            result = wage_chunks_create_params.map do |wage_chunk_params|
+                @wage_chunk = WageChunk.find_by(id: wage_chunk_params[:id])
+                if @wage_chunk
```
CarolineYao: Move it to service

backend/app/controllers/api/v1/admin/assignment_wage_chunks_controller.rb
```diff
@@ -10,24 +10,29 @@ def index
...
         start_transaction_and_rollback_on_exception do
-            service.perform
-            render_success service.values
+            result = wage_chunks_create_params.map do |wage_chunk_params|
+                service = AssignmentWageChunkCreateService.new(wage_chunk_params)
+                service.upsert
+                service.wage_chunk
+            end
+
+            delete_extra_wage_chunks(result.map(&:id))
```
siefkenj, line 25: Can you add a comment explaining what/why is going on here?

frontend/src/tests/wage-chunk-tests.js
```diff
@@ -1,12 +1,221 @@
...
+
+        const wageChunkData = {
+            hours: newPositionData.hours_per_assignment,
+        };
+
+        // make sure the position we created is in that list
+        expect(resp.payload).toContainObject(wageChunkData);
+
+        checkPropTypes(PropTypes.arrayOf(wageChunkPropTypes), resp.payload);
+    });
```
siefkenj, line 228: Is there a test to ensure when setting the wage chunks via an array (on the `assignments` route), that the non-listed wage chunks are deleted?

frontend/src/tests/wage-chunk-tests.js
```diff
@@ -1,12 +1,221 @@
...
-    it.todo("update wage_chunk");
-    it.todo(
-        "automatic creation of wage chunk for position when `hours` is set"
-    );
+export function wageChunksTests(api) {
+    const { apiGET, apiPOST } = api;
+    const session = databaseSeeder.seededData.session,
+        applicant = databaseSeeder.seededData.applicant,
+        contractTemplate = databaseSeeder.seededData.contractTemplate,
+        assignment = databaseSeeder.seededData.assignment;
```
wisemandenny: const { session, applicant, contractTemplate, assignment } = databaseSeeder.seededData;

frontend/src/tests/wage-chunk-tests.js
```diff
@@ -1,12 +1,221 @@
...
+        const wageChunkData = {
+            hours: assignment.hours,
+        };
+
+        // make sure the position we created is in that list
+        expect(resp.payload).toContainObject(wageChunkData);
+
+        checkPropTypes(PropTypes.arrayOf(wageChunkPropTypes), resp.payload);
+
+        originalWageChunk = resp.payload[0];
```
wisemandenny, line 44: What if the test fails before this line is executed? Will that not screw up future tests that depend on `originalWageChunk` being a certain value?

backend/app/controllers/api/v1/admin/wage_chunks_controller.rb
```diff
@@ -3,22 +3,28 @@
 class Api::V1::Admin::WageChunksController < ApplicationController
     # POST /wage_chunks
     def create
-        wage_chunk_id =
-            WageChunk.upsert!(wage_chunks_create_params, returning: %i[id])
-        render_succcess WageChunk.find(wage_chunk_id)
+        upsert
```
colezemel: The positions and sessions positions controller do the same thing but can we just get rid of the `create` function here if all it does is call the `upsert` function?

colezemel (changes requested):

backend/app/controllers/api/v1/admin/wage_chunks_controller.rb
CarolineYao: That is a good point. However, I am under the impression that `create` function is the more standard name; I have moved the `upsert` function body inside of `create` function, so that `create` function still do the same thing without calling another function

frontend/src/tests/wage-chunk-tests.js
CarolineYao: I have done that by:

- (line 111) ensure that the list of new wage chunks contains all wage chunks we have sent to the API
- (line 131) mark sure that only 2 wage chunks are returned.

But I agree that it is not direct enough to show the non-listed wage chunks are no longer there, so I have added line 118 to make sure that non-listed wage chunk's id is not in the `payload`.

Had to change the mock API a little bit to let it pass the test because the real API upserts the listed wage chunks first and then delete the extra ones (so that if a new wage chunk is added, the `id` will be different than the non-listed one) while the mock API delete the extra wage chunks before upserting listed wage chunks. Both way works, but for the sake of testing, I am changing the mock API to match real API because IMO comparing `id` shows more directly that the non-listed one is deleted instead of updated.

frontend/src/tests/wage-chunk-tests.js
CarolineYao: We use `originalWageChunk` as a template to create new wage chunk object so that if any required field is added to wage chunk, the value from `originalWageChunk` apply to the new wage chunk and it won't affect the test. 

It is also used to check if `originalWageChunk` is still in the assignment wage chunk list.

The place where it can screw up future tests would be if `get` method does not return anything. In this case, it would be fine for all the wage chunk test to fail since every wage chunk test case calls the `get` method at the end.

PR #379, opened 2020-06-15
Human readable dates

wisemandenny:
This PR affects the Sessions and Positions views
It modifies the way that dates are displayed from `2019-01-01T00:00:00.000Z` -> `January 1, 2019`.
It adds a formatDate method in the utils file which can be used throughout the program.

### DIFF
--- a/frontend/src/components/sessions.js
+++ b/frontend/src/components/sessions.js
@@ -6,8 +6,16 @@ import { docApiPropTypes } from "../api/defs/doc-generation";
 import { formatDate } from "../libs/utils";
 const DEFAULT_COLUMNS = [
     { Header: "Name", accessor: "name" },
-    { Header: "Start", accessor: "start_date" },
-    { Header: "End", accessor: "end_date" },
+    {
+        Header: "Start",
+        accessor: "start_date",
+        Cell: (row) => formatDate(row.value),
+    },
+    {
+        Header: "End",
+        accessor: "end_date",
+        Cell: (row) => formatDate(row.value),
+    },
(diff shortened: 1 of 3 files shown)


frontend/src/libs/utils.js
```diff
@@ -11,3 +11,14 @@ export function strip(x) {
...
+
+/**
+ * Formats the input date string to be human readable
+ * Input string is of the form 2019-01-01T00:00:00.000Z
+ * Output string is of the form January 1, 2019
+ * @param {string} dateString 
+ */
+export function formatDate(dateString) {
+    const date = new Date(dateString);
+    return `${date.toLocaleDateString("en-US", { month: 'long', year: 'numeric', day: 'numeric' })}`;
```
siefkenj: Let's use the Canadian locale. Also, 3-letter months is more compact.

frontend/src/components/sessions.js
```diff
@@ -21,11 +25,20 @@ const DEFAULT_COLUMNS = [
...
+                end_date: formatDate(session.end_date)
+            }))
+            : sessions;
+
     return (
         <React.Fragment>
             <h3>Sessions</h3>
             <ReactTable
-                data={sessions}
+                data={formattedSessions}
```
siefkenj: I think `ReactTable` can do individual cell formatters. If you format in the cell, then the data should stay sortable.

frontend/src/libs/utils.js
wisemandenny: I've made this change

Issue #376, opened 2020-06-05
UI Issues


siefkenj:
Here's a list of UI issues that needs to be addressed.

- [x] Remove TAPP/CP toggle (it's just TAPP); Make a hierarchical tab structure instead
  - `Admin` (this what is currently called TAPP)
  - Matching
  - Applications and Postings  
- [x] Landing page for `Admin`:
   - If no there is no active session, it is prominently displayed first; if there is an active session, show what session is selected and how to select a different one.
   - Brief descriptions of what the tabs do.
- [x] `Sessions`: Dates formatted in readable way
- [x] `Sessions`: Rates are editable
- [x] Restyle the "Add/Edit" buttons in a consistent way across pages (so they look good from a UI/UX perspective).
- [x] `Contract Templates` Ability to upload new template
- [x] `Contract Templates` Ability to preview template
- [x] `Contract Templates` Ability to download
- [x] `Instructors` "Add Instructor" dialog needs email
- [x] `Instructors` Instructors table needs a delete option
- [x] `Instructors` Instructors page needs an import/export option
- [x] `Positions` Readable dates
- [x] `Positions` Editable fields
- [x] `Positions` A sophisticated view (the spreadsheed doesn't and shouldn't include duties, etc., but we need to be able to see/edit those fields).
- [x] `Positions` Show `total number of hours`, `enrollment`, `waitlist`, and `number of TAs currently assigned`
- [x] `Assignments` Filter doesn't work on status
- [x] `Assignments` UI needs to be rethought (button placement, menus, etc.)
- [x] `Assignments` actions that aren't available should be disabled
- [x] `Assignments` should be able to see if an offer was withdrawn (instructor view)
- [x] `Summary` delete this page
- [x] `Import/Export` interface for instructions, positions, assignments.
- [x] `Applicants` view is missing.
- [x] `Applicants` should only be able to be deleted if they have no assignments
Longer term:
- [ ] Add `Reporting and Statistics` tab

wisemandenny:
@siefkenj 
> Sessions: Rates are editable

Which rate? Do you mean `model/sessions/active_session/rate1` (and `rate2`) or do you mean `model/sessions/_model_data/<array element index>/rate1` and (`rate2`)

siefkenj:
You will never touch `_modelData`. When you make an API call, it is automatically updated to be in sync with the backend. So, you will make API calls for `rate1` and `rate2`.

wisemandenny:
[...]

PR #380, opened 2020-06-16
Hierarchical navbar

wisemandenny:
This PR modifies the navbar component at the top of the main page as described in #376.


frontend/src/components/components.css
```diff
@@ -17,3 +17,17 @@
     border-radius: 15px;
     border: 2px solid rgb(0, 95, 204);
 }
+
+.navbar-tabs {
+    padding: 1rem 1rem 0 0.5rem !important;
```
siefkenj: I'm always skeptical when I see `!important`. Also, do you have a good reason for using `rem`s?

frontend/src/components/header.js
```diff
@@ -21,6 +21,14 @@ BootstrapNavLink.propTypes = {
     to: PropTypes.string,
 };
 
+function BootstrapNavItem(props) {
```
siefkenj, line 13: Is this needed? Also, all components should have a doc string, even if it's just `A wrapper for BootstrapNavLink so that the item fits in a nav bar`, or something like that...

frontend/src/components/header.js
```diff
@@ -42,65 +50,63 @@ BootstrapNavLink.propTypes = {
  * @param {object[]} props.routes
  * @returns
  */
+
 export function Header(props) {
     const { routes = [], infoComponents = null } = props;
+    const [key, setKey] = React.useState(routes[0].route);
```
siefkenj: So, this means deep linking is impossible?

frontend/src/components/header.js
```diff
@@ -42,65 +50,63 @@ BootstrapNavLink.propTypes = {
...
-                >
-                    {routes
-                        .filter((route) => !route.hidden)
-                        .map((route) => (
-                            <NavDropdown.Item
-                                key={route.route}
-                                as="span"
-                                tabIndex="0"
+        <Tab.Container activeKey={key} onSelect={(k) => setKey(k)}>
+            <Row className="justify-content-between right-padding">
```
siefkenj: I don't like these classnames. It's one thing if you are building a general framework for widget layout (which you shouldn't be, because that's the point of bootstrap...), but I think in this situation, they should be widget-specific. E.g., `nav-bar-item`

frontend/src/components/header.js
```diff
@@ -42,65 +50,63 @@ BootstrapNavLink.propTypes = {
...
+                        ))}
+                    </Row>
+                </Col>
+            </Row>
+            <Row className="navbar-subtabs">
+                <Tab.Content>
+                    <Nav variant="pills">
+                        {routes
+                            .filter((route) => route.route === key)
+                            .map((route) =>
```
siefkenj: I would rather this logic be done before the return statement, and then just included here. E.g.,

```
const activeTabs = routes.filter ...;

...

<Nav>
   {activeTabs}
</Nav>
```

frontend/src/views/admin/header.js
```diff
@@ -73,10 +73,10 @@ function AdminHeader() {
                     ],
                 },
                 {
-                    route: "/dashboard",
-                    name: "Dashboard",
-                    description: "List of all widgets",
-                    hidden: true,
+                    route: "/appointments_positions",
```
siefkenj: -> `appointments_and_positions`

frontend/src/views/admin/routes.js
```diff
@@ -0,0 +1,56 @@
...
+        name: "CP",
+        description: "Matching", //TODO: better desc
+        subroutes: [
+            {
+                route: "/statistics",
+                name: "Statistics",
+                description: "See statistics about accepted/rejected contracts",
+            },
+        ],
+    },
```
siefkenj: Let's just delete this for now.

frontend/src/components/header.js
wisemandenny: The `justify-content-between` is a builtin bootstrap classname.
`right-padding` -- I agree it's annoying, I need to dig through the react-bootstrap classnames documentation to figure out how to give this element right padding.

frontend/src/components/components.css
wisemandenny: I prefer using `em` and `rem` over `px` because it is responsive.
The usage of `!important` here is a shortcut. I am not very familiar with bootstrap's classname-method of css formatting, so I took a shortcut. I will fix this so that this class doesn't even need to exist (I will use bootstrap builtins)

frontend/src/components/header.js
wisemandenny: Now that I look at the code again, I have no idea why I did it this way.

wisemandenny: also, the `BootstrapNavLink` component isn't used in any other files. In fact, it is redefined in `frontend/src/views/dev_frame/index.js`, so I'm not sure why this function is exported at all. I will rework it and `BootstrapNavItem` into one function.

frontend/src/components/header.js
wisemandenny: Sure, makes sense to me.

frontend/src/components/header.js
```diff
@@ -10,17 +10,29 @@ import { Route, NavLink } from "react-router-dom";
...
+BootstrapNavItem.propTypes = {
     to: PropTypes.string,
+    active: PropTypes.bool,
 };
 
+/**
+ * Returns the top level route for a full path ("/tapp/sessions" -> "/tapp")
+ * @param {string} pathname
+ */
+function getRoute(pathname) {
```
siefkenj: You shouldn't be doing route parsing manually...`react-router-dom` should have all these functions built in.

frontend/src/components/header.js
```diff
@@ -42,65 +54,84 @@ BootstrapNavLink.propTypes = {
...
+            return (
+                <BootstrapNavItem
+                    eventKey={route.route}
+                    to={route.route}
+                    key={route.route}
+                >
+                    <h5
+                        className={
+                            active
+                                ? "font-weight-bold text-primary"
```
siefkenj: This is the same as embedding styles directly...come up with an appropriate class name, e.g., `nav-link.active` and use that.

frontend/src/components/header.js
wisemandenny: i am using react-router-dom to get the path, getRoute just clips off anything after the top level route (as described in the docstring)

siefkenj: Yes, but they should have facility for deep linking. If you set things up as subroutes, it should all work without any parsing.

siefkenj: https://stackoverflow.com/questions/41474134/nested-routes-with-react-router-v4-v5

frontend/src/views/admin/header.js
```diff
@@ -11,74 +11,70 @@ import {
...
+        name: "Matching",
+        description: "Matching",
+        subroutes: [
+            {
+                route: "/statistics",
+                name: "Statistics",
+                description: "See statistics about accepted/rejected contracts",
+            },
+        ],
+        // hidden: true,
```
siefkenj: Why are these comments here?

wisemandenny: removed them

frontend/src/components/header.js
siefkenj: Are you still working on this? I don't see it changed yet.

frontend/src/components/header.js
```diff
@@ -42,65 +44,77 @@ BootstrapNavLink.propTypes = {
...
+            <h5>{route.name}</h5>
+        </BootstrapNavItem>
+    ));
+
+    const activeSubRoutes = routes
+        .filter((route) => route.route.substring(1) === mainRoute)
+        .map((route) =>
+            (route.subroutes || []).map((subroute) => {
+                const fullroute = `${route.route}${subroute.route}`;
+                console.log(route, subroute);
```
siefkenj: Excess log

frontend/src/components/header.js
```diff
@@ -42,65 +44,77 @@ BootstrapNavLink.propTypes = {
...
+            eventKey={route.route}
+            to={route.route}
+            key={route.route}
+            className="primary"
+        >
+            <h5>{route.name}</h5>
+        </BootstrapNavItem>
+    ));
+
+    const activeSubRoutes = routes
```
siefkenj: This is really confusing...Maybe `availableSubRoutes` is a better name. Also, add a comment explaining what/why is going on.

frontend/src/components/header.js
```diff
@@ -42,65 +44,77 @@ BootstrapNavLink.propTypes = {
...
-                                    title={route.description}
-                                >
-                                    {route.name}
-                                </BootstrapNavLink>
-                            </NavDropdown.Item>
+        <Tab.Container activeKey={mainRoute} defaultActiveKey={"tapp"}>
+            <Row className="justify-content-between pr-3 ">
+                <Col md={"auto"}>
+                    <Nav
+                        className="flex-row pt-3 pr-3 pb-0 pl-2"
```
siefkenj: Create a new class name and move this to CSS. Only use the inline-bootstrap styles for minor things (e.g., you only need one or two bootstrap classes)

frontend/src/components/header.js
```diff
@@ -42,65 +44,77 @@ BootstrapNavLink.propTypes = {
...
+                        variant="tabs"
+                    >
+                        {activeMainRoutes}
+                    </Nav>
+                </Col>
+                <Col md={"auto"}>
+                    <Row>
+                        {infoComponents.map((component, index) => (
```
siefkenj, line 98: This will fail if `infoComponents` is `null`. Also, what are `infoComonents`?

wisemandenny: Infocomponents are the session selector and the user selector ui components. they were always passed in to header so I didn't check for null.

I just changed

`const { infoComponents = null } = props` to `const (infoComponents = []} = props`

siefkenj: Sounds good.

PR #388, opened 2020-06-30
Positions improvements

wisemandenny:
<h1>More details in Positions page</h1>

<h2>This PR addresses some points from Issue #376:</h2>
- `Positions` A sophisticated view (the spreadsheed doesn't and shouldn't include duties, etc., but we need to be able to see/edit those fields).
- `Positions` Show `total number of hours`, `enrollment`, `waitlist`, and `number of TAs currently assigned`

<h3>New features:</h3>

- <h4>Simple/Advanced Toggle</h4>
 
  - Simple view shows position code, position title, hours, start/end dates and instructors.

  - Advanced view loads all columns which have at least 1 piece of data in them for the current session
    - Displays a button toolbar which shows which columns are enabled and disabled currently
    - Users can choose which columns they want to display
    - If the user switches back to simple mode, only the selected columns revert to default
 - All information from the database can be displayed in the table if the user chooses to enable all columns
    - ~Color coding for `instructor_preference_level`~

I will write documentation for this code tomorrow.

### DIFF
--- a/frontend/src/components/positions-list.js
+++ b/frontend/src/components/positions-list.js
@@ -40,6 +40,65 @@ const DEFAULT_COLUMNS = [
         accessor: "contract_template.template_name",
     },
 ];
+function getPreferenceLevelColor(preferenceLevel) {
+    return ["danger", "warning", "primary", "info", "success"][
+        preferenceLevel + 1
+    ];
+}
+const ADVANCED_COLUMNS = [
+    {
+        Header: "Duties",
+        accessor: "duties",
+    },
+    {
+        Header: "Instructor Preferences",
+        accessor: "instructor_preferences",
+        Cell: (props) => (
+            <React.Fragment>
+                {props.value.map((preference = {}, index) => {
+                    const name = `${preference.instructor.first_name} ${preference.instructor.last_name}`;
+                    return (
+                        <Badge
+                            variant={getPreferenceLevelColor(
+                                preference.preference_level
+                            )}
+                            className="mr-1"
+                            key={`${name}-${index}`}
+                        >
+                            {name}
+                        </Badge>
+                    );
+                })}
+            </React.Fragment>
+        ),
+    },
+    {
+        Header: "Qualifications",
+        accessor: "qualifications",
+    },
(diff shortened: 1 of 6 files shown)


frontend/src/components/positions-list.js
```diff
@@ -39,8 +52,81 @@ const DEFAULT_COLUMNS = [
         Header: "Contract Template",
         accessor: "contract_template.template_name",
     },
+    {
+        Header: "Duties",
+        accessor: "duties",
+    },
+    {
+        Header: "Instructor Preferences",
```
siefkenj, line 60: Instructor preferences are for TAs, not for positions. This doesn't belong here...

wisemandenny: I had just added all the columns from the api doc for GET positions. If there are any else that shouldn't be viewable, please let me know so I can add them to the blacklst

frontend/src/views/positions/positions-list-toolbar.js
```diff
@@ -0,0 +1,148 @@
...
+        ...new Set(
+            positions
+                .reduce(
+                    (acc, position) => [...acc, ...Object.keys(position)],
+                    []
+                )
+                .filter((heading) => !BLACKLIST.includes(heading))
+        ),
+    ];
+}
```
siefkenj, line 32: I don't like magic...Let's list these out by hand. (The order of columns is also very important and must be specified by hand anyways.)

frontend/src/views/positions/positions-list-toolbar.js
```diff
@@ -0,0 +1,148 @@
...
+                    variant={
+                        isSelected ? "outline-primary" : "outline-secondary"
+                    }
+                >
+                    {isSelected ? "✓" : "❌"}
+                </Badge>
+            </ToggleButton>
+        );
+    });
+}
```
siefkenj, line 65: We don't want dynamic setting of columns. A "full information" button that opens a dialog with all the information would be good though. Such a dialog should tap into a new view made especially for showing/editing information about a single position.

PR #389, opened 2020-07-01
Instructor Delete Button

wisemandenny:
This PR adds a column in the instructors view which contains a "Delete" button. When this button is clicked, the instructor for the row is deleted.

~I am not sure if I need to confirm the delete (i.e. prompt the user "Are you sure?"), if so, I will add this to the feature branch.~

Issue #376

### DIFF
--- a/frontend/src/views/instructors/delete-instructor-dialog.js
+++ b/frontend/src/views/instructors/delete-instructor-dialog.js
@@ -1,0 +1,37 @@
+import React from "react";
+import { connect } from "react-redux";
+import { deleteInstructor, instructorsSelector } from "../../api/actions";
+import { Modal, Button } from "react-bootstrap";
+
+function DeleteInstructorDialog(props) {
+    const { show, onHide = () => {}, id, deleteInstructor } = props;
+    function removeInstructor() {
+        deleteInstructor({ id });
+        onHide();
+    }
+    return (
+        <Modal show={show} onHide={onHide}>
+            <Modal.Header closeButton>
+                <Modal.Title>Delete Instructor</Modal.Title>
+            </Modal.Header>
+            <Modal.Body>
+                Press OK to confirm deleting this instructor.
+            </Modal.Body>
+            <Modal.Footer>
+                <Button onClick={onHide} variant="secondary">
+                    Cancel
+                </Button>
+                <Button onClick={removeInstructor}>OK</Button>
+            </Modal.Footer>
+        </Modal>
+    );
+}
+
+export const ConnectedDeleteInstructorDialog = connect(
+    (state) => ({
+        instructors: instructorsSelector(state),
+    }),
+    {
+        deleteInstructor,
+    }
+)(DeleteInstructorDialog);
--- a/frontend/src/views/instructors/editable-instructors-list.js
+++ b/frontend/src/views/instructors/editable-instructors-list.js
@@ -1,4 +1,5 @@
 import React from "react";
+import { FaWindowClose } from "react-icons/fa";
 import { connect } from "react-redux";
 import { Button } from "react-bootstrap";
 import {
@@ -59,16 +60,17 @@ function EditableInstructorsList(props) {
         const { deleteOnClick, setDeleteInstructorId } = props;
         const disabled = workingInstructors.has(id);
 
-        return (
+        return disabled ? null : (
             <Button
-                variant="danger"
+                className="close"
+                title={`delete-instructor-${id}`}
                 onClick={() => {
                     setDeleteInstructorId(id);
                     deleteOnClick();
                 }}
                 disabled={disabled}
             >
-                <span>&times;</span>
+                <FaWindowClose />
             </Button>
         );
     }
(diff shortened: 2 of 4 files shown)


frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -63,6 +94,10 @@ function EditableInstructorsList(props) {
             accessor: "utorid",
             Cell: generateCell("utorid"),
         },
+        {
+            Header: "Delete",
+            Cell: generateDeleteCell(),
```
siefkenj: Why the indirection? Just define
```
function DeleteCell(props) { ... }
```
and then do `Cell: DeleteCell`

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -1,9 +1,34 @@
...
+/**
+ * A cell that renders a button for deleting the instructor from that row.
+ *
+ * @param {*} props
+ * @returns a button which when clicked deletes the instructor.
+ */
+function DeleteButtonCell(props) {
+    const {
+        deleteInstructor,
+        original: { id: applicantId },
```
siefkenj: De-structuring more than one level is hard to read (and generates confusing error messages). Refactor this to only destructure one level deep.

wisemandenny:
~deleting instructors who are assigned positions will fail, so need to filter the list of instructors and find the ones that aren't "deleteable" and disable their delete button~

- implemented in commit 0844bd6

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,6 +55,21 @@ function EditableInstructorsList(props) {
         );
     }
 
+    function DeleteButtonCell({ original: instructorInfo }) {
+        const { id } = instructorInfo;
+        const disabled = workingInstructors.has(id);
```
siefkenj, line 61: `workingInstructors` sounds like "functioning instructors". Maybe `assignedInstructors`?

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,6 +55,21 @@ function EditableInstructorsList(props) {
         );
     }
 
+    function DeleteButtonCell({ original: instructorInfo }) {
```
siefkenj, line 58: Add a comment explaining the strange use of `original`

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,6 +55,21 @@ function EditableInstructorsList(props) {
...
+        const { id } = instructorInfo;
+        const disabled = workingInstructors.has(id);
+
+        function onClick() {
+            deleteInstructor({ id });
+        }
+
+        return (
+            <Button variant="danger" onClick={onClick} disabled={disabled}>
+                <span>&times;</span>
```
siefkenj: Use the `Fa`icon for this.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,6 +55,21 @@ function EditableInstructorsList(props) {
...
+    function DeleteButtonCell({ original: instructorInfo }) {
+        const { id } = instructorInfo;
+        const disabled = workingInstructors.has(id);
+
+        function onClick() {
+            deleteInstructor({ id });
+        }
+
+        return (
+            <Button variant="danger" onClick={onClick} disabled={disabled}>
```
siefkenj: I don't know what it looks like, but I'm guessing it stands out? Make the delete button appear muted until you hover. Also, all buttons should have a `title`

wisemandenny: Disabled greys out the icon:
image

frontend/src/views/instructors/editable-instructors-list.js
wisemandenny: can you point me to an example of this?
global project search of "Fa" might take a while 😋

frontend/src/views/instructors/editable-instructors-list.js
siefkenj: That's still too much. Just have a grey "X" until you hover and then have it turn red. We will very rarely be deleting instructors.

wisemandenny: OK. Do you still want a column for it? It looks weird with a big column with just a tiny "x" inside.

frontend/src/views/instructors/delete-instructor-dialog.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import { connect } from "react-redux";
+import { deleteInstructor, instructorsSelector } from "../../api/actions";
+import { Modal, Button } from "react-bootstrap";
+
+function DeleteInstructorDialog(props) {
+    const { show, onHide = () => {}, id, deleteInstructor } = props;
+    function removeInstructor() {
+        deleteInstructor({ id });
```
wisemandenny, line 9: is this the deleteInstructor from props or the one that is imported on line 3?

siefkenj: From props

siefkenj:
In light of PR #397, I suggest a slight modification. Now there is a toggle switch "Delete instructors". Let's hide the "x" buttons when the toggle is inactive and when it is active, let's show the "x" column.

PR #395, opened 2020-07-05
Assignments Export

siefkenj:
Adds Import/Export ability to the Assignments page.

This PR also introduces some typescript files. Typescript and Javascript can be mixed between files, so we can gradually convert the files that we find useful to typescript.


siefkenj:
Addresses part of #376

PR #396, opened 2020-07-05
Navbar Refactor

siefkenj:
Refactors the navbar and added new styling

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -18,7 +18,9 @@ class App extends React.Component {
         return (
             <React.Fragment>
                 <AdminHeader />
-                <AdminRoutes />
+                <div className="view-container">
+                    <AdminRoutes />
+                </div>
                 <ConnectedNotifications />
             </React.Fragment>
         );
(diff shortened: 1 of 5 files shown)


frontend/src/App.jsx
```diff
@@ -18,7 +18,9 @@ class App extends React.Component {
         return (
             <React.Fragment>
                 <AdminHeader />
-                <AdminRoutes />
+                <div className="view-container">
+                    <AdminRoutes />
```
wisemandenny, line 22: does this not work if you do `<AdminRoutes className="view-container" />`?

frontend/src/components/header.js
```diff
@@ -47,7 +49,9 @@ BootstrapNavItem.propTypes = {
 
 export function Header(props) {
     const { routes = [], infoComponents = [] } = props;
-    let match = useRouteMatch("/:mainRoute/:subRoute?");
+    let match = useRouteMatch("/:mainRoute/:subRoute?") || {
+        params: { mainRoute: "tapp" },
```
wisemandenny, line 53: of course the way I wrote it didn't work... good catch

wisemandenny (commented):
Looks good. I like the refactor of the header component so that it doesn't need rows and cols.

frontend/src/App.jsx
siefkenj: I believe `AdminRoutes` returns a `React.Fragment`, so it cannot have a class name.

PR #400, opened 2020-07-10
Delete instructors

wisemandenny:
Revised PR for Deleting instructors from the instructors view
- Click delete instructors in the action area to enter delete mode
- Delete icons appear beside the instructors in the table, but only if they are not assigned to any position in this session
- Clicking the delete icon opens a confirmation dialog
- Clicking OK in the confirmation dialog deletes the instructor, otherwise closes the dialog.
#376


wisemandenny:
Sorry for closing and reopening... My finger slipped while using the GitHub app

frontend/src/views/instructors/delete-instructor-dialog.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import { connect } from "react-redux";
+import { deleteInstructor, instructorsSelector } from "../../api/actions";
+import { Modal, Button } from "react-bootstrap";
+
+function DeleteInstructorDialog(props) {
+    const { show, onHide, deleteInstructor, instructorToDelete } = props;
+    const { first_name, last_name, id } = instructorToDelete;
```
siefkenj: Let's call this `instructor` and have an `onHide` and `onDelete`.

frontend/src/views/instructors/delete-instructor-dialog.js
```diff
@@ -0,0 +1,39 @@
...
+                </Button>
+            </Modal.Footer>
+        </Modal>
+    );
+}
+
+export const ConnectedDeleteInstructorDialog = connect(
+    (state) => ({ instructors: instructorsSelector(state) }), // don't need instructors anymore?
+    { deleteInstructor }
+)(DeleteInstructorDialog);
```
siefkenj: Don't make a connected component. We want to connect as few things as possible. A simple `onDelete` handler that gets passed the instructor will let the parent decide how to deal with the delete event.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -29,7 +35,15 @@ function EditableCell(props) {
 }
 
 function EditableInstructorsList(props) {
-    const { upsertInstructor, ...rest } = props;
+    const {
+        upsertInstructor,
+        inDeleteMode,
+        setInDeleteMode,
+        setDeleteDialogVisible,
+        setInstructorToDelete,
```
siefkenj: Let's have `EditableInstructorsList` handle the deleting with internal state. Pass in a `deleteInstructor` prop, but all the `set*Visible...` stuff will be internal to the component.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,6 +56,14 @@ function EditableInstructorsList(props) {
         );
     }
 
+    const workingInstructors = new Set(
```
siefkenj: What is this variable name trying to say? Pick a better one.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,6 +56,14 @@ function EditableInstructorsList(props) {
...
+        positions
+            .map((position) =>
+                position.instructors.map((instructor) => instructor.id)
+            )
+            .flat()
```
siefkenj: `flat` is a very new feature. Avoid it where you can.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -65,7 +87,34 @@ function EditableInstructorsList(props) {
...
+        Header: "Delete",
+        Cell: (props) => {
+            const instructor = props.original;
+            const disabled = workingInstructors.has(instructor.id);
+            return disabled ? null : (
+                <Button
+                    onClick={() => deleteButtonOnClick(instructor)}
+                    variant="outline-danger"
+                >
+                    <FaTrash />
```
siefkenj: We already used a `FaTrash` in the side panel. Let's use an `X` icon here so the user doesn't get overwhelmed.

frontend/src/views/instructors/editable-instructors-list.js
siefkenj: A simple hash will do just as well as a set and be simpler to understand.

frontend/src/views/instructors/index.js
```diff
@@ -12,10 +12,30 @@ import {
...
+    utorid: "",
+    id: 0,
+};
 
 export function AdminInstructorsView() {
     const [addDialogVisible, setAddDialogVisible] = React.useState(false);
     const [inDeleteMode, setInDeleteMode] = React.useState(false);
+    const [deleteDialogVisible, setDeleteDialogVisible] = React.useState(false);
+    const [instructorToDelete, setInstructorToDelete] = React.useState(
+        BLANK_INSTRUCTOR
```
siefkenj: What? Why....? Use `null` to initialize and do appropriate validation checks before actually deleting.

frontend/src/views/instructors/index.js
```diff
@@ -12,10 +12,30 @@ import {
...
+    );
+
+    React.useEffect(() => {
+        if (!inDeleteMode || !deleteDialogVisible) {
+            setInstructorToDelete(BLANK_INSTRUCTOR); // reset instructor to delete if not in delete mode or confirm dialog closes
+        }
+    }, [inDeleteMode, deleteDialogVisible]);
```
siefkenj: This will disapear when `InstructorList` manages the delete dialog.

frontend/src/views/instructors/editable-instructors-list.js
wisemandenny: Need a bit of clarification here: when you say "...will be internal to the component", do you mean `Editable-Instructors-List`? Because I cannot put some of the visibility state manipulations in there since the `delete-instructor-dialog` component is inside the `index.js` view, and so the state needs to be visible from `index.js` in order to be passed into the `delete-instructor-dialog`. Could you elaborate a bit what you mean here?

siefkenj: put the `DeleteInstructorDialog` inside this component. Wrap everything in a `React.Fragment`

frontend/src/views/instructors/editable-instructors-list.js
wisemandenny: > A simple hash will do just as well as a set and be simpler to understand.

What do you mean?

siefkenj: `instructorCurrentlyAssignedHash[utorid] = true`

frontend/src/components/components.css
```diff
@@ -71,3 +71,12 @@
     display: flex;
     align-items: center;
 }
+
+.delete-instructor-button:hover {
+    color: red;
+    cursor: pointer;
```
siefkenj: The cursor shouldn't be set on `:hover` it should be set on the base class.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -29,7 +36,23 @@ function EditableCell(props) {
...
+    const {
+        upsertInstructor,
+        deleteInstructor,
+        inDeleteMode,
+        positions,
+        ...rest
+    } = props;
+
+    const [deleteDialogVisible, setDeleteDialogVisible] = React.useState(false);
+    const [instructorToDelete, setInstructorToDelete] = React.useState();
```
siefkenj: Use `null` as the default value

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -65,7 +118,22 @@ function EditableInstructorsList(props) {
...
     ];
 
-    return <InstructorsList columns={columns} {...rest} />;
+    return (
+        <React.Fragment>
+            <InstructorsList columns={columns} {...rest} />
+            <DeleteInstructorDialog
+                show={deleteDialogVisible}
+                onHide={() => {
+                    setDeleteDialogVisible(false);
```
siefkenj, line 130: Instead of using an effect on the dialog visibility, just make a callback here that does everything you want.

frontend/src/components/components.css
```diff
@@ -71,3 +71,20 @@
...
     align-items: center;
 }
+
+<<<<<<< HEAD
+.delete-instructor-button {
+    cursor: pointer;
+}
+.delete-instructor-button:hover {
+    color: red;
+=======
```
siefkenj: You didn't resolve your merge conflict....

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,7 +58,37 @@ function EditableInstructorsList(props) {
...
+                (instructorCurrentlyAssignedHash[instructor.id] = true)
+        )
+    );
+
     const columns = [
+        {
+            Header: (
+                <FaTrash className="delete-instructor-column-header-icon" />
+            ),
+            Cell: (props) => {
```
siefkenj: Comments to describe what this is and why it's here.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -65,7 +111,23 @@ function EditableInstructorsList(props) {
...
+        <React.Fragment>
+            <InstructorsList columns={columns} {...rest} />
+            <DeleteInstructorDialog
+                show={deleteDialogVisible}
+                onHide={() => {
+                    setDeleteDialogVisible(false);
+                    setInstructorToDelete(null);
+                }}
+                onDelete={() => {
+                    deleteInstructor({ id: instructorToDelete.id });
```
siefkenj: You don't need to pick the `id` prop. You can just pass `instructorToDelete`

frontend/src/views/instructors/delete-instructor-dialog.js
```diff
@@ -0,0 +1,26 @@
...
+                Are you sure you want to delete instructor{" "}
+                {`${instructor.last_name}, ${instructor.first_name}`}? This
+                action cannot be undone.
+            </Modal.Body>
+            <Modal.Footer>
+                <Button onClick={onHide} variant="light">
+                    Cancel
+                </Button>
+                <Button onClick={onDelete} title="Delete Instructor">
+                    OK
```
siefkenj: This should be `Delete`. For the user to understand what `Ok` means, they need to read the dialog text, which is an additional cognitive burden.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,7 +58,38 @@ function EditableInstructorsList(props) {
...
+        )
+    );
+
     const columns = [
+        {
+            Header: (
+                <FaTrash className="delete-instructor-column-header-icon" />
+            ),
+            Cell: (props) => {
+                // Renders a custom cell with a delete button if the instructor is not currently assigned to any position, otherwise nothing
```
siefkenj: This should be a docstring. It would be better to split this out as its own function component.

frontend/src/views/instructors/editable-instructors-list.js
```diff
@@ -42,7 +58,38 @@ function EditableInstructorsList(props) {
...
+                const instructor = props.original;
+                const disabled =
+                    instructorCurrentlyAssignedHash[instructor.id] === true; // if the instructor is currently assigned, don't return a delete button
```
siefkenj: Prefer comments on the line before over comments at the end of a line.

wisemandenny: We should add this to prettify rules.

frontend/src/views/instructors/delete-instructor-dialog.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Modal, Button } from "react-bootstrap";
+
+export function DeleteInstructorDialog(props) {
+    const { show, onHide, onDelete, instructor } = props;
+    return !instructor ? null : ( // needed to catch the error of null instructor
```
siefkenj: I don't understand this comment. Also, move it before the `return`

PR #401, opened 2020-07-11
Add action areas

wisemandenny:
This PR adds the `ActionsList`, and `ContentArea` to all views in the Admin route.
#376

### DIFF
--- a/frontend/src/views/assignments/assignment-details.js
+++ b/frontend/src/views/assignments/assignment-details.js
@@ -9,6 +9,8 @@ import {
     upsertWageChunksForAssignment,
 } from "../../api/actions";
 import { EditableField } from "../../components/edit-field-widgets";
+import { ActionButton } from "../../components/action-buttons";
+import { FaInfoCircle } from "react-icons/fa";
 function AssignmentDetails(props) {
     const {
         assignment,
@@ -89,9 +91,12 @@ function ViewAssignmentDetailsButton(props) {
     const [visibleAssignments, setVisibleAssignments] = React.useState([]);
     return (
         <React.Fragment>
-            <Button onClick={() => setVisibleAssignments(selectedAssignments)}>
+            <ActionButton
+                icon={<FaInfoCircle />}
+                onClick={() => setVisibleAssignments(selectedAssignments)}
+            >
                 View Selected Assignment Details
-            </Button>
+            </ActionButton>
             {visibleAssignments.map((assignment, i) => (
                 <ConnectedAssignmentDetails assignment={assignment} key={i} />
             ))}
--- a/frontend/src/views/assignments/assignment-details.js
+++ b/frontend/src/views/assignments/assignment-details.js
@@ -1,6 +1,5 @@
 import React from "react";
 import { connect } from "react-redux";
-import { Button } from "react-bootstrap";
 import { offerTableSelector } from "../offertable/actions";
 import {
     assignmentsSelector,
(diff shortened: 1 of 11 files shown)


frontend/src/views/assignments/assignment-details.js
```diff
@@ -89,9 +90,12 @@ function ViewAssignmentDetailsButton(props) {
...
     return (
         <React.Fragment>
-            <Button onClick={() => setVisibleAssignments(selectedAssignments)}>
+            <ActionButton
+                icon={<FaInfoCircle />}
+                onClick={() => setVisibleAssignments(selectedAssignments)}
+            >
                 View Selected Assignment Details
-            </Button>
+            </ActionButton>
```
siefkenj: This whole button is going away/changing. Leave it where it is for now.

frontend/src/views/assignments/offer-actions.js
```diff
@@ -56,14 +64,24 @@ function OfferActionButtons(props) {
...
-                <Button onClick={acceptOffers}>Set as Accepted</Button>
-                <Button onClick={rejectOffers}>Set as Rejected</Button>
-            </ButtonGroup>
+            <ActionButton icon={<FaPlusCircle />} onClick={createOffers}>
+                Create Offer
+            </ActionButton>
+            <ActionButton icon={<FaMinusCircle />} onClick={withdrawOffers}>
+                Withdraw Offer
+            </ActionButton>
+            <ActionButton icon={<FaArrowCircleRight />} onClick={emailOffers}>
```
siefkenj: We can do better with the icons. Search through and try to think of what icons would be meaningful (e.g, here a "mail" icon would be better)

wisemandenny: I really looked. If you would like to suggest one, I would appreciate it, because I scrolled through the react icons/fa icons for about 1hr trying to find a set of icons that we can use for mail. The only one that doesn't really make super intuitive sense is the email one, I think.

wisemandenny:
I'm not sure what's going on with the unit tests, it seems like there's something wrong with the typescript configuration

siefkenj:
I'm not sure either....the tests passed on the last merge and you didn't touch any typescript stuff. Did you rebase from master recently?

frontend/src/views/assignments/offer-actions.js
siefkenj: `<i class="fa fa-envelope"></i>` for mail
`<i class="fa fa-user-times"></i>` or `<i class="fa fa-ban"></i>` for withdraw 
`<i class="fa fa-ban"></i>` for nag
`<i class="fa fa-check-circle"></i>` or `<i class="fa fa-pencil-square-o"></i>` for accept

Something like that

wisemandenny:
Yeah, I rebased off of master before starting this branch.

siefkenj:
Rebase again.

frontend/src/views/assignments/offer-actions.js
wisemandenny: `FaUserPlus` for add offer
`FaUserTimes` for withdraw offer
`FaEnvelope` for Email
`FaUserClock` for nag (am i misunderstanding what this does? it seems like ban doesn't really make sense...)
`FaCheck` for Accept Offer
`FaBan` for Reject Offer

siefkenj: A nag is like a reprimand.

PR #405, opened 2020-07-15
Minor fixes

siefkenj:
Cleanup some styles and avoid a crash when there aren't valid instructors loaded.

### DIFF
--- a/frontend/src/api/actions/positions.js
+++ b/frontend/src/api/actions/positions.js
@@ -200,7 +200,11 @@ export const positionsSelector = createSelector(
                 ...rest
             }) => ({
                 ...rest,
-                instructors: instructor_ids.map((x) => instructorsById[x]),
+                // When the instructor list references an instructor that we haven't loaded
+                // we don't want the frontend to crash, so filter out any null instructors
+                instructors: instructor_ids
+                    .map((x) => instructorsById[x])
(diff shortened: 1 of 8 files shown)


frontend/src/api/actions/positions.js
```diff
@@ -200,7 +200,11 @@ export const positionsSelector = createSelector(
                 ...rest
             }) => ({
                 ...rest,
-                instructors: instructor_ids.map((x) => instructorsById[x]),
+                // When the instructor list references an instructor that we haven't loaded
+                // we don't want the frontend to crash, so filter out any null instructors
+                instructors: instructor_ids
+                    .map((x) => instructorsById[x])
+                    .filter((x) => x),
```
wisemandenny, line 207: Is this really the same as `x !== null`?
Cool...

wisemandenny (approved):
Looks good. `CellButtonDelete` is much improved over what I wrote. Thanks!

frontend/src/api/actions/positions.js
siefkenj: It's not the same. It's the same as `!!x === true`, but `null` and `undefined` are both falsey. If someone has a utorid of `""`, that would be a problem, but that's diallowed :-)

PR #410, opened 2020-07-22
Initial steps for mailer

reidka:
You can create an offer from the assignments table and then click on the "email offer" button.  It send an email to the appropriate TA.  For development, we are using a simply SMTP server called mailcatcher.  Instead of sending the email to the intended user, it catches the email and outputs it to a web page a http://localhost:1080/

Looking for advice on:
- I'm not sure if the "emailed_date" is set in the correct place, because I'm not sure when update is called.
- The url feels too hard-coded, but I'm not sure where to put it.
- Clicking on the accept or reject from the contract view does change the state, but it doesn't automatically update in the assignments table.


backend/app/mailers/offer_mailer.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class OfferMailer < ApplicationMailer
+    # default from comes from config environment
+    layout 'mailer'
+
+    def contract_email(assignment, link)
+        @offer = assignment.active_offer
+        @url = link
+        mail(to: assignment.applicant.email, subject: 'TA Position Offer')
```
siefkenj, line 10: We could add the position code in the subject. That way when they hit reply, we can tell the emails apart...

backend/app/views/offer_mailer/contract_email.text.erb
```diff
@@ -0,0 +1,10 @@
+Dear <%= @offer.first_name %> <%= @offer.last_name %>,
+
+  We would like to offer you a position as a TA for <%= @offer.position_code %> <%= @offer.position_title %>.
+
+
```
siefkenj, line 5: So many newlines?

backend/config/application.rb
```diff
@@ -38,5 +38,8 @@ class Application < Rails::Application
         config.contract_template_dir =
             ENV.fetch('CONTRACT_TEMPLATE_DIR', '/storage/contract_templates')
                 .presence || '/storage/contract_templates'
+        
+        # This was added to use in the emails that send contract links.
+        config.base_url = ENV.fetch('BASE_URL', 'localhost:3000')
```
siefkenj, line 43: `localhost:3000` is only the route during debug...it's a little worrysome seeing it show up here, but then again, normally it should be using `BASE_URL`

siefkenj:
I'll merge this now so we can do testing and fix things up later.

PR #417, opened 2020-07-23
instructors load even if a session isn't selected

### LINKED ISSUE #411: `instructors` not fetched until an active session is set
The `/admin/instructors` route does not depend on there being an active session. Currently, the init procedure doesn't fetch instructors until an active session is set, but they should be fetched right away (just like sessions are).

wisemandenny:
Fixes #411

### DIFF
--- a/frontend/src/api/actions/init.js
+++ b/frontend/src/api/actions/init.js
@@ -80,6 +80,9 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
     return async (dispatch, getState) => {
         const parsedGlobals = { mockAPI: null, activeSession: null };
 
+        // These actions do not need to finish in a specific order, so we can wait for them to finish at the end of this function to speed up startup.
+        const asyncActions = [fetchActiveUser, fetchSessions, fetchInstructors];
+
         /**
          * A helper function to determine if the `currentStage`
          * should be run
@@ -94,8 +97,6 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
                 "toggleMockAPI",
                 "setActiveUser",
                 "setActiveUserRole",
-                "fetchInstructors",
-                "fetchSessions",
                 "setActiveSession",
                 "updateGlobals",
                 "fetchSessionDependentData",
@@ -155,17 +156,9 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
             toggleMockApi(globals.mockAPI);
         }
 
-        if (shouldRunStage("setActiveUser")) {
-            await dispatch(fetchActiveUser());
-        }
-
         if (shouldRunStage("setActiveUserRole")) {
             const activeRole = activeRoleSelector(getState());
             await dispatch(setActiveUserRole(activeRole, { skipInit: true }));
-        }
-
-        if (shouldRunStage("fetchSessions")) {
-            await dispatch(fetchSessions());
         }
 
         if (shouldRunStage("setActiveSession")) {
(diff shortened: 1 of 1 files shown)


frontend/src/api/actions/init.js
```diff
@@ -94,6 +94,7 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
                 "toggleMockAPI",
                 "setActiveUser",
                 "setActiveUserRole",
+                "fetchInstructors",
```
siefkenj: This should come right after `"fetchSessions"`

frontend/src/api/actions/init.js
```diff
@@ -187,6 +188,10 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
             );
         }
 
+        if (shouldRunStage("fetchInstructors")) {
+            await dispatch(fetchInstructors());
```
siefkenj: It doesn't actually matter when this finishes as long as it finishes by the end of the thunk. So, you can save this to a list of "actions that can finish any time" array and then do an `await Promise.all(...)` at the very end of the file. That will make the whole startup process *slightly* faster.

wisemandenny: OK, Cool. I just aped the other code so your suggestion makes more sense.

frontend/src/api/actions/init.js
wisemandenny: No problem, but I'm wondering why?

siefkenj: `sessions` are a more basic object than `instructors`. As some point in the future, there might be a reason to fetch sessions but not instructors. If they are in the other order, that confuses the matter.

wisemandenny:
@siefkenj 
There are several instances in `frontend/src/api/actions/init.js` that have this form:
~~~
if (sessions.find((session) => session.id === activeSession.id)) {
    return true;
}
return false;
~~~
Why not just `return sessions.find((session) => session.id === activeSession.id))` ?

frontend/src/api/actions/init.js
wisemandenny: I added this now. I am sure there is a better variable name for these actions than `asyncActions` but I couldn't think of it...

siefkenj:
There's probably not a good reason for it.

frontend/src/api/actions/init.js
```diff
@@ -80,6 +80,9 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
     return async (dispatch, getState) => {
         const parsedGlobals = { mockAPI: null, activeSession: null };
 
+        // These actions do not need to finish in a specific order, so we can wait for them to finish at the end of this function to speed up startup.
+        const asyncActions = [fetchActiveUser, fetchSessions, fetchInstructors];
```
siefkenj: Incorrect. `fetchAcitveUser` and `fetchSessions` **must** finish before the others. `fetchInstructors` is the only one that is independent ATM

siefkenj: The whole point of an init system is to have things happen in order. Look carefully and where the results of `fetchActiveUser` and `fetchSessions` are used.

frontend/src/api/actions/init.js
```diff
@@ -80,6 +80,9 @@ export function initFromStage(stage, options = { startAfterStage: false }) {
     return async (dispatch, getState) => {
         const parsedGlobals = { mockAPI: null, activeSession: null };
 
+        // These actions don't need to finish in a specific order, so we can wait for them to finish at the end of this function to speed up startup.
+        const asyncActions = [fetchInstructors];
```
siefkenj: Still no. This doesn't save any time.

You need to dispatch the action right away, but only await the action at the end.

wisemandenny: so line 84 should be: `const asyncActions = [dispatch(fetchInstructors())];`
and then at the bottom should be `await Promise.all(asyncActions);`?

siefkenj: Yes

PR #418, opened 2020-07-23
Edit Session Info (all fields)

wisemandenny:
This PR addresses #376 - `Sessions` should be editable.

You can now edit all of the session fields as an admin.

### DIFF
--- a/frontend/src/views/sessions/index.tsx
+++ b/frontend/src/views/sessions/index.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { useSelector } from "react-redux";
 import { ConnectedAddSessionDialog } from "./add-session-dialog";
 import { activeSessionSelector } from "../../api/actions";
-import { SessionsList } from "../../components/sessions";
+import { SessionsList as ConnectedSessionsList } from "../../components/sessions";
 import {
     ActionsList,
     ActionButton,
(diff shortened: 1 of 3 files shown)


frontend/src/views/sessions/index.tsx
```diff
@@ -68,7 +64,7 @@ export function AdminSessionsView() {
                         setAddDialogVisible(false);
                     }}
                 />
-                <ConnectedSessionList />
+                <SessionsList />
```
siefkenj: Rename it `ConnectedSessionList` to be consistent with the rest of the code.

frontend/src/views/sessions/index.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
-import { connect, useSelector } from "react-redux";
+import { useSelector } from "react-redux";
 import { ConnectedAddSessionDialog } from "./add-session-dialog";
-import { sessionsSelector, activeSessionSelector } from "../../api/actions";
-import { SessionsList } from "../../components/sessions";
+import { activeSessionSelector } from "../../api/actions";
+import { SessionsList as ConnectedSessionsList } from "../../components/sessions";
```
siefkenj: No....the convention in this codebase is that every widget connected to the Redux store has a name that starts with `Connected`. Changing it on import confuses the matter.

wisemandenny: I made a big brain fart here. Let me fix that...

wisemandenny:
i broke it again. going to reopen soon...

PR #421, opened 2020-07-25
Enable Offers Tests

siefkenj:
The offers tests that were only running on the mockAPI are now running on the mockAPI and real API.


backend/app/views/offer_mailer/email_nag.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<html>
+  <head>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <style>
+    </style>
+  </head>
+  <body>
+  <p>Dear <%= @first_name %> <%= @last_name %></p>
+  <p>
+    This is your <%= @nag_count %> reminder. Please accept
```
wisemandenny: Does this come out as "This is your 1 reminder" or "This is your 1st reminder"?

wisemandenny (approved):

backend/app/views/offer_mailer/email_nag.html.erb
siefkenj: Good point!

PR #423, opened 2020-07-25
Assignments filter improvements

### LINKED ISSUE #414: Assignments Filter Improvements
When you first create an assignment, it has `null` status, which makes it display a blank cell in the status column.
Admins need to be able to select all such assignments, so that they can bulk create/email contracts.

Further, it should also be possible to filter the assignments table on multiple fields. A suggested implementation was to type in a string in the filter box, press enter, and it should appear as a "chip" next to the filter bar. This can be repeated as necessary to add as many criteria as needed.

Implementation for `null` status filtering should not change the db. We can just display "no contract" in grey/subdued text in the status bar by using the frontend only.

wisemandenny:
This PR addresses issue #414 
- Upgrades the filtering capability of the `FilterableTable` component
  - Users must now press `Enter` when they complete typing a filter string
  - When new filter strings are added, they are displayed as Chips
  - Each chip has the text of the filter string and an `x` button, which when clicked removes the chip/filter item
  - If a user tries to add the same filter string repeatedly, new chips will NOT be added
- Adds a `No Offer` string in the `OfferTable` component, which is added in the front end only (no DB interaction), so that users can search for all assignments with no offer created.
- Moves a string utility function (`capitalize`) to the `libs/utils` file, adding/modifying `import` statements as needed in the rest of the codebase


siefkenj:
Let's filter as we type instead of waiting for enter

frontend/src/components/filterable-table.js
```diff
@@ -49,18 +52,21 @@ function FilterableTable(props) {
...
 
-    const filteredData = filterString
-        ? data.filter((row) =>
-              rowToStr(row).includes(filterString.toLowerCase())
-          )
-        : data;
+    const filteredData =
+        filterStrings.length === 0
+            ? data
+            : filterStrings.reduce((filteredData, filterString) => {
```
siefkenj: This is confusing to me...Use something like `rows.filter(row => filterStrings.every(string => row.match(string)))`

frontend/src/components/filterable-table.js
```diff
@@ -184,15 +190,52 @@ function FilterableTable(props) {
...
+
+    function removeFilterString(filterString) {
+        const updatedFilterStrings = filterStrings.filter(
+            (oldString) => oldString !== filterString
+        );
+        setFilterStrings(updatedFilterStrings);
+    }
+
+    function addFilterString(filterString) {
+        const formattedFilterString = filterString.toLowerCase();
```
siefkenj: It will confuse the user to see the strings become lower case when they are displayed. Instead, lower case them before they are used to filter, but display what they typed.

wisemandenny: OK, good point

wisemandenny:
How about this re: filtering as we type:
When a user types in the filter box, it behaves exactly as before (filtering as we type), and then if/when they press enter, the string gets added to the chips. Since there still needs to be some way to "submit" a filter query, otherwise there won't be any way to filter by multiple critieria (the point of this PR).

frontend/src/components/filterable-table.js
```diff
@@ -184,15 +190,52 @@ function FilterableTable(props) {
...
+        if (!filterStrings.includes(formattedFilterString)) {
+            const newFilterStrings = Array.from(filterStrings); // so we don't mutate the filterStrings object
```
siefkenj: `[...filterStrings]` to create a copy

frontend/src/components/filterable-table.js
```diff
@@ -184,15 +190,52 @@ function FilterableTable(props) {
...
+            newFilterStrings.push(formattedFilterString);
+            setFilterStrings(newFilterStrings);
```
siefkenj: This whole thing can become `setFilterStrings([...newFilterStrings, formattedFilterString])`

frontend/src/components/filterable-table.js
```diff
@@ -184,15 +190,52 @@ function FilterableTable(props) {
...
+                        className="filter-chip"
+                        pill
+                        variant="info"
+                        key={filterString}
+                    >
+                        <FaTimes
+                            className="filter-chip-icon"
+                            onClick={() => removeFilterString(filterString)}
+                        />
+                        {capitalize(filterString)}{" "}
```
siefkenj: No artificial capitalization

frontend/src/components/filterable-table.js
```diff
@@ -184,15 +190,52 @@ function FilterableTable(props) {
...
+                    }}
                 />
+                {filterStrings.map((filterString) => (
+                    <Badge
+                        className="filter-chip"
+                        pill
+                        variant="info"
+                        key={filterString}
+                    >
+                        <FaTimes
```
siefkenj, line 250: Close button on the right side

frontend/src/components/filterable-table.js
siefkenj: Also, quit doing end-of-line comments ;-)

frontend/src/components/components.css
```diff
@@ -106,3 +106,20 @@
     margin-right: 0.8em;
     margin-bottom: 0.3em;
 }
+
+.filter-chip {
+    padding-left: 1em;
+    padding-right: 2em;
+    margin-left: 1em;
```
siefkenj, line 111: This seems like a lot of room around the chip. Does it look okay?

frontend/src/components/filterable-table.js
wisemandenny: yes, it should

frontend/src/components/components.css
wisemandenny: It looks like this:
image
Not crazy about the "info" color. Might change it to primary or light

frontend/src/components/filterable-table.js
wisemandenny: Replaced with :
~~~
 const filteredData = data.filter((row) =>
        filterStrings.every((string) => rowToStr(row).match(string))
    );
~~~
Much better!

siefkenj: let's optimize a little bit and only convert the row to a string once for each item.

frontend/src/components/components.css
wisemandenny: Those padding attributes weren't doing anything. I had them set so high because I kept increasing them to see if anything would change. Since they didn't I removed them. The margin left I kept so that the chips aren't touching each other. I also changed the color to "light" (grey bg/black text).

frontend/src/components/filterable-table.js
```diff
@@ -184,15 +194,62 @@ function FilterableTable(props) {
...
+        const updatedFilterStrings = filterStrings.filter(
+            (oldString) => oldString !== filterString
+        );
+        setFilterStrings(updatedFilterStrings);
+    }
+
+    function addFilterString(filterString) {
+        // string.localeCompare(, , { sensitivity: 'accent'}) == case insensitive string comparison
+        if (
+            filterStrings.every(
```
siefkenj, line 211: You're doing a different check for inclusion (the locale method) on the list than for filtering (lowercase). Use the same check.

siefkenj: Use `!filterStrings.some` It makes more sense.

frontend/src/components/filterable-table.js
```diff
@@ -49,18 +51,26 @@ function FilterableTable(props) {
...
+            filterStrings.every((string) =>
+                rowToStr(row).match(string.toLowerCase())
+            )
+        )
+        .filter(
+            (row) =>
+                currentFilterString
+                    ? rowToStr(row).includes(currentFilterString.toLowerCase())
+                    : true // if there is no typed query, don't filter the typed query
+        );
```
siefkenj, line 73: Don't do two filters. Create a `[...filterStrings, currentFilterString]` object and use that.

frontend/src/components/filterable-table.js
```diff
@@ -49,18 +51,26 @@ function FilterableTable(props) {
         setSelected([..._selected]);
     };
 
-    const [filterString, setFilterString] = React.useState("");
+    const [filterStrings, setFilterStrings] = React.useState([]);
```
siefkenj, line 54: We need some comments explaining what these similarly-named things are.

siefkenj:
I'm going to merge this now and fix the code issues.

PR #424, opened 2020-07-25
ReactTables now display all rows

### LINKED ISSUE #415: ReactTable only displays 20 rows
If you add more than 20 rows in the various tables in the admin views, only the first 20 will be displayed, with no ability to see the rest. 

This same issue appears in the import dialog boxes. If you try to import a spreadsheet with >20 rows, only the first 20 will be displayed in the "import preview"

wisemandenny:
This PR fixes #415 
Now react tables display exactly the number of rows of data without any pagination.

### DIFF
--- a/frontend/src/components/applicants.tsx
+++ b/frontend/src/components/applicants.tsx
@@ -40,11 +40,13 @@ export function ApplicantsList(props: {
     columns?: any[];
 }) {
     const { applicants, columns = DEFAULT_COLUMNS } = props;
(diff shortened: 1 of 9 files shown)


wisemandenny:
I use a lot of `data?.length` typescript things. (The ? operator). Does this do anything in non-tsx files?

siefkenj:
`x?.y` should do the same thing in regular JS if our version of babel is new enough.

PR #477, opened 2021-01-21
Instructor permission test

Accelsnow:
Completing the following TODOs in `frontend/src/tests/instructor-permission-test.js`

- can't update applications
- can't update applicants
- can't update contract template
- can't update position
- can't update session
- can't update instructors except for self (i.e. active user)


Accelsnow:
The current permission test does not use the correct permission module. Need to refactor.

Accelsnow:
There are 2 problems unresolved:

1. It appears that the mock API does not have some of the permission modules implemented. Some tests fails on mock API but succeeded on backend API.

```
PASS  src/tests/import-export.test.js
 FAIL  src/tests/api.test.js (10.422s)
  ● Mock API tests › Instructor permissions tests › can't update position

    expected API response to have {"status": "error"} but received {"message": "", "payload": {"ad_close_date": null, "ad_hours_per_assignment": null, "ad_num_assignments": null, "ad_open_date": null, "contract_template_id": 1, "current_enrollment": null, "current_waitlisted": null, "desired_num_assignments": null, "duties": null, "end_date": "2020-12-31T00:00:00.000Z", "hours_per_assignment": 70, "id": 2, "instructor_ids": [1, 2], "instructor_utorids": ["smithh", "garciae"], "position_code": "MAT136H1F", "position_title": "test title", "qualifications": null, "start_date": "2020-02-10T00:00:00.000Z"}, "status": "success"}

      69 |                 updatedApplicant
      70 |             );
    > 71 |             expect(respInvalidRouteWithSession).toHaveStatus("error");
         |                                                 ^
      72 |             checkPropTypes(errorPropTypes, respInvalidRouteWithSession);
      73 |         }
      74 | 

      at cantUpdateTestFrame (src/tests/instructor-permission-test.js:71:49)
      at Object.<anonymous> (src/tests/instructor-permission-test.js:209:9)

  ● Mock API tests › Instructor permissions tests › can't update contract template

    expected API response to have {"status": "error"} but received {"message": "", "payload": {"id": 1, "template_file": "default-template.html", "template_name": "test template"}, "status": "success"}

      69 |                 updatedApplicant
      70 |             );
    > 71 |             expect(respInvalidRouteWithSession).toHaveStatus("error");
         |                                                 ^
      72 |             checkPropTypes(errorPropTypes, respInvalidRouteWithSession);
      73 |         }
      74 | 

      at cantUpdateTestFrame (src/tests/instructor-permission-test.js:71:49)
      at Object.<anonymous> (src/tests/instructor-permission-test.js:228:9)

  ● Mock API tests › Instructor permissions tests › can't update applications

    expected API response to have {"status": "error"} but received {"message": "", "payload": {"applicant_id": 2, "comment": "<Autogenerated>", "id": 1, "program": "test program", "session_id": 1}, "status": "success"}

      78 |                 updatedApplicant
      79 |             );
    > 80 |             expect(respInvalidRouteWithoutSession).toHaveStatus("error");
         |                                                    ^
      81 |             checkPropTypes(errorPropTypes, respInvalidRouteWithoutSession);
      82 |         }
      83 | 

      at cantUpdateTestFrame (src/tests/instructor-permission-test.js:80:52)
      at Object.<anonymous> (src/tests/instructor-permission-test.js:266:9)

Test Suites: 1 failed, 1 passed, 2 total
Tests:       3 failed, 25 todo, 177 passed, 205 total
Snapshots:   0 total
Time:        10.622s
Ran all test suites.
```

2. Tests for instructors modifying themselves have not yet been written. It is unsure how to modify an existing instructor. `/admin/instructors` does not allow existing utorid to be passed as an argument, which means it can not update existing instructors (despite having the corresponding id). '/instructor/instructors' gives a 404 regardless of the body.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -16,6 +16,86 @@ export function instructorsPermissionTests(api) {
...
+        await action();
+
+        let respSwitchBackUser = await apiPOST(
+            `/debug/active_user`,
+            originalUser
+        );
+        expect(respSwitchBackUser).toHaveStatus("success");
+    };
+
+    const cantUpdateTestFrame = async (
```
siefkenj: Prefer `function cantUpdateTestFrame` to `const cantUpdateTestFrame = () => {}`.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -16,6 +16,86 @@ export function instructorsPermissionTests(api) {
...
+
+        if (withoutSessionModRoute !== null) {
+            let respInvalidRouteWithoutSession = await apiPOST(
+                withoutSessionModRoute,
+                updatedApplicant
+            );
+            expect(respInvalidRouteWithoutSession).toHaveStatus("error");
+            checkPropTypes(errorPropTypes, respInvalidRouteWithoutSession);
+        }
+
```
frankcbw: I noticed that the above two if-statements are pretty much identical except with different variable names.  Should we refactor the chunk into a separate helper function?

Accelsnow:
The backend code for route `POST instructor/instructor` is not yet implemented. The corresponding test code is therefore commented. They should be restored once the backend is implemented.

siefkenj:
As soon as PR #492 is merged, the POST route for `/instructor/instructors` should be ready to test :-)

Accelsnow:
I have merged the master branch PRs and got the tests to work.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +124,80 @@ export function instructorsPermissionTests(api) {
...
+            if (self.id !== inst.id) {
+                otherInst = inst.id;
+            }
+        });
+
+        if (otherInst === null) {
+            throw Error("There is no other instructor other than self!");
+        }
+
+        let modifiedOtherInst = { ...otherInst };
```
frankcbw: I am a bit confused here. Isn't `otherInst` just `inst.id`, which is an ID value?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -16,6 +16,77 @@ export function instructorsPermissionTests(api) {
     const { apiGET, apiPOST } = api;
     let session = null;
 
+    async function cantUpdateTestFrame(
```
siefkenj: I have no idea what this function does based on its name. All utility functions should have doc strings explaining what they do. Look in `utils.js` for examples (install `Document This` extension for VSCode if you want help generating doc strings).

This function takes too many arguments. If your function takes more than 3 arguments, it is usually a sign that you need to rethink how the code works.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -16,6 +16,77 @@ export function instructorsPermissionTests(api) {
...
+        modFunc
+    ) {
+        let respFetch = await apiGET(fetchRoute);
+        expect(respFetch).toHaveStatus("success");
+        const prevExistingApplicant = respFetch.payload[0];
+        let updatedApplicant = { ...prevExistingApplicant };
+        modFunc(updatedApplicant);
+
+        // instructor should not have route to modify directly
+        if (withSessionModRoute !== null) {
```
siefkenj: Do you really mean `!== null` here? Are you sure it shouldn't be `!= null`?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -94,7 +244,17 @@ export function instructorsPermissionTests(api) {
...
+    it("can't update applications", async () => {
+        await cantUpdateTestFrame(
+            `/instructor/sessions/${session.id}/applications`,
+            null,
+            `/instructor/applications`,
+            `/admin/applications`,
+            (application) => {
+                application.program = "test program";
+            }
+        );
```
siefkenj: I don't like how most of the code was moved out of the tests. I cannot look at this code an understand what steps the test is actually testing.

Take an approach more like #487 with setup and teardown functions.

frontend/src/tests/instructor-permission-test.js
Accelsnow: You are right. This is a mistake. otherInst should have been assigned to the instructor user object.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,11 +15,57 @@ export function instructorsPermissionTests(api) {
...
+        expect(respSwitchBackUser).toHaveStatus("success");
+    }
 
     beforeAll(async () => {
         await databaseSeeder.seed(api);
         await databaseSeeder.seedForInstructors(api);
         session = databaseSeeder.seededData.session;
+        instructorOnlyUser = {
+            utorid: "test",
```
siefkenj: This utorid could easily be used in other tests and that would cause confusion..make it something less predictable.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,11 +15,57 @@ export function instructorsPermissionTests(api) {
...
+            roles: ["instructor"],
+        };
+
+        let respAddInstOnlyUser = await apiPOST(
```
siefkenj: Either you reuse the response or you define it with `const`. Also, if an instructor user is added but is not an instructor in any courses, the tests will mostly return empty results...

A user automatically gets instructor permissions if they are assigned as an instructor for a course. <- if that functionality isn't tested yet, it should be.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
+    it("can't update instructors except for self (i.e. active user)", async () => {
+        // we use the default user's admin privilege here to fetch an unrelated instructor
+        let respFetchInst = await apiGET(`/instructor/instructors`);
+        expect(respFetchInst).toHaveStatus("success");
+
+        let respFetchDefaultUser = await apiGET(`/instructor/active_user`);
+        expect(respFetchDefaultUser).toHaveStatus("success");
+        const defaultUser = respFetchDefaultUser.payload;
+
+        const newDefaultUser = { ...defaultUser };
```
siefkenj: Huh? I am really confused by this name. Why are you calling it `DefaultUser`? Why would you be changing a default user in a test?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
     });
 
-    it.todo("can't update session");
+    it("can't update session", async () => {
+        await switchToInstructorOnlyUser();
+
+        const newSession = {
+            start_date: "2020-01-01",
+            end_date: "2020-01-02",
+            name: "test",
```
siefkenj: Bad name. Also, a session that lasts only one day? Make realistic sample data if you are able.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
+
+        resp = await apiPOST("/admin/sessions", newSession);
+        expect(resp).toHaveStatus("error");
+        checkPropTypes(errorPropTypes, resp);
+
+        await restoreDefaultUser();
+    });
 
     it("fetch positions", async () => {
         let resp = await apiGET(`/instructor/sessions/${session.id}/positions`);
```
siefkenj, line 269: What user are we when these tests run?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
+            contract_template_id: 190,
+            qualifications: null,
+            duties: "Tutorials",
+            ad_hours_per_assignment: null,
+            ad_num_assignments: null,
+            ad_open_date: null,
+            ad_close_date: null,
+            desired_num_assignments: null,
+            current_enrollment: null,
+            current_waitlisted: null,
```
siefkenj: You can leave off all `null` fields.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
 
-    it.todo("can't update position");
+    it("can't update position", async () => {
+        await switchToInstructorOnlyUser();
+
+        const newPosition = {
+            position_code: "test code",
+            position_title: "Test Position",
+            start_date: "2020-01-01",
+            end_date: "2020-01-02",
```
siefkenj: You don't need start and end dates. They are automagically inferred.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
+            qualifications: null,
+            duties: "Tutorials",
+            ad_hours_per_assignment: null,
+            ad_num_assignments: null,
+            ad_open_date: null,
+            ad_close_date: null,
+            desired_num_assignments: null,
+            current_enrollment: null,
+            current_waitlisted: null,
+            instructor_ids: [89],
```
siefkenj: What???? You *cannot* hard code any IDs. I'm really surprised this passed tests when you ran it locally. (When you run it locally, the IDs change on every run). Though, I guess since you were hoping the test failed....This is really quite concerning, because the test might fail for a completely different reason than you were hoping for.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
+        };
+
+        let resp = await apiPOST(
+            `/instructor/sessions/${session.id}/positions`,
+            newPosition
+        );
+        expect(resp).toHaveStatus("error");
+        checkPropTypes(errorPropTypes, resp);
```
siefkenj: You don't need to do this. `.toHaveStatus("error")` already checks that it is a proper object.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,21 +99,114 @@ export function instructorsPermissionTests(api) {
...
+
+        resp = await apiPOST(`/admin/positions`, newPosition);
+        expect(resp).toHaveStatus("error");
+        checkPropTypes(errorPropTypes, resp);
+
+        resp = await apiPOST(`/instructor/positions`, newPosition);
```
siefkenj, line 304: Reusing `resp`. Good :-)

frontend/src/tests/instructor-permission-test.js
```diff
@@ -85,7 +247,42 @@ export function instructorsPermissionTests(api) {
...
-    it.todo("can't update applicants");
+    it("can't update applicants", async () => {
+        await switchToInstructorOnlyUser();
+
+        const newApplicant = {
+            first_name: "Test",
+            last_name: "Test",
+            email: "test@test.com",
+            phone: "1111111111",
+            utorid: "test",
```
siefkenj: So...you had an instructor with `utorid == "test"` and now you have an applicant with `utorid == "test"`. Do you see why this might be a problem?

Solution: make better sample data.

Accelsnow:
The current commit has the newly added `assigning to be instructor of a course grants instructor role` test failing. I am unsure if this functionality is not yet implemented, or am I using the wrong route for doing so.

siefkenj:
Tests failing. Still more work to do...

Accelsnow:
> Tests failing. Still more work to do...



The failed test was the one you mentioned in the previous code review. I first created a user without any role, then created a position with that user's id in the instructor id (assigning user to be the instructor of a course). However, the user does not seem to gain the instructor role afterwards.

siefkenj:
> > 
> 
> The failed test was the one you mentioned in the previous code review. I first created a user without any role, then created a position with that user's id in the instructor id (assigning user to be the instructor of a course). However, the user does not seem to gain the instructor role afterwards.

Sounds like some investigation is in order!

Accelsnow:
> > > 
> > 
> > 
> > The failed test was the one you mentioned in the previous code review. I first created a user without any role, then created a position with that user's id in the instructor id (assigning user to be the instructor of a course). However, the user does not seem to gain the instructor role afterwards.
> 
> Sounds like some investigation is in order!

I did some investigation in the Ruby code. In `position_controller`, the param permission was set so that the instructor_ids must be passed in as a list; However, later in the `perform` call in `position_service.rb`, it seems like Ruby is querying for instructor using this exact list as the instructor id, which does not make sense and fails to find. I made sure that the user id is the id of the newly created user by checking both `/admin/users` and `/debug/users`. 
The detailed log on tapp_backend:
```

Started POST "/api/v1/admin/sessions/23/positions" for 172.22.0.3 at 2021-02-11 15:57:26 +0000
  User Load (0.1ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 27], ["LIMIT", 1]]
  ↳ app/services/active_user_service.rb:36:in `active_user'
Processing by Api::V1::Admin::SessionPositionsController#create as HTML
  Parameters: {"position_code"=>"course_assign_grant_instructor_test_position_code", "position_title"=>"course assign grant instructor test position title", "hours_per_assignment"=>1, "contract_template_id"=>22, "duties"=>"Tutorials", "instructor_ids"=>[29], "session_id"=>"23"}
  Session Load (0.1ms)  SELECT "sessions".* FROM "sessions" WHERE "sessions"."id" = $1 LIMIT $2  [["id", 23], ["LIMIT", 1]]
  ↳ app/controllers/api/v1/admin/session_positions_controller.rb:24:in `find_session'
  Position Load (0.1ms)  SELECT "positions".* FROM "positions" WHERE "positions"."session_id" = $1 AND "positions"."id" IS NULL LIMIT $2  [["session_id", 23], ["LIMIT", 1]]
  ↳ app/controllers/api/v1/admin/session_positions_controller.rb:20:in `find_position'
  CACHE Session Load (0.0ms)  SELECT "sessions".* FROM "sessions" WHERE "sessions"."id" = $1 LIMIT $2  [["id", 23], ["LIMIT", 1]]
  ↳ app/controllers/api/v1/admin/session_positions_controller.rb:24:in `find_session'
   (0.1ms)  BEGIN
  ↳ app/services/position_service.rb:18:in `perform'
  Session Load (0.1ms)  SELECT "sessions".* FROM "sessions" WHERE "sessions"."id" = $1 LIMIT $2  [["id", 23], ["LIMIT", 1]]
  ↳ app/services/position_service.rb:18:in `perform'
  ContractTemplate Load (0.1ms)  SELECT "contract_templates".* FROM "contract_templates" WHERE "contract_templates"."id" = $1 LIMIT $2  [["id", 22], ["LIMIT", 1]]
  ↳ app/services/position_service.rb:18:in `perform'
  Position Exists? (0.1ms)  SELECT 1 AS one FROM "positions" WHERE "positions"."position_code" = $1 AND "positions"."session_id" = $2 LIMIT $3  [["position_code", "course_assign_grant_instructor_test_position_code"], ["session_id", 23], ["LIMIT", 1]]
  ↳ app/services/position_service.rb:18:in `perform'
  Position Create (0.2ms)  INSERT INTO "positions" ("session_id", "position_code", "position_title", "hours_per_assignment", "created_at", "updated_at", "contract_template_id") VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING "id"  [["session_id", 23], ["position_code", "course_assign_grant_instructor_test_position_code"], ["position_title", "course assign grant instructor test position title"], ["hours_per_assignment", 1.0], ["created_at", "2021-02-11 15:57:26.925222"], ["updated_at", "2021-02-11 15:57:26.925222"], ["contract_template_id", 22]]
  ↳ app/services/position_service.rb:18:in `perform'
  Instructor Load (0.1ms)  SELECT "instructors".* FROM "instructors" WHERE "instructors"."id" = $1  [["id", 29]]
  ↳ app/services/position_service.rb:19:in `perform'
ERROR: Couldn't find Instructor with 'id'=[29]
Couldn't find Instructor with 'id'=[29]
Traceback:
	/services/position_service.rb:19:in `perform'
	/controllers/api/v1/admin/positions_controller.rb:34:in `block in upsert'
	/controllers/concerns/transaction_handler.rb:8:in `block in start_transaction_and_rollback_on_exception'
	/controllers/concerns/transaction_handler.rb:7:in `start_transaction_and_rollback_on_exception'
	/controllers/api/v1/admin/positions_controller.rb:31:in `upsert'
	/controllers/api/v1/admin/session_positions_controller.rb:13:in `create'
[active_model_serializers] Rendered ActiveModel::Serializer::Null with Hash (0.04ms)
   (0.1ms)  COMMIT
  ↳ app/controllers/concerns/transaction_handler.rb:7:in `start_transaction_and_rollback_on_exception'
Completed 200 OK in 8ms (Views: 0.2ms | ActiveRecord: 0.9ms | Allocations: 10415)
```

I am unfamiliar with how Ruby makes these queries exactly, so I may need some help to determine the cause of this issue.

siefkenj:
You got a conflict with master. You need to merge `upstream/master` and resolve the conflict.

Accelsnow:
Instructor id is not the same as user id. Use utorid to fetch instructor id.

Accelsnow:
In `instructor-permission-test.js`, the logic for applying instructor role to a user is in a conditional clause that checks if `Rails.application.config.allow_basic_auth` is defined. It is commented as "If we're configured to use Apache's Basic Auth, use that. The alternative is Shibboleth (not yet implemented)". The above-mentioned condition happens to never be true when I run `npm test` on the frontend, and thus the instructor-role-applying code is never executed.

backend/app/services/active_user_service.rb
```diff
@@ -17,37 +17,46 @@ def self.active_user(request)
...
+              )
+            end
         end
 
         # rubocop:disable Style/RaiseArgs
         raise NotImplementedError.new 'active_user Route is not implemented yet for production mode'
         # rubocop:enable Style/RaiseArgs
     end
+
+    def self.grant_roles
```
siefkenj: I'm surprised Rubocop let this through, but I dont think you should have a `self` here. And maybe you should call it `granted_roles` or `computed_roles`

backend/app/services/active_user_service.rb
```diff
@@ -17,37 +17,46 @@ def self.active_user(request)
...
+            else
+                # If we're here, the database doesn't have the specified active user,
+                # so createa fake one with all the permissions. We *do not* save this
+                # user to the database.
+                return(
+                  User.new(
                     utorid: 'defaultactive', roles: %w[admin instructor ta]
                 )
-            )
+              )
```
siefkenj: Some weird spacing is going on here...did you run Prettier with the Ruby plugin installed?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
     it("fetch applicants", async () => {
+        await switchToInstructorOnlyUser();
         let resp = await apiGET(
             `/instructor/sessions/${session.id}/applicants`
         );
         expect(resp).toHaveStatus("success");
+
+        await restoreDefaultUser();
+        resp = await apiGET(`/instructor/sessions/${session.id}/applicants`);
+        expect(resp).toHaveStatus("success");
```
siefkenj: Are you going to test that the correct applicants show up?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+        };
+
+        let resp = await apiPOST(
+            `/instructor/sessions/${session.id}/contract_templates`,
+            newTemplate
+        );
+        expect(resp).toHaveStatus("error");
+
+        resp = await apiPOST(
+            `/admin/sessions/${session.id}/contract_templates`,
```
siefkenj, line 387: Put some comment that you're testing both `/instructor` and `/admin` prefix.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+    });
 
     it("fetch contract templates", async () => {
+        await switchToInstructorOnlyUser();
         let resp = await apiGET(
             `/instructor/sessions/${session.id}/contract_templates`
         );
         expect(resp).toHaveStatus("success");
+
+        await restoreDefaultUser();
```
siefkenj, line 368: Why? You already tested fetching as the instructor....

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+            hours_per_assignment: 1.0,
+            contract_template_id: existingContractTemplateId,
+            duties: "Tutorials",
+        };
+
+        let resp = await apiPOST(
+            `/instructor/sessions/${session.id}/positions`,
+            newPosition
+        );
+        expect(resp).toHaveStatus("error");
```
siefkenj, line 293: You tested that you cannot create a position. But did you ever test that you cannot update one?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+    });
 
     it("fetch positions", async () => {
+        await switchToInstructorOnlyUser();
         let resp = await apiGET(`/instructor/sessions/${session.id}/positions`);
         expect(resp).toHaveStatus("success");
+
+        await restoreDefaultUser();
+        resp = await apiGET(`/instructor/sessions/${session.id}/positions`);
+        expect(resp).toHaveStatus("success");
```
siefkenj: Are you going to check that the payload contains the positions that are supposed to be there?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+        await switchToInstructorOnlyUser();
+
+        const newSession = {
+            start_date: "2020-04-10T01:00:00.000Z",
+            end_date: "2020-11-31T00:01:20.000Z",
+            name: "instructor_permission_cant_update_session_test_session",
+        };
+
+        let resp = await apiPOST("/instructor/sessions", newSession);
+        expect(resp).toHaveStatus("error");
```
siefkenj, line 245: This is creating, not updating.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+        );
+    });
 
     it("fetch sessions", async () => {
-        let resp = await apiGET("/instructor/instructors");
+        await switchToInstructorOnlyUser();
+        let resp = await apiGET("/instructor/sessions");
+        expect(resp).toHaveStatus("success");
+
+        await restoreDefaultUser();
```
siefkenj, line 231: Why are you testing as default user?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+        expect(respModDefaultUser.payload.email).toEqual(
+            "instructor_permission_instructor_update_test@test.com"
+        );
+    });
 
     it("fetch sessions", async () => {
-        let resp = await apiGET("/instructor/instructors");
+        await switchToInstructorOnlyUser();
+        let resp = await apiGET("/instructor/sessions");
+        expect(resp).toHaveStatus("success");
```
siefkenj, line 229: What about the payload?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -53,54 +206,238 @@ export function instructorsPermissionTests(api) {
...
+        let respFetchDefaultUser = await apiGET(`/instructor/active_user`);
+        expect(respFetchDefaultUser).toHaveStatus("success");
+        const defaultUser = respFetchDefaultUser.payload;
+
+        const defaultUserWithUpdatedInformation = { ...defaultUser };
+        defaultUserWithUpdatedInformation.email =
+            "instructor_permission_instructor_update_test@test.com";
+
+        // instructor can not modify other instructors
+        // in this example, the instructor only user cant update default user's information
```
siefkenj: Only can't? This is confusing. Please rephrase this comment.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+        resp = await apiPOST(`/debug/active_user`, emptyRoleUserData);
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET(`/debug/active_user`);
+        expect(resp).toHaveStatus("success");
+        expect(resp.payload.utorid).toEqual(emptyRoleUserData.utorid);
+        expect(resp.payload.roles).toEqual(
+            expect.arrayContaining(["instructor"])
+        );
+
+        // This requires further confirmation. The following logic for checking roles may be wrong.
```
siefkenj: Why is this comment here?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+        };
+
+        let resp = await apiPOST(`/debug/users`, emptyRoleUserData);
+        expect(resp).toHaveStatus("success");
+
+        resp = await apiGET(`/debug/users`);
+        expect(resp).toHaveStatus("success");
+
+        let emptyRoleUserExists = false;
+        resp.payload.forEach((user) => {
```
siefkenj: `resp.payload.find((user) => user.utorid === emptyRoleUserData.utorid)` will be null if nothing is found, otherwise it will be the user object. Alternatively, you could use `.some` to get a boolean.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+            }
+        });
+
+        expect(emptyRoleUserExists).toBeTruthy();
+
+        const instructorData = {
+            utorid: emptyRoleUserUtorid,
+            first_name: "course_assign_grant_instructor_test_first_name",
+            last_name: "course_assign_grant_instructor_test_last_name",
+            email: "course_assign_grant_instructor_test_email",
```
siefkenj: Only the UTORid needs to be unique. You can make the other fields human names.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+        };
+        resp = await apiPOST(`/admin/instructors`, instructorData);
+        expect(resp).toHaveStatus("success");
+
+        resp = await apiGET(`/admin/instructors`);
+        expect(resp).toHaveStatus("success");
+        let instructorId;
+        resp.payload.forEach((instructor) => {
```
siefkenj: You can use `.find` for this.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+            if (instructor.utorid === emptyRoleUserUtorid) {
+                instructorId = instructor.id;
+            }
+        });
+        expect(instructorId).toBeDefined();
+
+        const defaultTestPosition = {
```
siefkenj: Why is `default` part of the name?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+
+        // upon successfully making the following position upsert, the no-role user should be granted instructor role
+        resp = await apiPOST(
+            `/admin/sessions/${session.id}/positions`,
+            defaultTestPosition
+        );
+        expect(resp).toHaveStatus("success");
+
+        // set active user to trigger role update
+        resp = await apiPOST(`/debug/active_user`, emptyRoleUserData);
```
siefkenj: Didn't you already post this data? Why is a second posting required...? Smells like a mistake or a bug.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -15,14 +15,167 @@ export function instructorsPermissionTests(api) {
...
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET(`/debug/active_user`);
```
siefkenj: There's no reason for all this to be taking place for the active user. In fact, that's kind of weird...the active user should be an `admin` the whole time, and a new user should be created and retrieved.

backend/app/services/active_user_service.rb
Accelsnow: Since the above `active_user` function was defined with `self` (`def self.active_user(request)`), therefore `granted_roles` needs to have `self` too. Otherwise it results in a NameError.

frontend/src/tests/instructor-permission-test.js
Accelsnow: I would argue checking the content of the response should be checked inside the `instructor-tests.js`, while `intructor-permission-tests.js` only checks the accessibility of different routes given a certain set of permissions.

frontend/src/tests/instructor-permission-test.js
Accelsnow: For the first test, the user has only the instructor role assigned; for the second test, it is a user with mixed roles (admin, ta and instructor). The intuition here is to test with a different combination of permissions and assure that as long as the instructor is one of the roles, this route would be accessible. (it may be redundent)

frontend/src/tests/instructor-permission-test.js
Accelsnow: the default user also has the instructor role, and it also has some other roles. The purpose for testing as both the instructor-only-user and the default user is to make sure users with either only the instructor role or many roles are able to access instructor routes.

frontend/src/tests/instructor-permission-test.js
Accelsnow: I would argue checking the content of the response should be checked inside the instructor-tests.js, while intructor-permission-tests.js only checks the accessibility of different routes given a certain set of permissions.

frontend/src/tests/instructor-permission-test.js
Accelsnow: I would argue checking the content of the response should be checked inside the instructor-tests.js, while intructor-permission-tests.js only checks the accessibility of different routes given a certain set of permissions.

frontend/src/tests/instructor-permission-test.js
Accelsnow: posting to active_user sets the current active user, which is what triggers the role-granting logic (since the granted_roles function is defined in the active_user service).

backend/app/services/active_user_service.rb
Accelsnow: This is auto-corrected by rubocop. The previous indentation did not pass ruby linting tests defined in tests/run-ruby-liniting-tests.sh.

```
snowsr@ADR-LINUX:~/projects/tapp-accel/tests$ ./run-ruby-linting-tests.sh
This script assumes that all docker images are up and running.
If they are not, please run `docker-compose up` and rerun this script.
Running Unit Tests...
WARNING: The RAILS_SERVE_STATIC_FILES variable is not set. Defaulting to a blank string.
WARNING: The RAILS_LOG_TO_STDOUT variable is not set. Defaulting to a blank string.
WARNING: The ALLOW_BASIC_AUTH variable is not set. Defaulting to a blank string.
WARNING: The TAPP_ADMINS variable is not set. Defaulting to a blank string.
Creating tapp_backend_run ... done
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.

Please also note that can also opt-in to new cops by default by adding this to your config:
  AllCops:
    NewCops: enable

Gemspec/DateAssignment: # (new in 1.10)
  Enabled: true
Style/HashConversion: # (new in 1.10)
  Enabled: true
For more information: https://docs.rubocop.org/rubocop/versioning.html
Inspecting 121 files
.......................................................................C.................................................

Offenses:

app/services/active_user_service.rb:40:13: C: [Correctable] Layout/ClosingParenthesisIndentation: Indent ) to column 14 (not 12)
            )
            ^

121 files inspected, 1 offense detected, 1 offense auto-correctable
```

Accelsnow:
Permission - only see data you should see
Multi-role - do it only once

Accelsnow:
backend/app/serializer/user_serializer.rb
implement user role assignment logic here

Accelsnow:
The enhanced permission check (check correct data visibility) will be done in another PR after this PR is merged.

frontend/src/tests/instructor-permission-test.js
siefkenj: With the updates, we shouldn't need this any mor.

frontend/src/tests/instructor-permission-test.js
siefkenj: Ditto

backend/app/models/user.rb
```diff
@@ -15,6 +15,23 @@ class User < ApplicationRecord
...
+                               roles + %w[ta]
+                           end
+
+        calculated_roles += %w[instructor] if Instructor.find_by(utorid: utorid)
+        # Check if they are listed as an instructor for any course
+        if Rails.application.config.always_admin.include?(utorid)
+            calculated_roles += %w[admin]
+        end
+
+        calculated_roles
```
siefkenj: We may return a list that contains duplicate entries. There's complicated logic to make sure `ta` is added only once, but not the other things. Let's just add the roles and then uniquify the list before we return it.

PR #478, opened 2021-01-21
[frontend] implement tests for updating offer when emailed

### LINKED ISSUE #476: [frontend] tests for update of offer when emailed again
implement tests for update of an offer's emailed_date attribute whenever emailed again

Altair59:
Fix #476:
implement tests for:
update the emailed_date field of the assignment when the assignment is re-sent

### DIFF
--- a/frontend/src/tests/offer-tests.js
+++ b/frontend/src/tests/offer-tests.js
@@ -78,7 +78,7 @@ export function offersTests(api) {
         let previousDate = newOffer.emailed_date;
 
         // sleep for 1s to create time diff
-        await new Promise(r => setTimeout(r, 1000));
+        await new Promise((r) => setTimeout(r, 2000));
         const resp = await apiPOST(
             `/admin/assignments/${assignment.id}/active_offer/email`
         );
--- a/frontend/src/tests/offer-tests.js
+++ b/frontend/src/tests/offer-tests.js
@@ -74,7 +74,7 @@ export function offersTests(api) {
         expect(newOffer.emailed_date).not.toBeNull();
     });
 
-    it("email the active offer again", async () => {
+    it("when an offer is emailed a second time, the emailed date is updated to the most recent emailed date", async () => {
         // record previous emailed_date
         const resp = await apiGET(
             `/admin/assignments/${assignment.id}/active_offer`
--- a/frontend/src/tests/offer-tests.js
+++ b/frontend/src/tests/offer-tests.js
@@ -75,26 +75,28 @@ export function offersTests(api) {
     });
 
     it("email the active offer again", async () => {
-        let previousDate = newOffer.emailed_date;
-
-        // sleep for 1s to create time diff
-        await new Promise((r) => setTimeout(r, 2000));
-        const resp = await apiPOST(
-            `/admin/assignments/${assignment.id}/active_offer/email`
+        // record previoud emailed_date
+        const resp = await apiGET(
+            `/admin/assignments/${assignment.id}/active_offer`
         );
         expect(resp).toHaveStatus("success");
+        let previousDate = resp.payload.emailed_date;
 
+        // send the email again
+        const resp1 = await apiPOST(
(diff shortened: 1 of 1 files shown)


frontend/src/tests/offer-tests.js
```diff
@@ -74,9 +74,30 @@ export function offersTests(api) {
         expect(newOffer.emailed_date).not.toBeNull();
     });
 
-    it.todo(
-        "when an offer is emailed a second time, the emailed date is updated to the most recent emailed date"
-    );
+    it("email the active offer again", async () => {
+        // record previoud emailed_date
```
siefkenj: You should install a spell checker that spell-checks comments to catch things like `previoud`.

frontend/src/tests/offer-tests.js
```diff
@@ -74,9 +74,30 @@ export function offersTests(api) {
         expect(newOffer.emailed_date).not.toBeNull();
     });
 
-    it.todo(
-        "when an offer is emailed a second time, the emailed date is updated to the most recent emailed date"
-    );
+    it("email the active offer again", async () => {
```
siefkenj: Change the description back to an appropriate one

Accelsnow:
The PR message can be a little bit clearer about which(or both) of the following case it is testing:

A. it is creating a timestamp for a newly created email

or 

B. it is updating the timestamp for an email that was re-sent

frontend/src/tests/offer-tests.js
```diff
@@ -74,9 +74,30 @@ export function offersTests(api) {
...
-    it.todo(
-        "when an offer is emailed a second time, the emailed date is updated to the most recent emailed date"
-    );
+    it("when an offer is emailed a second time, the emailed date is updated to the most recent emailed date", async () => {
+        // record previous emailed_date
+        const resp = await apiGET(
+            `/admin/assignments/${assignment.id}/active_offer`
+        );
+        expect(resp).toHaveStatus("success");
+        let previousDate = resp.payload.emailed_date;
```
osamaramihafez: Can we get an active offer from the database seeder rather than using a fetch to get the offer? It's important if we can because using the database seeder would allow us to avoid any issues that might arise from a failed fetch.

frontend/src/tests/offer-tests.js
```diff
@@ -74,9 +74,30 @@ export function offersTests(api) {
...
+
+        // send the email again
+        const resp1 = await apiPOST(
+            `/admin/assignments/${assignment.id}/active_offer/email`
+        );
+        expect(resp1).toHaveStatus("success");
+        // check if date is updated
+        newOffer = resp1.payload;
+        expect(newOffer.emailed_date).not.toEqual(previousDate);
+
```
frankcbw, line 95: Preferably to make things more consistent by 
1) saving 'resp.payload' into a variable named 'previousOffer'
or 
2) saving 'resp1.payload.emailed_date' as a new variable named 'newDate'

frontend/src/tests/offer-tests.js
Accelsnow: previousDate is a constant. can use `const` here instead of `let`

frontend/src/tests/offer-tests.js
```diff
@@ -74,9 +74,30 @@ export function offersTests(api) {
...
+        let previousDate = resp.payload.emailed_date;
+
+        // send the email again
+        const resp1 = await apiPOST(
+            `/admin/assignments/${assignment.id}/active_offer/email`
+        );
+        expect(resp1).toHaveStatus("success");
+        // check if date is updated
+        newOffer = resp1.payload;
+        expect(newOffer.emailed_date).not.toEqual(previousDate);
```
Altair59: Use stronger check than not toEqual. (Check if the new date is actually later than prev date)

frontend/src/tests/offer-tests.js
Altair59: thank you! will fix this! :)

siefkenj:
Great job!

PR #480, opened 2021-01-22
Implement applicant test. 

frankcbw:
Implemented tests 
- "delete an applicant with no associated assignments"
- "fail to delete an applicant with an associated assignment"

P.S.
The second test case is failing under Mock API,  thus removed from the Mock API for now

### DIFF
--- a/frontend/src/tests/applicant-tests.js
+++ b/frontend/src/tests/applicant-tests.js
@@ -129,9 +129,9 @@ export function applicantTests(api) {
     it("delete an applicant with no associated assignments", async () => {
         // upsert and delete an applicant without a session
         const newApplicant1 = {
-            first_name: "Tommy3",
-            last_name: "Smith3",
-            utorid: "smith3",
+            first_name: "June",
+            last_name: "Kim",
+            utorid: "kim",
         };
         const resp = await apiPOST(`/admin/applicants`, newApplicant1);
         expect(resp).toHaveStatus("success");
@@ -148,30 +148,25 @@ export function applicantTests(api) {
         expect(resp3).toHaveStatus("success");
         expect(resp3.payload.map((x) => x.id)).not.toContain(newApplicant1.id);
 
-        // upsert and delete an applicant with a session
-        const newApplicant2 = {
-            first_name: "June",
-            last_name: "Kim",
-            utorid: "kim",
-        };
+        // upsert and delete an applicant via session route
         const resp4 = await apiPOST(
             `/admin/sessions/${session.id}/applicants`,
-            newApplicant2
+            newApplicant1
         );
         expect(resp4).toHaveStatus("success");
         checkPropTypes(applicantPropTypes, resp4.payload);
-        Object.assign(newApplicant2, resp4.payload);
+        Object.assign(newApplicant1, resp4.payload);
 
         // try deleting the applicant
         const resp5 = await apiPOST(`/admin/applicants/delete`, {
-            id: newApplicant2.id,
+            id: newApplicant1.id,
         });
         expect(resp5).toHaveStatus("success");
 
         // verify the applicant is deleted
         const resp6 = await apiGET(`/admin/sessions/${session.id}/applicants`);
         expect(resp6).toHaveStatus("success");
-        expect(resp6.payload.map((x) => x.id)).not.toContain(newApplicant2.id);
+        expect(resp6.payload.map((x) => x.id)).not.toContain(newApplicant1.id);
     });
 
     it("fail to delete an applicant with an associated assignment", async () => {
--- a/frontend/src/tests/applicant-tests.js
+++ b/frontend/src/tests/applicant-tests.js
@@ -12,11 +12,13 @@ export function applicantTests(api) {
     let session;
     let applicant;
     let newApplicant;
+    let contractTemplate;
 
     beforeAll(async () => {
         await databaseSeeder.seed(api);
         session = databaseSeeder.seededData.session;
         applicant = databaseSeeder.seededData.applicant;
+        contractTemplate = databaseSeeder.seededData.contractTemplate;
     }, 30000);
 
     it("applicants associated with a session", async () => {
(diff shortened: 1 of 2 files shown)


frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +128,92 @@ export function applicantTests(api) {
...
+    it("fail to delete an applicant with an associated assignment", async () => {
+        const newApplicant1 = {
+            first_name: "Tommy4",
+            last_name: "Smith4",
+            utorid: "smith4",
+        };
+        const resp = await apiPOST(`/admin/applicants`, newApplicant1);
+        expect(resp).toHaveStatus("success");
+        checkPropTypes(applicantPropTypes, resp.payload);
+        Object.assign(newApplicant1, resp.payload);
```
osamaramihafez: I may be mistaken, but I believe that you can have multiple assignments to a position for each applicant. So, we could use the database seeder instead of retesting the applicant insert.

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +128,92 @@ export function applicantTests(api) {
...
+
+        const newAssignment = {
+            note: "",
+            position_id: position.id,
+            applicant_id: newApplicant1.id,
+            start_date: "2019-09-09T00:00:00.000Z",
+            end_date: "2019-12-31T00:00:00.000Z",
+        };
+        const resp2 = await apiPOST("/admin/assignments", newAssignment);
+        expect(resp2).toHaveStatus("success");
```
osamaramihafez: I don't think we need to verify if the assignment was successfully added, this would be done in a different test case (specific to assignments).

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +128,92 @@ export function applicantTests(api) {
...
 
-    it.todo("fail to delete an applicant with an associated assignment");
+        const resp3 = await apiPOST(`/admin/applicants/delete`, {
+            id: newApplicant1.id,
+        });
+        expect(resp3).toHaveStatus("error");
+
+        const resp4 = await apiGET(`/admin/applicants`);
+        expect(resp4.payload).toContainObject(newApplicant1);
```
osamaramihafez: Just out of curiosity, if an error is raised, will the backend keep the inserted data? It shouldn't, to my understanding, but I'm not sure. If the case is that it doesn't insert data if there is an error, then we can say that this isn't a necessary check.

frankcbw: The error is raised when it tries to delete the applicant. Therefore, I think it requires to verify the data is intact.

frontend/src/tests/applicant-tests.js
frankcbw: I think you are right. The reason I create a new applicant, in this case, is that the test case requires an applicant without any associated assignment. It may be harder to distinguish which applicants meet that requirement in pre-seeded data.

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +128,101 @@ export function applicantTests(api) {
...
+            hours_per_assignment: 70,
+            start_date: "2019/09/09",
+            end_date: "2019/12/31",
+            contract_template_id: contractTemplate.id,
+        };
+
+        const { payload: position } = await apiPOST(
+            `/admin/sessions/${session.id}/positions`,
+            newPosition
+        );
```
Altair59: In order to ensure an assignment is successfully posted, I believe a check here to see if the position is successfully posted is needed.

frontend/src/tests/applicant-tests.js
Altair59: I believe this check is necessary. For the purpose of this test, we need to ensure that we are deleting an applicant with a valid assignment.

frontend/src/tests/applicant-tests.js
frankcbw: Isn't that tested by position-tests.js?

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +128,101 @@ export function applicantTests(api) {
         expect(resp.payload).toContainObject(newApplicant);
     });
 
-    it.todo("delete an applicant with no associated assignments");
+    it("delete an applicant with no associated assignments", async () => {
+        // upsert and delete an applicant without a session
+        const newApplicant1 = {
+            first_name: "Tommy3",
+            last_name: "Smith3",
```
siefkenj: Prefer actual names rather than putting numbers in names.

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +128,101 @@ export function applicantTests(api) {
...
+            utorid: "smith3",
+        };
+        const resp = await apiPOST(`/admin/applicants`, newApplicant1);
+        expect(resp).toHaveStatus("success");
+        checkPropTypes(applicantPropTypes, resp.payload);
```
siefkenj: proptypes should already be checked in the upsert tests. No need to test again.

frontend/src/tests/applicant-tests.js
siefkenj: Look at the seed data. There should be an applicant with an associated assignment already. That should make this tests one or two lines long in total.

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +126,67 @@ export function applicantTests(api) {
...
+            first_name: "June",
+            last_name: "Kim",
+            utorid: "kim",
+        };
+        const resp = await apiPOST(`/admin/applicants`, newApplicant1);
+        expect(resp).toHaveStatus("success");
+        Object.assign(newApplicant1, resp.payload);
+
+        // try deleting the applicant
+        const resp2 = await apiPOST(`/admin/applicants/delete`, {
```
siefkenj: Unless you actually need it, do one `let resp` at the beginning and then reuse `resp` throughout.

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +126,67 @@ export function applicantTests(api) {
...
+            id: newApplicant1.id,
+        });
+        expect(resp2).toHaveStatus("success");
 
-    it.todo("fail to delete an applicant with an associated assignment");
+        // verify the applicant is deleted
+        const resp3 = await apiGET(`/admin/applicants`);
+        expect(resp3).toHaveStatus("success");
+        expect(resp3.payload.map((x) => x.id)).not.toContain(newApplicant1.id);
```
siefkenj: I'd feel safer if you double checked that `newApplicant1.id != null` before you did this check.
Since this would be an error in the test setup, you can just `throw new Error(...)` if it's null.

frontend/src/tests/applicant-tests.js
```diff
@@ -126,7 +126,67 @@ export function applicantTests(api) {
...
+        expect(resp6.payload.map((x) => x.id)).not.toContain(newApplicant1.id);
+    });
+
+    it("fail to delete an applicant with an associated assignment", async () => {
+        // Retrieve all assignments from the current session
+        const resp0 = await apiGET(`/admin/sessions/${session.id}/assignments`);
+        expect(resp0).toHaveStatus("success");
+
+        // pick an associated applicant id
+        const idToDelete = resp0.payload.filter(
```
siefkenj: `.find` is what you want.

frankcbw: Thanks Jason!

frontend/src/tests/applicant-tests.js
frankcbw: I wonder if checking `newApplicant1.id != null` is really necessary here because if it verifies that the response status is "success" above it, right?

siefkenj:
Read this page https://eslint.org/docs/rules/require-atomic-updates and see if you can get your code to pass the linter.

frontend/src/tests/applicant-tests.js
```diff
@@ -12,6 +12,7 @@ export function applicantTests(api) {
     let session;
     let applicant;
     let newApplicant;
+    let resp;
```
siefkenj: No global `resp` variable. It should be defined inside each function.

frankcbw: Got it

siefkenj:
Great job! Don't forget to fetch upstream and merge upstream/master before your next PR.

PR #481, opened 2021-01-22
initial implementation for assignment insertion with null dates

osamaramihafez:


frontend/src/tests/assignment-tests.js
```diff
@@ -111,24 +113,36 @@ export function assignmentsTests(api) {
...
+            hours_per_assignment: 70,
+            start_date: "2019-09-05T00:00:00.000Z",
+            end_date: "2020-01-05T00:00:00.000Z",
+            contract_template_id: contractTemplate.id,
+        };
+
+        var resp = await apiPOST(
+            `/admin/sessions/${session.id}/positions`,
+            newPositionData
         );
```
Altair59: I believe a check here to see if the position is successfully posted in needed, in order to ensure the assignment goes to a valid position.

frontend/src/api/mockAPI/assignments.js
```diff
@@ -98,6 +98,19 @@ export class Assignment extends MockAPIController {
                 active_offer_url_token: activeOffer.url_token,
             });
         }
+        // Add position start and end dates if they are not defined
+        const position = this.getPosition(matchingAssignment)
+        if (!matchingAssignment.start_date) {
+            Object.assign(ret, {
+                start_date: position.start_date
+            });
```
siefkenj: No need to use `Object.assign` here. Just `ret.start_date = ...`

osamaramihafez: I was just following the same format as from lines 89 to 100, which seems to use Object.assign rather than individually assigning the attributes. Should we always use this format moving forwards? I've changed it to ret.start_date =  ...

frontend/src/tests/assignment-tests.js
```diff
@@ -111,24 +111,37 @@ export function assignmentsTests(api) {
...
+            position_title: "Basic Computer Science",
+            hours_per_assignment: 70,
+            start_date: "2019-09-05T00:00:00.000Z",
+            end_date: "2020-01-05T00:00:00.000Z",
+            contract_template_id: contractTemplate.id,
+        };
 
-        expect(withUpdatedAssignment.map((x) => x.id)).toContain(
-            updatedAssignment.id
+        var resp = await apiPOST(
```
siefkenj: Use `let` over `var`

frontend/src/api/mockAPI/assignments.js
siefkenj: Yeah. Those lines are weird too and probably should be changed...

siefkenj:
Have you figured out Prettier yet? You're still failing code formatting.

siefkenj:
Great job!

PR #485, opened 2021-01-26
Instructor view fixes

siefkenj:
Adds Ddah routes for the `/api/v1/instructor` prefix, as well as make the react tables read-only for fields an instructor is not allowed to edit.

### DIFF
--- a/backend/app/controllers/api/v1/instructor/ddahs_controller.rb
+++ b/backend/app/controllers/api/v1/instructor/ddahs_controller.rb
@@ -1,0 +1,105 @@
+# frozen_string_literal: true
+
+class Api::V1::Instructor::DdahsController < ApplicationController
+    before_action :validate_instructor
+
+    # GET /sessions/:session_id/ddahs
+    def index
+        render_success Ddah
+                           .by_session(params[:session_id])
+                           .by_instructor(@active_instructor.id)
+    end
+
+    # GET /ddahs/:ddah_id
+    def show
+        find_ddah
+        unless @ddah
+            render_error(
+                message:
+                    "Instructor '#{
+                        @active_instructor.utorid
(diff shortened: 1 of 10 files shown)


backend/app/controllers/api/v1/instructor/ddahs_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+        @active_user = ActiveUserService.active_user request
+
+        # If we're here, we have `instructor` permissions, but that doesn't mean
+        # we actually *are* an instructor.
+        @active_instructor = Instructor.find_by(utorid: @active_user.utorid)
+        render_error(message: 'Not an instructor') unless @active_instructor
+    end
+
+    def find_or_create
+        @ddah = nil
```
Accelsnow, line 60: This is more of a question rather than a review.
Is it a convention to use this instance variable style to manage responses like finding an instructor/ddah? Why are we not using return values instead? I just think it is a little bit confusing since these variables are not even defined upon class initialization, which makes it a bit hard to track, so I assume it is a Ruby thing.

siefkenj: Using instance variables is a form of caching, so we don't need to redo database queries.

backend/config/routes.rb
```diff
@@ -40,166 +40,176 @@ def matches?(request)
...
+                        # Application
+                        resources :applications, only: :create
+
+                        # Assignments
+                        resources :assignments, only: %i[show create] do
+                            resources :wage_chunks,
+                                      controller: :assignment_wage_chunks,
+                                      only: %i[index create]
+                            resources :offers,
+                                      path: :active_offer,
```
frankcbw, line 85: More of a question, that why a `path:` is specified for `:offers` but not for `:wage_chunks`?

PR #487, opened 2021-01-26
Implement user permission tests

Accelsnow:
Working on `/frontend/tests/user-permissioin-tests.js`.

TODO-tests to be implemented:

1. An instructor can access instructor routes
2. A TA can access TA routes
3. An instructor who is also a TA can access instructor and TA routes

### DIFF
--- a/frontend/src/tests/user-permission-tests.js
+++ b/frontend/src/tests/user-permission-tests.js
@@ -1,4 +1,4 @@
-import { expect, it, beforeAll, afterEach } from "./utils";
+import { expect, it, beforeAll } from "./utils";
 import { databaseSeeder } from "./setup";
 
 /**
--- a/frontend/src/tests/user-permission-tests.js
+++ b/frontend/src/tests/user-permission-tests.js
@@ -14,10 +14,28 @@ import { databaseSeeder } from "./setup";
 export function userPermissionsTests(api) {
     const { apiGET, apiPOST } = api;
 
-    const newUserData = {
+    const newInstructorData = {
         utorid: "userxx",
         roles: ["instructor"],
     };
+
+    async function runTestWithUser(userData, testBody) {
+        let resp = await apiPOST("/admin/users", userData);
+        // Save any extra attributes we got, such as an `id`
+        Object.assign(userData, resp.payload);
+
+        // Set the active user to the newly-created user
+        resp = await apiPOST("/debug/active_user", { id: userData.id });
+        expect(resp).toHaveStatus("success");
+
+        await testBody();
+
+        // Set the active_user back to the default.
+        resp = await apiPOST("/debug/active_user", {
+            id: databaseSeeder.seededData.active_user.id,
+        });
+        expect(resp).toHaveStatus("success");
+    }
 
     beforeAll(async () => {
         await databaseSeeder.seed(api);
@@ -30,26 +48,22 @@ export function userPermissionsTests(api) {
     });
 
     it("A non-admin cannot access the admin route", async () => {
-        let resp = await apiPOST("/admin/users", newUserData);
-        // Save any extra attributes we got, such as an `id`
-        Object.assign(newUserData, resp.payload);
-
-        // Set the active user to the newly-created user
-        resp = await apiPOST("/debug/active_user", { id: newUserData.id });
-        expect(resp).toHaveStatus("success");
-
-        // Try to fetch an admin route
-        resp = await apiGET("/admin/active_user");
-        expect(resp).toHaveStatus("error");
-
-        // Set the active_user back to the default.
-        resp = await apiPOST("/debug/active_user", {
-            id: databaseSeeder.seededData.active_user.id,
+        await runTestWithUser(newInstructorData, async () => {
+            // Try to fetch an admin route
+            let resp = await apiGET("/admin/active_user");
+            expect(resp).toHaveStatus("error");
         });
-        expect(resp).toHaveStatus("success");
     });
 
-    it.todo("An instructor can access instructor routes");
+    it("An instructor can access instructor routes", async () => {
+        // test instructor route with a user with only the instructor role
+        await runTestWithUser(newInstructorData, async () => {
+            // Try to fetch an instructor route
+            let resp = await apiGET("/instructor/instructors");
+            expect(resp).toHaveStatus("success");
+        });
+    });
+
     it.todo("A TA can access TA routes");
     it.todo(
         "An instructor who is also a TA can access instructor and TA routes"
--- a/frontend/src/tests/user-permission-tests.js
+++ b/frontend/src/tests/user-permission-tests.js
@@ -30,7 +30,7 @@ export function userPermissionsTests(api) {
     };
 
     async function setupUserPerm(userData) {
-        let respPrevUser = await apiGET("/active_user");
+        const respPrevUser = await apiGET("/active_user");
         expect(respPrevUser).toHaveStatus("success");
         const prevUserId = respPrevUser.payload.id;
 
--- a/frontend/src/tests/user-permission-tests.js
+++ b/frontend/src/tests/user-permission-tests.js
@@ -4,7 +4,7 @@ import { databaseSeeder } from "./setup";
 /**
  * Tests for the API. These are encapsulated in a function so that
  * different `apiGET` and `apiPOST` functions can be passed in. For example,
- * they may be functions that make actual requests via http or they may be from 
+ * they may be functions that make actual requests via http or they may be from
  * the mock API.
  * @param {object} api
  * @param {Function} api.apiGET A function that when passed a route will return the get response
--- a/frontend/src/tests/user-permission-tests.js
+++ b/frontend/src/tests/user-permission-tests.js
@@ -36,6 +36,7 @@ export function userPermissionsTests(api) {
 
         // Set the active user to the newly-created user
         resp = await apiPOST("/debug/active_user", { id: newUserData.id });
+        expect(resp).toHaveStatus("success");
 
         // Try to fetch an admin route
         resp = await apiGET("/admin/active_user");
@@ -45,6 +46,7 @@ export function userPermissionsTests(api) {
         resp = await apiPOST("/debug/active_user", {
             id: databaseSeeder.seededData.active_user.id,
         });
+        expect(resp).toHaveStatus("success");
     });
 
     it.todo("An instructor can access instructor routes");
--- a/frontend/src/tests/user-permission-tests.js
+++ b/frontend/src/tests/user-permission-tests.js
@@ -16,33 +16,30 @@ export function userPermissionsTests(api) {
 
     let defaultUser;
 
-    const newInstructorOnlyUserData = {
-        utorid: "userxx",
+    const instructorOnlyUserData = {
+        utorid: "user_permission_test_instructor_only_utorid",
         roles: ["instructor"],
     };
 
-    const newTAOnlyUserData = {
-        utorid: "tauser",
+    const TAOnlyUserData = {
+        utorid: "user_permission_test_ta_only_utorid",
         roles: ["ta"],
     };
 
-    const newTAInstructorUserData = {
-        utorid: "tainstuser",
+    const TAInstructorUserData = {
+        utorid: "user_permission_test_ta_and_instructor_utorid",
         roles: ["ta", "instructor"],
     };
 
     /**
      * Switch current active user to the user represented by the given user data. If given user does not exist, one will
      * be created.
-     * @param userData the user to be set to the active user
+     * @param newActiveUser the user to be set to the active user
      * @returns {Promise<void>}
      */
-    async function switchToTargetUser(userData) {
-        let resp = await apiPOST("/debug/users", userData);
-        expect(resp).toHaveStatus("success");
-
-        // Set the active user to the newly-created user
-        resp = await apiPOST("/debug/active_user", { id: resp.payload.id });
+    async function switchToTargetUser(newActiveUser) {
+        // Set the active user to the target user
+        let resp = await apiPOST("/debug/active_user", newActiveUser);
         expect(resp).toHaveStatus("success");
     }
 
@@ -61,17 +58,22 @@ export function userPermissionsTests(api) {
 
     beforeAll(async () => {
         await databaseSeeder.seed(api);
-        const respOriginalUser = await apiGET(`/debug/active_user`);
-        expect(respOriginalUser).toHaveStatus("success");
+        let resp = await apiGET(`/debug/active_user`);
+        expect(resp).toHaveStatus("success");
         defaultUser = {
-            utorid: respOriginalUser.payload.utorid,
-            roles: respOriginalUser.payload.roles,
+            utorid: resp.payload.utorid,
+            roles: resp.payload.roles,
         };
-    }, 30000);
 
-    afterEach(async () => {
-        await restoreDefaultUser();
-    }, 3000);
+        resp = await apiPOST("/debug/users", instructorOnlyUserData);
+        expect(resp).toHaveStatus("success");
+
+        resp = await apiPOST("/debug/users", TAOnlyUserData);
+        expect(resp).toHaveStatus("success");
+
+        resp = await apiPOST("/debug/users", TAInstructorUserData);
+        expect(resp).toHaveStatus("success");
+    }, 30000);
 
     it("Admin user can access admin route", async () => {
         // the default user has all roles, including the admin role
@@ -80,9 +82,9 @@ export function userPermissionsTests(api) {
     });
 
     it("A non-admin cannot access the admin route", async () => {
-        await switchToTargetUser(newInstructorOnlyUserData);
+        await switchToTargetUser(instructorOnlyUserData);
 
-        // Try to fetch an admin route
+        // Try to fetch admin routes
         let resp = await apiGET("/admin/active_user");
         expect(resp).toHaveStatus("error");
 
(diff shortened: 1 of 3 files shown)


Accelsnow:
`instructor/instructors` route gets `204 No Content` response from the ruby backend.

```
Started GET "/api/v1/instructor/instructors" for 172.21.0.6 at 2021-01-26 20:03:20 +0000
  User Load (0.1ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 316], ["LIMIT", 1]]
  ↳ app/services/active_user_service.rb:36:in `active_user'
Processing by Api::V1::Instructor::InstructorsController#index as HTML
  CACHE User Load (0.0ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 LIMIT $2  [["id", 316], ["LIMIT", 1]]
  ↳ app/services/active_user_service.rb:36:in `active_user'
  Instructor Load (0.1ms)  SELECT "instructors".* FROM "instructors" WHERE "instructors"."utorid" = $1 LIMIT $2  [["utorid", "smithh"], ["LIMIT", 1]]
  ↳ app/controllers/api/v1/instructor/instructors_controller.rb:11:in `index'
Completed 204 No Content in 1ms (ActiveRecord: 0.2ms | Allocations: 1083)
```

This causes the frontend test prop type check to fail, since the type check expects an object as the HTTP response body, while 204 response body is an empty string.

```
● Console

    console.error node_modules/prop-types/checkPropTypes.js:20
      Warning: Failed PropType verification type: Invalid PropType verification `<root>` of type `string` supplied to `the client`, expected `object`.
    console.warn src/tests/utils.js:284
      Proptypes error call stack: Error: 
          at checkPropTypes (/app/src/tests/utils.js:282:19)
          at apiGET (/app/src/tests/utils.js:94:5)
          at runMicrotasks (<anonymous>)
          at processTicksAndRejections (internal/process/task_queues.js:97:5)
          at Object.<anonymous> (/app/src/tests/user-permission-tests.js:72:20)
```

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,11 +14,41 @@ import { databaseSeeder } from "./setup";
...
+        utorid: "tauser",
+        roles: ["ta"],
+    };
+
+    const newTAInstructorData = {
+        utorid: "tainstuser",
+        roles: ["ta", "instructor"],
+    };
+
+    async function runTestWithUser(userData, testBody) {
```
Accelsnow: Need to parse this function into a setup-teardown style instead of passing a function into a function

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,11 +14,43 @@ import { databaseSeeder } from "./setup";
...
+        roles: ["ta"],
+    };
+
+    const newTAInstructorData = {
+        utorid: "tainstuser",
+        roles: ["ta", "instructor"],
+    };
+
+    async function setupUserPerm(userData) {
+        let respPrevUser = await apiGET("/active_user");
```
Altair59: Maybe I am wrong, but I believe since respPrevUser is not changed we can use const here, as what is prevUserId

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,11 +14,43 @@ import { databaseSeeder } from "./setup";
 export function userPermissionsTests(api) {
     const { apiGET, apiPOST } = api;
 
-    const newUserData = {
+    const newInstructorData = {
```
siefkenj: This is user data, not instructor data...

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,11 +14,43 @@ import { databaseSeeder } from "./setup";
...
         utorid: "userxx",
         roles: ["instructor"],
     };
 
+    const newTAData = {
```
siefkenj: How about, `userTA` and `userInstructor`, etc. for variable names?

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,11 +14,43 @@ import { databaseSeeder } from "./setup";
...
+        utorid: "tauser",
+        roles: ["ta"],
+    };
+
+    const newTAInstructorData = {
+        utorid: "tainstuser",
+        roles: ["ta", "instructor"],
+    };
+
+    async function setupUserPerm(userData) {
```
siefkenj: Don't use abbreviations unless they're more common than the full word. E.g., `Perm`->`Permissions`

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,11 +14,43 @@ import { databaseSeeder } from "./setup";
...
+        const respPrevUser = await apiGET("/active_user");
+        expect(respPrevUser).toHaveStatus("success");
+        const prevUserId = respPrevUser.payload.id;
+
+        let resp = await apiPOST("/admin/users", userData);
```
siefkenj: A mix of `let` and `const`? Make a choice and stick with it.

frontend/src/tests/user-permission-tests.js
siefkenj: The name of this function doesn't clue me in on what's happening. Try again.

siefkenj: Function is missing a doc string. All utility functions should have doc strings. (Look in `utils.js` for examples). You can install the `Document This` extension for VSCode to make creating doc strings easier.

frontend/src/tests/user-permission-tests.js
```diff
@@ -30,26 +62,54 @@ export function userPermissionsTests(api) {
     });
 
     it("A non-admin cannot access the admin route", async () => {
-        let resp = await apiPOST("/admin/users", newUserData);
-        // Save any extra attributes we got, such as an `id`
-        Object.assign(newUserData, resp.payload);
-
-        // Set the active user to the newly-created user
-        resp = await apiPOST("/debug/active_user", { id: newUserData.id });
+        const prevUserId = await setupUserPerm(newInstructorData);
```
siefkenj: Based on this pattern, it might be nicer to have a class that is initialized with a user e.g. named `ActiveUserSwitcher`, and then you can call `await ActiveUserSwitcher.makeActive()` and `await ActiveUserSwitcher.revert()` or something like that.

siefkenj: If you create a class for this, it should also have its own test (maybe an existing `active_user` can do double duty?)

frontend/src/tests/user-permission-tests.js
```diff
@@ -30,26 +62,54 @@ export function userPermissionsTests(api) {
...
 
         // Try to fetch an admin route
-        resp = await apiGET("/admin/active_user");
+        let resp = await apiGET("/admin/active_user");
```
siefkenj, line 85: You test a single route...let's test many.

frontend/src/tests/user-permission-tests.js
```diff
@@ -30,26 +62,54 @@ export function userPermissionsTests(api) {
...
-        resp = await apiPOST("/debug/active_user", {
-            id: databaseSeeder.seededData.active_user.id,
-        });
+        await restoreOriginalUserPerm(prevUserId);
+    });
+
+    it("An instructor can access instructor routes", async () => {
+        const prevUserId = await setupUserPerm(newInstructorData);
+
+        // the default user has all roles, including the instructor role
```
siefkenj, line 110: Pick a user with only the instructor role.

siefkenj: Also test for a user with multiple roles.

frontend/src/tests/user-permission-tests.js
```diff
@@ -30,26 +62,54 @@ export function userPermissionsTests(api) {
...
-        "An instructor who is also a TA can access instructor and TA routes"
-    );
+    it("A TA can access TA routes", async () => {
+        // the default user has all roles, including the TA role
+        let resp = await apiGET("/ta/active_user");
+        expect(resp).toHaveStatus("success");
+
+        const prevUserId = await setupUserPerm(newTAData);
+
+        // Try to fetch an instructor route
```
siefkenj: instructor -> ta

frontend/src/tests/user-permission-tests.js
```diff
@@ -30,26 +62,54 @@ export function userPermissionsTests(api) {
...
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET("/instructor/instructors");
+        expect(resp).toHaveStatus("success");
+
+        const prevUserId = await setupUserPerm(newTAInstructorData);
+
+        // Try to fetch an instructor route
+        resp = await apiGET("/ta/active_user");
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET("/instructor/instructors");
```
siefkenj: Why are you using `/instructor/instructors` instead of `/instructor/active_user`?

Accelsnow: can you explain the question a bit more? I used `/instructor/instructors` here just because it is an instructor route.

Accelsnow: > Why are you using `/instructor/instructors` instead of `/instructor/active_user`?

can you explain the question a bit more? I used /instructor/instructors here just because it is an instructor route.

frontend/src/tests/user-permission-tests.js
Accelsnow: > instructor -> ta

i do not understand this review. would you please explain this more?

frontend/src/tests/user-permission-tests.js
siefkenj: `/instructor/active_user` is also an instructor route. You used `/ta/active_user`, so `/instructor/active_user` seems like the natural choice. You should have a reason if you deviate from the natural choice.

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,127 @@ import { databaseSeeder } from "./setup";
...
+    };
+
+    /**
+     * Switch current active user to the user represented by the given user data. If given user does not exist, one will
+     * be created.
+     * @param userData the user to be set to the active user
+     * @returns {Promise<void>}
+     */
+    async function switchToTargetUser(userData) {
+        let resp = await apiPOST("/debug/users", userData);
```
siefkenj: You create a new user each time before you switch to it? That doesn't seem right...

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,127 @@ import { databaseSeeder } from "./setup";
...
+        let respSwitchBackUser = await apiPOST(
+            `/debug/active_user`,
+            defaultUser
+        );
+        expect(respSwitchBackUser).toHaveStatus("success");
+    }
+
     beforeAll(async () => {
         await databaseSeeder.seed(api);
+        const respOriginalUser = await apiGET(`/debug/active_user`);
```
siefkenj: Just use `resp`

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,127 @@ import { databaseSeeder } from "./setup";
...
     }, 30000);
 
+    afterEach(async () => {
+        await restoreDefaultUser();
+    }, 3000);
+
     it("Admin user can access admin route", async () => {
         // the default user has all roles, including the admin role
-        const resp = await apiGET("/admin/active_user");
+        let resp = await apiGET("/admin/active_user");
```
siefkenj, line 77: Why did you change this?

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,127 @@ import { databaseSeeder } from "./setup";
 export function userPermissionsTests(api) {
     const { apiGET, apiPOST } = api;
 
-    const newUserData = {
+    let defaultUser;
+
+    const newInstructorOnlyUserData = {
```
siefkenj: Pick a better name. Why is the word `new` in it?

frontend/src/tests/user-permission-tests.js
siefkenj: The comment says the word "instructor" but you mean to have the word "ta"

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,127 @@ import { databaseSeeder } from "./setup";
...
+        const respOriginalUser = await apiGET(`/debug/active_user`);
+        expect(respOriginalUser).toHaveStatus("success");
+        defaultUser = {
+            utorid: respOriginalUser.payload.utorid,
+            roles: respOriginalUser.payload.roles,
+        };
     }, 30000);
 
+    afterEach(async () => {
+        await restoreDefaultUser();
```
siefkenj: It seems weird to have an `afterEach` with no `beforeEach`. It makes it hard to follow exactly what's going on in each test. Just repeat the one line each time you need it.

siefkenj:
Failing code formatting tests....

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,142 @@ import { databaseSeeder } from "./setup";
...
-    const newUserData = {
-        utorid: "userxx",
+    let defaultUser;
+
+    const instructorOnlyUserData = {
+        utorid: "user_permission_test_instructor_only_utorid",
         roles: ["instructor"],
     };
 
+    const TAOnlyUserData = {
```
siefkenj: lower case `ta`. There's no need for the word `Data` to be in this variable name.

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,142 @@ import { databaseSeeder } from "./setup";
...
+        roles: ["ta", "instructor"],
+    };
+
+    /**
+     * Switch current active user to the user represented by the given user data. If given user does not exist, one will
+     * be created.
+     * @param newActiveUser the user to be set to the active user
+     * @returns {Promise<void>}
+     */
+    async function switchToTargetUser(newActiveUser) {
```
siefkenj: No need for word `Target`

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,142 @@ import { databaseSeeder } from "./setup";
...
+        expect(resp).toHaveStatus("success");
+    }
+
+    /**
+     * Restores the active user to the default user (the user logged during test setup in beforeAll).
+     *
+     * @returns {Promise<void>}
+     */
+    async function restoreDefaultUser() {
+        let respSwitchBackUser = await apiPOST(
```
siefkenj: If you're not doing something meaningful with the response objects, just name them `resp`.

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,142 @@ import { databaseSeeder } from "./setup";
...
     beforeAll(async () => {
         await databaseSeeder.seed(api);
+        let resp = await apiGET(`/debug/active_user`);
+        expect(resp).toHaveStatus("success");
+        defaultUser = {
+            utorid: resp.payload.utorid,
+            roles: resp.payload.roles,
+        };
+
+        resp = await apiPOST("/debug/users", instructorOnlyUserData);
```
siefkenj: Comment saying why all these calls are being made.

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,142 @@ import { databaseSeeder } from "./setup";
...
+        expect(resp).toHaveStatus("error");
+
+        resp = await apiGET("/admin/applicants");
+        expect(resp).toHaveStatus("error");
+
+        await restoreDefaultUser();
+    });
+
+    it("An instructor can access instructor routes", async () => {
+        // the default user has all roles, including the instructor role
```
siefkenj: Are we sure we're currently the default user? I would feel safer with a `restoreDefaultUser()` before this.

frontend/src/tests/user-permission-tests.js
```diff
@@ -14,42 +14,142 @@ import { databaseSeeder } from "./setup";
...
-    it.todo("An instructor can access instructor routes");
-    it.todo("A TA can access TA routes");
-    it.todo(
-        "An instructor who is also a TA can access instructor and TA routes"
-    );
+    it("An instructor who is also a TA can access instructor and TA routes", async () => {
+        // the default user has all roles, including both the instructor role and the TA role
+        let resp = await apiGET("/ta/active_user");
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET("/instructor/instructors");
```
siefkenj: Again, `/instructor/instructors` is a weird route to be checking when you have `/instructor/active_user` that you could check instead.

frontend/src/tests/user-permission-tests.js
```diff
@@ -4,52 +4,148 @@ import { databaseSeeder } from "./setup";
...
     beforeAll(async () => {
         await databaseSeeder.seed(api);
+        let resp = await apiGET(`/debug/active_user`);
+        expect(resp).toHaveStatus("success");
+        defaultUser = {
+            utorid: resp.payload.utorid,
+            roles: resp.payload.roles,
+        };
+
+        // create in the database the user with only instructor role
```
siefkenj: It's clear what the code is doing, so your comment doesn't need to address that. What you comment should address is *why* you're doing it.

frontend/src/tests/user-permission-tests.js
```diff
@@ -4,52 +4,148 @@ import { databaseSeeder } from "./setup";
...
-        resp = await apiPOST("/debug/active_user", {
-            id: databaseSeeder.seededData.active_user.id,
-        });
+        await restoreDefaultUser();
+    });
+
+    it("An instructor can access instructor routes", async () => {
+        // switch to instructor only user
+        await switchToUser(instructorOnlyUser);
+        let resp = await apiGET("/instructor/instructors");
```
siefkenj: Switch these to `/instructor/active_user`

siefkenj:
Looks great!

Issue #394, opened 2020-07-02
[Test] import/export functions


siefkenj:
We need tests for the import/export functions that test importing actual spreadsheets and json files.

wisemandenny:
start with isolated functions, e.g. diff, read, import, export

Altair59:
1. import/export tests need to test once `prepareData` and round trip once
2. test each type of Object's `prepareSpreadsheet` function
3. test each type of Object's `prepareMinimal` function
4. tests on correct Object data and some with strange characters

Altair59:
create 4 PRS:
1. hoist all prepareData functions
2. improve object data and object file
3. import test
4. export test

PR #488, opened 2021-01-26
[OUTDATED] implement export/import to/from file tests

### LINKED ISSUE #486: [frontend] Complete TODO Unit Tests
- **user-permission-tests**
- [x] An instructor can access instructor routes
- [x] A TA can access TA routes
- [x] An instructor who is also a TA can access instructor and TA routes

- **template-tests**
- [x] upload a template

- **reporting-tag-tests**
- [ ] get all reporting_tags associated to positions for session
- [ ] get all reporting_tags associated to wage_chunks for session
- [ ] get reporting_tags for position
- [ ] get reporting_tags for wage_chunk
- [ ] create and delete reporting_tags for position
- [ ] create and delete reporting_tags for wage_chunk

- **import-export-tests**
- [x] Export data to a JSON/CSV/XLSX
- [x] Import data from a JSON/CSV/XLSX

- **ddah-tests**
- [x] getting ddahs for one session will not return ddahs for another session

Altair59:
#486 implement tests for exporting data to a JSON/CSV/XLSX file and importing data from a JSON/CSV/XLSX file


frontend/src/tests/import-export.test.js
```diff
@@ -44,6 +55,89 @@ describe("Import/export library functionality", () => {
...
+
+        throw new Error(
+            `Cannot process data to format "${dataFormat}"; try "spreadsheet" or "json".`
+        );
+    }
+
+    /**
+     * Construct a FileAPI `File` object based on pre-defined instructor data and input file type
+     *
+     * @param {"xlsx" | "csv" | "json"} dataFormat
```
frankcbw: I noticed that "spreadsheet" is used for dataFormat, instead of "xlsx". Would it be easier and more consistent if use "xlsx" instead?

frontend/src/tests/import-export.test.js
```diff
@@ -44,6 +55,89 @@ describe("Import/export library functionality", () => {
         },
     ];
 
+    /**
+     *  Create a `File` object containing of the specified format.
```
siefkenj: Extra space char...

frontend/src/tests/import-export.test.js
```diff
@@ -44,6 +55,89 @@ describe("Import/export library functionality", () => {
...
+     * This function is a copy of function dataToFile in ../libs/importExportUtils.
+     * The only difference is this function returns a FileAPI File object but not a File object.
+     * The reason is Jest is running on NodeJS which does not support File object.
+     * By calling this function only for testing, the app's export functionality will not be affected,
+     * while the test can still test the export functionality.
```
siefkenj: Can't you do `global.File = FileAPI.File` and then use the original function?

frontend/src/tests/import-export.test.js
```diff
@@ -107,6 +201,146 @@ describe("Import/export library functionality", () => {
...
+                year: "numeric",
+                month: "numeric",
+                day: "numeric",
+            })}.xlsx`
+        );
+
+        // invalid file type, should throw error
+        expect(() => getInstructorsDataFile("nonsense")).toThrow(Error);
+
+        const resultCorrect = instructorData.map((instructor) =>
```
siefkenj: I don't understand this variable name..

frontend/src/tests/import-export.test.js
```diff
@@ -107,6 +201,146 @@ describe("Import/export library functionality", () => {
...
+            }, {});
+        });
+        const resultXLSX = normalizeImport(
+            {
+                fileType: "spreadsheet",
+                data: dataXLSX,
+            },
+            instructorSchema
+        );
+        expect(resultXLSX).toEqual(resultCorrect);
```
siefkenj: I think we can trust the `xslx` to be working correctly as long as the CSV is working correctly.

frontend/src/tests/import-export.test.js
```diff
@@ -107,6 +201,146 @@ describe("Import/export library functionality", () => {
...
+            }, {});
+        });
+        const resultCSV = normalizeImport(
+            {
+                fileType: "spreadsheet",
+                data: dataCSV,
+            },
+            instructorSchema
+        );
+        expect(resultCSV).toEqual(resultCorrect);
```
siefkenj: This tests for a round trip, but it doesn't test that the output has the required fields. We need to test the CSV to make sure it has first names/last names/etc. for the required data. This will be happening in every test, so you might make a utility function that's something like `csvContainsData(csv, requiredData)` to test if a CSV has what it's supposed to.

A big reason for this test is to catch bugs that we had before where, for example, the export function exported `fist_name` twice and never exported `last_name`

frontend/src/tests/samples/instructors.json
```diff
@@ -0,0 +1,22 @@
...
+        },
+        {
+            "first_name": "Tommy",
+            "last_name": "Smith",
+            "email": "a@b.com",
+            "utorid": "food"
+        },
+        {
+            "first_name": "Grandpa",
+            "last_name": "Boobie",
```
siefkenj: Maybe change this...

frontend/src/tests/import-export.test.js
Altair59: the file-api I used has different constructor than the original File API, and it requires a name field to exist thus conflicts with the original File API's constructor.

siefkenj: Can you create a shim that imitates the native file object? I really don't like this duplicated code.

frontend/src/tests/import-export.test.js
siefkenj: It seems like this still hasn't been dealt with...

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,171 @@
...
+        );
+    }
+
+    /**
+     * Construct a FileAPI `File` object based on pre-defined object data and input file type
+     *
+     * @param {"xlsx" | "csv" | "json"} dataFormat
+     * @returns {FileAPI.File}
+     */
+    function getInstructorsDataFile(dataFormat) {
```
siefkenj: Isn't this another redefinition of an existing function? Can you export it from where it's originally defined?

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,171 @@
...
+                    ),
+                }),
+            },
+            dataFormat,
+            "instructors"
+        );
+    }
+
+    /**
+     * Tests on importing data from files
```
siefkenj: This comment doesn't give me a clue about what the function does. Try again.

frontend/src/tests/import-export.test.js
```diff
@@ -107,6 +231,173 @@ describe("Import/export library functionality", () => {
         ).not.toEqual(targetData);
     });
 
-    it.todo("Export data to a JSON/CSV/XLSX");
-    it.todo("Import data from a JSON/CSV/XLSX");
+    it("Export data to a JSON/CSV/XLSX", () => {
+        // export instructor data a CSV File object
+        const fileCSV = getInstructorsDataFile("csv");
+        expect(fileCSV.name).toEqual(
```
siefkenj: It's not important to test the name.

frontend/src/tests/import-export.test.js
siefkenj: Please change this variable name to something that makes sense.

frontend/src/tests/test_data.js
```diff
@@ -0,0 +1,258 @@
+import instructorsJSON from "./samples/instructors.json";
```
siefkenj: `test_data.js` should not be in the test's directory. It should be in a subfolder where we can keep all the data-related files together.

frontend/src/tests/test_data.js
```diff
@@ -0,0 +1,258 @@
...
+import applicantsJSON from "./samples/applicants.json";
+import positionJSON from "./samples/positions.json";
+import wrong1positionsJSON from "./samples/wrong1positions.json";
+import wrong3positionsJSON from "./samples/wrong3positions.json";
+import assignmentsJSON from "./samples/assignments.json";
+import ddahsJSON from "./samples/ddahs.json";
```
siefkenj: `samples` is not such a good name. `test-data` or `testing-data` or something would be better. Also did you look into Jest `snapshots`?

frontend/src/tests/test_data.js
```diff
@@ -0,0 +1,258 @@
+import instructorsJSON from "./samples/instructors.json";
+import applicantsJSON from "./samples/applicants.json";
+import positionJSON from "./samples/positions.json";
+import wrong1positionsJSON from "./samples/wrong1positions.json";
```
siefkenj: Be more descriptive with the file names. It's okay if they're really long as long as it's clear what the data is.

frontend/src/tests/test_data.js
Altair59: I was looking at [https://jestjs.io/docs/en/snapshot-testing](url) about snapshots. I am kind of confused about how this should work in the case of our import from file test. I think the manual comparison of contents is enough maybe? Can you give me a hint on how I may utilize snapshots?

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,173 @@
...
+    applicantSchema,
+    positionSchema,
+    assignmentSchema,
+    ddahSchema,
+    instructorData,
+    applicantData,
+    positionData,
+    assignmentData,
+    ddahData,
+} from "./test-data-file/test_data";
```
siefkenj: Is this directory a file? If not, it shouldn't have the word `file` in its name.

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,173 @@
...
+
+/* eslint-env node */
+var FileAPI = require("file-api"),
```
siefkenj: Don't use `var`

siefkenj: Why aren't you using `import` syntax?

frontend/src/tests/import-export.test.js
Altair59: Yes it is, but the original one ("prepareData" in \frontend\src\views\instructors\import-export.js) is inside another function, and I can not export it because it's not at top level.

frontend/src/tests/import-export.test.js
Altair59: Instead of going round trip (using normalizeImport to get the data back and check), should I manually check that the "File" object created contains all columns from the data and each cell is identical with the data?

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,226 @@
...
+    assignmentSchema,
+    ddahSchema,
+    instructorData,
+    applicantData,
+    positionData,
+    assignmentData,
+    ddahData,
+} from "./test-data-collection/test_data";
+
+/* eslint-env node */
```
siefkenj, line 27: Why is this comment here and not the top of the file?

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,226 @@
...
-            "Last Name": "last_name",
-            Surname: "last_name",
-            "Family Name": "last_name",
-            Last: "last_name",
-        },
-        requiredKeys: ["utorid"],
-        primaryKey: "utorid",
-    };
+    /**
+     * Construct a FileAPI `File` object based on pre-defined object data and input file type
```
siefkenj, line 38: I don't understand this description

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,226 @@
...
+                    ),
+                }),
+            },
+            fileType,
+            `${objectType}s`
+        );
+    }
+
+    /**
+     * Given correct object data, object type, correct object schema, and JSON File object,
```
siefkenj, line 67: I don't understand this description...

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,226 @@
...
+     * carry out tests to check whether exporting correct object data to file
+     * will produce correct object data file.
+     *
+     * @param {[object]} objectData
+     * @param {string} objectType
+     * @param {object} objectSchema
+     * @param {string} fileType
+     * @returns
+     */
+    function testExportToFile(objectData, objectType, objectSchema) {
```
siefkenj, line 145: These functions shouldn't be defined in the `describe` block. They should probably be in their own utils file.

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,226 @@
...
+    /**
+     * Given correct object data, object type, correct object schema, and file type,
+     * carry out tests to check whether importing from corresponding file containing correct corresponding object data
+     * will produce correct object data.
+     *
+     * @param {[object]} objectData
+     * @param {string} objectType
+     * @param {object} objectSchema
+     * @param {string} fileType
+     * @param {string} correctness
```
siefkenj, line 176: What does this do, and why is it a string?

frontend/src/tests/import-export.test.js
```diff
@@ -2,47 +2,226 @@
...
+                        ],
+                },
+                objectSchema
+            );
+            expect(resultJSON).toEqual(objectData);
+        } else if (fileType == "spreadsheet") {
+            // import object data from a CSV File
+            const workbook1 = XLSX.readFile(
+                __dirname +
+                    `/test-data-collection/${correctness}${objectType}s.csv`
```
siefkenj, line 203: Yuck...it's not a good idea to do string manipulation when looking for files...it's better to hard-code the file names.

Altair59:
1. import test need to only test Object data -> File object -> Object data once, pick just one type of Object
2. test each type of Object's prepareData function
3. test each type of Object's prepareMinimal function
4. tests on correct Object data and some with strange characters

siefkenj:
>     1. import test need to only test Object data -> File object -> Object data once, pick just one type of Object
> 
>     2. test each type of Object's prepareData function
> 
>     3. test each type of Object's prepareMinimal function
> 
>     4. tests on correct Object data and some with strange characters

Did you add this list to the actually issue?

Altair59:
> 
> 
> > ```
> > 1. import test need to only test Object data -> File object -> Object data once, pick just one type of Object
> > 
> > 2. test each type of Object's prepareData function
> > 
> > 3. test each type of Object's prepareMinimal function
> > 
> > 4. tests on correct Object data and some with strange characters
> > ```
> 
> Did you add this list to the actually issue?

Yes, they are in issue #394

Altair59:
This PR becomes outdated, and should only be used for reference. Please see latest PRs on testing export/import functionalities.
#507

PR #489, opened 2021-01-28
Implement ddah test

frankcbw:
Implemented test case "getting ddahs for one session will not return ddahs for another session"


Altair59 (commented):
As I understand, this code checks that every ddah fetched from one session is associated to one valid assignment from this session, thus can fulfill the test's goal to see if any ddah is from another session.

Accelsnow:
It feels weird that the original test suite creates a global ddah object for all test cases, which creates a layer of dependency among all test cases. Is there a way to make each of the ddah tests independent?

frankcbw:
One potential issue:

Under test "getting ddahs for one session will not return ddahs for another session", I tried to firstly add more ddahs with different assignments'ID, as below:

```
it("getting ddahs for one session will not return ddahs for another session", async () => {
        const newDdah = {
            assignment_id: assignments[1].id,
            duties: [
                {
                    order: 2,
                    hours: 25,
                    description: "Marking the midterm",
                },
                {
                    order: 1,
                    hours: 4,
                    description: "Initial training",
                },
                {
                    order: 3,
                    hours: 40,
                    description: "Running tutorials",
                },
            ],
        };

        const resp0 = await apiPOST(`/admin/ddahs`, newDdah);
        expect(resp0).toHaveStatus("success");

        // retrieve all ddahs for the session
        resp1 = await apiGET(`/admin/sessions/${session.id}/ddahs`);
        expect(resp1).toHaveStatus("success");
        const allDdah = resp1.payload;

        // retrieve all the assignments for the seesion
        const resp2 = await apiGET(`/admin/sessions/${session.id}/assignments`);
        expect(resp2).toHaveStatus("success");
        const allAssignmentId = resp2.payload.map((y) => y.id);

        // check that all the ddahs are associated with an assignment from the session
        allDdah.forEach((d) => {
            expect(allAssignmentId).toContain(d.assignment_id);
        });
    });
```
And it fails only MockAPI only with error message:
```
● Mock API tests › DDAH tests › create ddah from assignment route

    expected API response to have {"status": "success"} but received {"message": "Error: Duplicate entry exists for property assignment_id with value 3", "status": "error"}

      197 |             newDdah
      198 |         );
    > 199 |         expect(resp1).toHaveStatus("success");
          |                       ^
      200 |         expect(resp1.payload).toMatchObject({
      201 |             ...newDdah,
      202 |             assignment_id: newAssignment.id,
```

Without the insertion at the beginning, both API and Mock passed.

frontend/src/tests/ddah-tests.js
```diff
@@ -88,9 +88,22 @@ export function ddahTests(api) {
...
+
+        // retrieve all the assignments for the seesion
+        const resp2 = await apiGET(`/admin/sessions/${session.id}/assignments`);
+        expect(resp2).toHaveStatus("success");
+        const allAssignmentId = resp2.payload.map((y) => y.id);
+
+        // check that all the ddahs are associated with an assignment from the session
+        allDdah.forEach((d) => {
+            expect(allAssignmentId).toContain(d.assignment_id);
+        });
```
siefkenj, line 162: The default seed data only has one session and so only puts DDAHs in for that session. You need to create another session, a new assignment for that session, and a DDAH for that assignment. Then make sure fetching the DDAHs for the seed session and the newly created session return different objects.

frontend/src/tests/ddah-tests.js
```diff
@@ -77,20 +79,91 @@ export function ddahTests(api) {
...
         expect(resp1.payload).toContainObject(ddah);
     });
 
-    it.todo(
-        "getting ddahs for one session will not return ddahs for another session"
-    );
+    it("getting ddahs for one session will not return ddahs for another session", async () => {
+        // first create a new session filled with new position, assignment, and contract.
+        const newSession = await addSession(
+            { apiGET: apiGET, apiPOST: apiPOST },
```
siefkenj: You can just pass `api`

frontend/src/tests/ddah-tests.js
```diff
@@ -77,20 +79,91 @@ export function ddahTests(api) {
...
+            { contract_templates: true }
+        );
+        let resp1 = await api.apiGET(
```
siefkenj: No need to call it `resp1` if there is no `resp2`

frontend/src/tests/ddah-tests.js
```diff
@@ -77,20 +79,91 @@ export function ddahTests(api) {
...
+
+        const newPosition = {
+            position_code: "MAT135F",
+            position_title: "Calculus I",
+            hours_per_assignment: 70,
+            start_date: newSession.start_date,
+            end_date: newSession.end_date,
+            contract_template_id: newContract.id,
+        };
+        resp1 = await api.apiPOST(
```
siefkenj: `apiPOST` is in the global namespace. No need to use `api.apiPOST`

osamaramihafez:
Why are many commits from other branches appearing as changes on this branch? It makes it seem like there are too many file changes.

siefkenj:
> Why are many commits from other branches appearing as changes on this branch? It makes it seem like there are too many file changes.

You need to merge `upstream/master` into your branch

frankcbw:
That is what happened when I tried to fetch and merge from `upstream/master`. I am not sure if it is supposed to show such many fille changes, because It does make it difficult for reviewing.

frankcbw:
Also, have fixed issue #493

PR #491, opened 2021-02-01
[backend] Add Reporting Tag Routes

siefkenj:
Add various `/reporting_tags` routes to the backend.

### DIFF
--- a/backend/app/controllers/api/v1/admin/session_reporting_tags_controller.rb
+++ b/backend/app/controllers/api/v1/admin/session_reporting_tags_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class Api::V1::Admin::SessionReportingTagsController < ApplicationController
+    before_action :find_session
+
+    # GET /sessions/:session_id/positions/reporting_tags
+    def index_by_position
+        render_success ReportingTag.position_reporting_tags_by_session(
+                           @session.id
+                       )
+    end
+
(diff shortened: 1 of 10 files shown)


backend/app/controllers/api/v1/admin/wage_chunk_reporting_tags_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+            reporting_tag =
+                ReportingTag.find_or_create_by(name: reporting_tag_name)
+            reporting_tag.save!
+
+            # A reporting tag can show up at most once in a reporting tags list,
+            # so check if it's already included
+            unless @wage_chunk.reporting_tag_ids.include?(reporting_tag.id)
+                @wage_chunk.reporting_tags << reporting_tag
+            end
+            @wage_chunk.save!
```
Accelsnow, line 23: Can we move this save action inside the unless clause since it is only updated inside?

backend/app/controllers/api/v1/admin/session_reporting_tags_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+                           @session.id
+                       )
+    end
+
+    private
+
+    def find_session
+        @session = Session.find(params[:session_id])
+    end
+end
```
Altair59, line 25: Is my understanding correct that we can get all reporting_tags from a session by session_id, but to update or delete them we have to make request under otther subroutes (wage_chunks and positions)?

PR #492, opened 2021-02-01
[backend] Add `instructor/instructors` POST route

siefkenj:
Adds the ability for an instructor to update their name/email, but *not* their UTORid. Also, they cannot update the information of any other instructor.


backend/app/controllers/api/v1/instructor/instructors_controller.rb
```diff
@@ -13,10 +13,48 @@ def index
...
+
+        render_on_condition(
+            object: @instructor,
+            condition: proc { @instructor.update!(instructor_params) }
+        )
+    end
+
+    private
+
+    # Find an instructor looking up by `id` first and then by `utorid`
```
Accelsnow, line 49: If we use utorid as a primary key and it is required by the route, why would we want to also query by id?
To think the other way around, in our last meeting we talked about how having an id implies it is an update, then why would we want to also update by utorid when id is not given? Is it a special case?

siefkenj: It is possible to change an instructor's UTORid

backend/app/controllers/api/v1/instructor/instructors_controller.rb
```diff
@@ -13,10 +13,48 @@ def index
...
+                    }'"
+            )
+        end
+
+        render_on_condition(
+            object: @instructor,
+            condition: proc { @instructor.update!(instructor_params) }
+        )
+    end
+
```
Altair59, line 46: As I understand, this code is correct. This code firstly checks if the current active user matches the instructor proposed to be changed by utorid, since the rule is that instructor can only update itself. If the check passes, the instructor is updated.

PR #496, opened 2021-02-12
Implement ddah test (new)

### LINKED ISSUE #493: [frontend]Potential insufficient testcase: get all ddahs associated with a session
Regarding the test case below, which is a part of ddah-tests, it should get all associated ddahs from a session. Thus, I don't think checking if the payload contains one specific ddah is sufficient.

```
it("get all ddahs associated with a session", async () => {
        resp1 = await apiGET(`/admin/sessions/${session.id}/ddahs`);
        expect(resp1).toHaveStatus("success");
        expect(resp1.payload).toContainObject(ddah);
    });
```

frankcbw:
- Implemented test case "getting ddahs for one session will not return ddahs for another session"
- removed unnecessary resp variables
- fixed issue #493

### DIFF
--- a/frontend/src/tests/ddah-tests.js
+++ b/frontend/src/tests/ddah-tests.js
@@ -99,7 +99,7 @@ export function ddahTests(api) {
             `/admin/sessions/${newSession.id}/contract_templates`
         );
         expect(resp).toHaveStatus("success");
-        console.log(resp.payload);
+        // console.log(resp.payload);
         expect(resp.payload.length).toBeGreaterThan(0);
         const newContract = resp.payload[0];
 
--- a/frontend/src/tests/ddah-tests.js
+++ b/frontend/src/tests/ddah-tests.js
@@ -98,7 +98,7 @@ export function ddahTests(api) {
         let resp = await apiGET(
             `/admin/sessions/${newSession.id}/contract_templates`
         );
-        expect(resp).toHaveStatus("error");
+        expect(resp).toHaveStatus("success");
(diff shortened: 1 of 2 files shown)


frontend/src/tests/ddah-tests.js
```diff
@@ -65,32 +67,100 @@ export function ddahTests(api) {
...
     });
 
     it("get all ddahs associated with a session", async () => {
-        resp1 = await apiGET(`/admin/sessions/${session.id}/ddahs`);
-        expect(resp1).toHaveStatus("success");
-        expect(resp1.payload).toContainObject(ddah);
+        let resp = await apiGET(`/admin/sessions/${session.id}/ddahs`);
+        expect(resp).toHaveStatus("success");
+        expect(resp.payload.length).toEqual(1);
+        expect(resp.payload).toContainObject(ddah);
```
Accelsnow, line 92: Why are we using toEqual() on line 73 while using toContainObject() on line 91? It might be better to use the same comparison method for all ddah tests.

frankcbw: Although I didn't write the first test case (i.e. "create a ddah"), I think the reason behind that is the first test case is testing for creating a ddah, and hence it tries to verify every aspect of the ddah in a more detailed way. On the other hand, the other test case ("get all ddahs associated with a session") wants to make sure all ddahs associated with a session can be retrieved correctly.

siefkenj:
Can you update the description of this PR and fix the failing tests?

PR #498, opened 2021-02-18
Delete ddah button

Accelsnow:
Closes #463 

Implemented ddah deletion button on the frontend and necessary ruby routes on the backend.

PR #499, opened 2021-02-18
Offer withdraw confirmation

### LINKED ISSUE #464: Confirmation when withdrawing multiple offers
When multiple offers are checked and "withdraw" is selected, a confirmation dialog should appear.

Accelsnow:
Closes #464 

Implemented warning dialog when withdrawing from multiple offers.


Accelsnow:
show `name position (hours)` confirmation information in main body. Number of operations 
header to "withdrawing multiple offers"
"cancel, withdraw"

frontend/src/views/assignments/offer-actions.js
```diff
@@ -58,11 +59,29 @@ function OfferActionButtons(props) {
...
+    const showOfferWithdrawConfirmation = () => setShow(true);
+
     function createOffers() {
         for (const assignment of selectedAssignments) {
             offerForAssignmentCreate(assignment);
         }
     }
+    function confirmOfferWithdraw() {
+        // if withdrawing multiple offers at once, show confirmation
+        if (selectedAssignments && selectedAssignments.length > 1) {
```
siefkenj: Will `selectedAssignments` ever not be a list? If it can be `null`, use `selectedAssignments?.length > 1`

frontend/src/views/assignments/offer-actions.js
```diff
@@ -58,11 +59,29 @@ function OfferActionButtons(props) {
...
+            showOfferWithdrawConfirmation();
+        } else {
+            // does not need confirmation if only withdrawing one offer
+            executeConfirmedOfferWithdraw();
+        }
+    }
+    function executeConfirmedOfferWithdraw() {
+        withdrawOffers();
+        hideOfferWithdrawConfirmation();
```
siefkenj: This can safely be called at any time, whether the dialog is open or not, so move it into `withdrawOffers` and we can eliminate a function.

frontend/src/views/assignments/offer-actions.js
```diff
@@ -147,6 +166,23 @@ function OfferActionButtons(props) {
             >
                 Set as Rejected
             </ActionButton>
+            <Modal show={show} onHide={hideOfferWithdrawConfirmation}>
+                <Modal.Header closeButton>
+                    <Modal.Title>Caution</Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    You are withdrawing from multiple offers! Are you sure?
```
siefkenj: Number of offers to withdraw?

frontend/src/views/assignments/offer-actions.js
```diff
@@ -147,6 +171,36 @@ function OfferActionButtons(props) {
...
+                    setDdahDeletionConfirmationVisible(false);
+                }}
+            >
+                <Modal.Header closeButton>
+                    <Modal.Title>Withdrawing Multiple Offers</Modal.Title>
+                </Modal.Header>
+                <Modal.Body>
+                    You are withdrawing from the following{" "}
+                    {selectedAssignments?.length} offers!
+                    <br />
```
siefkenj: See comments on PR #498

frontend/src/views/assignments/offer-actions.js
```diff
@@ -147,6 +163,47 @@ function OfferActionButtons(props) {
...
+                    </div>
+                    <div className="mb-3">
+                        {selectedAssignments.map((selectedAssignment) => {
+                            return (
+                                <li>
+                                    {`${selectedAssignment.applicant.first_name} 
+                                    ${selectedAssignment.applicant.last_name}: 
+                                    ${selectedAssignment.position.position_code} -
+                                    ${selectedAssignment.position.position_title} 
+                                    (${selectedAssignment.hours} hrs)`}
```
siefkenj: Same comments as in other PR apply.

frontend/src/views/assignments/offer-actions.js
```diff
@@ -147,6 +230,38 @@ function OfferActionButtons(props) {
             >
                 Set as Rejected
             </ActionButton>
+            <Modal
```
siefkenj: Make the entire modal a component. Then you won't need to separate out the Table as well. This is also getting big enough it should probably be in its own file.

frontend/src/views/assignments/offer-actions.js
```diff
@@ -58,15 +59,30 @@ function OfferActionButtons(props) {
         setOfferForAssignmentRejected,
     } = props;
 
+    const [
+        ddahDeletionConfirmationVisible,
+        setDdahDeletionConfirmationVisible,
```
frankcbw: I personally feel that the names are a bit too long...

frankcbw: Maybe something like `showWithdrawConfirm` or similar?

siefkenj: `showDeleteConfirmation`

frontend/src/views/assignments/withdraw-assignment-confirmation.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export function MultiWithdrawOfferConfirmation(props) {
+    const dispatch = useDispatch();
+    const { data, visible, setVisible, withdrawOffers } = props;
+
+    // We want to minimize the re-render of the table. Since some bindings for columns
+    // are generated on-the-fly, memoize the result so we don't trigger unneeded re-renders.
+
+    data.sort((a1, a2) => {
```
siefkenj: You should make a separate `compareAssignment` function.

frontend/src/views/assignments/offer-actions.js
```diff
@@ -58,15 +59,29 @@ function OfferActionButtons(props) {
         setOfferForAssignmentRejected,
     } = props;
 
+    const [showWithdrawConfirmation, setShowWithdrawConfirmation] = React.useState(
```
frankcbw: A bit curious about where the `setShowWithdrawConfirmation` setter is defined?

frankcbw: I figured out, the `useState` hook takes care of it.

frontend/src/views/assignments/withdraw-assignment-confirmation.js
```diff
@@ -0,0 +1,101 @@
...
+            <Modal.Header closeButton>
+                <Modal.Title>Withdrawing Multiple Offers</Modal.Title>
+            </Modal.Header>
+            <Modal.Body>
+                <div className="mb-3">
+                    You are withdrawing from the following {data?.length} offers
+                </div>
+                <div className="mb-3">
+                    <AdvancedFilterTable
+                        filterable={true}
```
frankcbw: Do we need to make it filterable?

frontend/src/views/assignments/withdraw-assignment-confirmation.js
```diff
@@ -0,0 +1,101 @@
...
+            <Modal.Footer>
+                <Button
+                    onClick={() => {
+                        setVisible(false);
+                    }}
+                    variant="light"
+                >
+                    Cancel
+                </Button>
+                <Button onClick={withdrawOffers}>Withdraw</Button>
```
frankcbw: I think it will need to `setVisible(false)` after clicking withdraw

frontend/src/views/assignments/withdraw-assignment-confirmation.js
frankcbw: I wonder if we really need to sort them in advance? Because the table component we are using already has a sorting feature.

siefkenj:
You need to merge master, otherwise your Travis builds will always fail.

frontend/src/views/assignments/withdraw-assignment-confirmation.js
Accelsnow: > using

Yes we do. Last meeting we discussed that they need to be sorted by this order by default, but still support additional sorting.

frontend/src/views/assignments/withdraw-assignment-confirmation.js
Accelsnow: > I think it will need to `setVisible(false)` after clicking withdraw

It is invoked inside the withdrawOffers function

PR #500, opened 2021-02-19
[frontend] add a confirm dialog - creating new offers for withdrawns

### LINKED ISSUE #451: Warning when creating an new offer in place of a withdrawn one
We can create a new offer if the active offer is withdrawn. However, there should be a confirmation dialog in this case, otherwise it is easy to use the "select all" button and accidentally create a new offer when you had no intention of doing so.

frankcbw:
Regarding issue #451, added a confirmation dialog if selected assignments contain any withdrawn offers.

Also noticed that the withdraw option is not disabled if the selected assignments already contain withdrawn offers.


frankcbw:
Adding information like name, position, hours, total number of offers.

frontend/src/views/assignments/offer-actions.js
```diff
@@ -59,6 +59,15 @@ function OfferActionButtons(props) {
     } = props;
 
     function createOffers() {
+        if (
+            selectedAssignments.some(
+                (assignment) => assignment.active_offer_status === "withdrawn"
+            )
+        ) {
+            window.confirm(
```
Accelsnow: Do we want to use native window confirm dialog here or should we use react bootstrap's Modal with extra information displayed?

frankcbw: Don't worry. I will change to React bootstrap after you have finished your PR since we are modifying the same file.

frontend/src/views/assignments/create-assignment-confirmation.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+    return 0;
+}
+
+export function CreateOfferConfirmation(props) {
+    const dispatch = useDispatch();
+    const { data, visible, setVisible, createOffers } = props;
+
+    // We want to minimize the re-render of the table. Since some bindings for columns
+    // are generated on-the-fly, memoize the result so we don't trigger unneeded re-renders.
```
siefkenj, line 20: This comment is far away from where the memoization is actually happening...

siefkenj:
Was this superseded by #538?

frankcbw:
> Was this superseded by #538?

Yes, I believe so.

siefkenj:
Resolved by #538

PR #501, opened 2021-02-19
hoist all prepareData functions

Altair59:
Hoist the prepareData functions of instructor, applicant, position, assignment, and ddah for the purpose of import from file test

### DIFF
--- a/frontend/src/libs/exportUtils.ts
+++ b/frontend/src/libs/exportUtils.ts
@@ -20,7 +20,6 @@ import {
     MinimalDdah,
     Ddah,
 } from "../api/defs/types";
-import { dataToFile } from "./importExportUtils";
 
 /**
  * Determine whether `wageChunks` can be derived from `session`. E.g.,
--- a/frontend/src/libs/exportUtils.ts
+++ b/frontend/src/libs/exportUtils.ts
@@ -20,6 +20,7 @@ import {
     MinimalDdah,
     Ddah,
 } from "../api/defs/types";
+import { dataToFile } from "./importExportUtils";
 
 /**
  * Determine whether `wageChunks` can be derived from `session`. E.g.,
--- a/frontend/src/libs/importExportUtils.js
+++ b/frontend/src/libs/importExportUtils.js
@@ -5,7 +5,6 @@
 import FuzzySet from "fuzzyset";
 import XLSX from "xlsx";
 import * as chrono from "chrono-node";
-import { prepareMinimal } from "./exportUtils";
 
 /**
  * Validates `data` based on the specified `schema`. At the moment this
@@ -333,11 +332,10 @@ export function dataToFile(formatters, dataFormat, filePrefix = "") {
 /**
  * Return an array of [hours, duty, hours duty, ...] for the specified `ddah`
  *
- * @export
  * @param {Ddah} ddah
  * @returns {((string | number)[])}
  */
-export function flattenDuties(ddah) {
+function flattenDuties(ddah) {
     const ret = [];
     const duties = [...ddah.duties];
     duties.sort((a, b) => a.order - b.order);
@@ -353,11 +351,10 @@ export function flattenDuties(ddah) {
 /**
  * Format a date as YYYY-MM-DD for inserting into a spreadsheet
  *
- * @export
  * @param {*} date
  * @returns
  */
-export function formatDateForSpreadsheet(date) {
+function formatDateForSpreadsheet(date) {
     try {
         return date && new Date(date).toJSON().slice(0, 10);
     } catch (e) {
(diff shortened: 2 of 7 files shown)


frontend/src/views/applicants/import-export.tsx
```diff
@@ -18,6 +18,48 @@ import {
     ApplicantsDiffList,
 } from "../../components/applicants";
 
+/**
+ * Converts a list of applicants into a File object
```
siefkenj: That's not what it does...

frontend/src/views/assignments/import-export.js
```diff
@@ -86,6 +86,112 @@ function formatWageChunksToList(wageChunks) {
...
+ * @export
+ * @returns
+ */
+export function prepareData(
+    selectedAssignmentIds,
+    session,
+    createPayPeriodHeaders,
+    formatWageChunksToList,
+    formatDateForSpreadsheet
+) {
```
siefkenj: Don't do this...

Make a `prepareDataFactory(filterFunc)` function which returns a `prepareData` function that looks like all the others, except, `filterFunc` is applied to the data before the rest of the processing.

`filterFunc` should be optional and default to the identity function.

frontend/src/views/applicants/import-export.tsx
Altair59: Changed to comment from original code on master branch

frontend/src/views/assignments/import-export.js
Altair59: created prepareDataFactory and refactored the object's prepareData feature pretty much.

frontend/src/libs/importExportUtils.js
```diff
@@ -328,3 +329,306 @@ export function dataToFile(formatters, dataFormat, filePrefix = "") {
...
+    // Make a function that converts a list of instructors into a `File` object.
+    return function prepareData(objects, dataFormat) {
+        if (objectFilter) {
+            objects = objectFilter(objects);
+        }
+        return dataToFile(
+            {
+                toSpreadsheet: () => prepareSpreadsheet[objectType],
+                toJson: () => ({
+                    [objectType + "s"]: objects.map((object) =>
```
siefkenj: Nope. We're not doing this

frontend/src/libs/importExportUtils.js
```diff
@@ -328,3 +329,306 @@ export function dataToFile(formatters, dataFormat, filePrefix = "") {
...
+/**
+ * A factory function which produces corresponding prepareData function,
+ *  for each type of object (possibly needs filtering or associated session).
+ *
+ * @param {string} objectType
+ * @param {Function | null} objectFilter
+ * @param {Session | null} session
+ * @returns {Function}
+ */
+export function prepareDataFactory(
```
siefkenj: This is too general. We don't want a general factory function. Only one for the specific case where we need it.

Altair59: There are two objects having special cases among five objects. 

1. `Ddah` requires filtering with `selectedDdahIds`
2. `Assignment` requires filtering with `selectedAssignmentIds` and also has different `prepareMinimal` parameters than all the other objects

Just want to confirm, do we want only one factory function for these two objects? Or do we need one factory function for each of them

siefkenj: Nope. A separate factory function for each of those two.

Altair59: Also, I noticed that in the original code, only` Ddah` made `prepareDdahsSpreadsheet` and passed it to `dataToFile`, the other objects do the parsing to spreadsheet inside `dataToFile` parameter. In order for consistency, I extracted all object's parsing to spreadsheet out and created a constant `prepareSpreadsheet`, which functions similarly as `prepareMinimal`. I wonder do we want to keep this change or completely revert to the original code on Master and just make one `prepareData` factory for each of the two cases `Ddah` and `Assignment`?

siefkenj: It's a nice idea to break them all out. It would be nice if we converted the file to TypeScript so that we could have some safety guarantees.

Why are all the helper functions, like `flattenDuties` exported? I would hope they are only used in prepare spreadsheet function...

frontend/src/views/ddahs/import-export.tsx
```diff
@@ -26,6 +27,35 @@ import { ddahTableSelector } from "../ddah-table/actions";
...
+ * @param {Function | null} ddahFilter
+ * @returns {Function}
+ */
+export function prepareDataFactory(ddahFilter?: Function) {
+    // Make a function that converts a list of ddahs into a `File` object.
+    return function prepareData(
+        ddahs: Ddah[],
+        dataFormat: "csv" | "json" | "xlsx"
+    ) {
+        if (ddahFilter && ddahFilter instanceof Function) {
```
siefkenj: If `ddahFilter == false` it will not be an instance of `Function`, so you don't need the first check.

frontend/src/libs/exportUtils.ts
```diff
@@ -20,6 +20,7 @@ import {
     MinimalDdah,
     Ddah,
 } from "../api/defs/types";
+import { dataToFile } from "./importExportUtils";
```
siefkenj: Is this import used at all in this file?

siefkenj:
Great!

PR #503, opened 2021-03-01
crafted object data for export

Altair59:
added instructor, applicant, position, assignment, and ddah object data ready for export test


frontend/src/libs/exportUtils.ts
```diff
@@ -41,6 +41,7 @@ function wageChunksMatchPositionAndSession(
     if (wageChunks.length === 1) {
         const chunk: WageChunk = wageChunks[0];
         if (
+            // shouldn't we also check chunk.hours === assignment.hours? Also sum of hours === assignment.hours when there are multiple chunks?
```
Altair59: @siefkenj Do we need to check chunk.hours === assignment.hours? Also sum of hours === assignment.hours when there are multiple chunks?

siefkenj: Yes, I think we probably should...

siefkenj: Remove this comment and make an issue for it.

Altair59: moved the issue to #504

siefkenj:
Great job!

PR #509, opened 2021-03-09
[Frontend] Fixed table horizontal scrolling issue

frankcbw:
Fixed horizontal scrolling issue in advanced-filter-table, where table headers and table body were not aligned correctly when scrolling.


frankcbw:
Unused lines of code are commented out in case of revision needed later. 
Will be removed when the PR is ready to merge.

Accelsnow:
Looks good! This should automatically resolve the scrolling issue in the delete ddah dialog and offer withdraw dialog, as they also use the AdvaceFilterTable implementations

siefkenj:
Remove the comments and remove `react-custom-scrollbars` from `package.json` if it is no longer needed.

frontend/src/components/filter-table/advanced-filter-table.tsx
```diff
@@ -186,27 +174,19 @@ export function AdvancedFilterTable({
                 <HiddenRowWarning visible={isHiddenRowsSelected} />
                 <div {...table.getTableBodyProps()} className="tbody">
                     <AutoSizer>
-                        {({ width, height }) => {
+                        {({ height }) => {
                             // Don't let the table get too short no matter what
                             //height = Math.max(height, 300);
```
siefkenj: Why is this commented out? If it's not needed, it should be removed.

PR #510, opened 2021-03-09
Export Round Trip test

Altair59:
Add round trip test to Export object data to CSV/XLSX/JSON tests


frontend/src/libs/diffUtils.ts
```diff
@@ -18,8 +18,8 @@ import {
     Ddah,
     Duty,
 } from "../api/defs/types";
-import { prepareMinimal, prepareFull } from "./exportUtils";
-import { matchByUtoridOrName } from "./importExportUtils";
+import { prepareMinimal, prepareFull } from "./import-export/prepareJson";
+import { matchByUtoridOrName } from "./import-export/matchByUtoridOrName";
```
siefkenj, line 22: Let's make a `import-export/index.js` that re-exports everything from the other files. That way we can do `import { prepareMinimal } from "./import-export";` It'll make things a bit cleaner and we won't have to worry about exactly how things are broken up.

frontend/src/tests/import-export.test.js
```diff
@@ -107,6 +115,25 @@ describe("Import/export library functionality", () => {
...
+        const instructorCSV = prepareInstructorData(instructorData, "csv");
+        const workbook = XLSX.read(instructorCSV.fileBits[0], {
+            type: "array",
+        });
+        const sheet = workbook.Sheets[workbook.SheetNames[0]];
+        let dataCSV = XLSX.utils.sheet_to_json(sheet, { header: 1 });
+        const keys = dataCSV.shift();
+        // transform to array of objects
+        dataCSV = dataCSV.map(function (row) {
+            return keys.reduce(function (obj, key, i) {
```
siefkenj, line 130: `reduce` is rarely the right option in Javascript. This would be much clearer as a for loop.

There is also `Object.fromEntries` that you can look at.

PR #511, opened 2021-03-10
Test for import object data from CSV/XLSX/JSON and compute diffs

### LINKED ISSUE #512: Import object data do not check start_date is strictly before end_date
Import functions of position and assignment do not have checks on the input start_date is strictly before end_date. We probably need a check to enforce this...

Altair59:
Craft XLSX/JSON/CSV files containing objects data for import and test the import functionality


Altair59:
Import tests use various import/export util functions, and the structure of these functions are refactored in #510. Currently the import tests still use util functions based on the original `importUtils` and `importExportUtils`. We need a merge of these changes.

Altair59:
the `Ddah`s use special `normalizeImport` function than others, which is in `ddah` view and cannot be exported unless we move this function outside the view. However, the function also depends on various import/export utils, which are refactored in #510. I believe it will be easier to be dealt with after we merged #510 and resolve conflicts first. @siefkenj

siefkenj:
@Altair59 #510 is merged now :-)

frontend/src/libs/import-export/parse-spreadsheet.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-env node */
+import XLSX from "xlsx";
+
+export function parseSpreadsheet(fileName) {
+    const workbook = XLSX.readFile(
+        __dirname + `/../../tests/import-export-data/${fileName}`
```
siefkenj: What is this? This code is not appropriate here. You should only be reading files from the file system within a test, but this function is exported to the entire app...

frontend/src/libs/import-export/parse-spreadsheet.js
```diff
@@ -0,0 +1,19 @@
...
+    );
+    const sheet = workbook.Sheets[workbook.SheetNames[0]];
+    let dataCSV = XLSX.utils.sheet_to_json(sheet, { header: 1 });
+    // transform to array of objects
+    const keys = dataCSV.shift();
+    dataCSV = dataCSV.map(function (row) {
+        return keys.reduce(function (obj, key, i) {
```
siefkenj: Don't use `reduce`. It's confusing.

frontend/src/tests/import-export-data/applicants_missing_required_keys.csv
```diff
@@ -0,0 +1,5 @@
+﻿Last Name,First Name,UTORid,Student Number,email,Phone, age
+Doe,John,,10000000,goofy-duck@donald.com,4166666666,
```
siefkenj: Why are the numbers quoted in `applicants_correct.csv` and not in this one?

frontend/src/tests/import-export-data/import-data.js
```diff
@@ -0,0 +1,13 @@
+import instructorsJSON from "./instructors.json";
+import applicantsJson from "./applicants.json";
+import positionsJson from "./positions.json";
+import assignmentsJson from "./assignments.json";
+import ddahsJson from "./ddahs.json";
+
+export const objectJSON = {
+    instructors: instructorsJSON["instructors"],
```
siefkenj: `instructorJSON.instructors`

frontend/src/tests/import-export.test.js
```diff
@@ -190,4 +185,622 @@ describe("Import/export library functionality", () => {
         const ddahJson = ddahData.map(prepareMinimal.ddah);
         expect(ddahJson).toMatchSnapshot();
     });
+
+    it("Import Instructors from JSON/CSV/XLSX", () => {
+        // import correct instructors from XLSX
+        const normalizedSpreadsheetInstructors = normalizeImport(
+            {
+                fileType: "spreadsheet",
+                data: parseSpreadsheet("instructors_correct.xlsx"),
```
siefkenj, line 194: Put the fancy `__dirname` logic here, rather than in the `parseSpreadsheet` function. Also, if `parseSpreadsheet` is only used for tests, then it should only exist in the `tests` directory.

frontend/src/tests/import-export.test.js
```diff
@@ -190,4 +185,622 @@ describe("Import/export library functionality", () => {
...
+                        last_name: "Garcia",
+                        utorid: "garciae",
+                    },
+                ],
+            })
+            .map((x) => x.utorid);
+    }
+    // compute the difference with existing positions
+    const positionsDiff = diffImport.positions(normalizedSpreadsheetPositions, {
+        positions: [
```
siefkenj: You have some data stored in external `json` files and some of it inline. Why?

frontend/src/tests/import-export.test.js
Altair59: in this case, should I move the `parseSpreadsheet` to `import-export.test` or create a file which exports it under `tests` directory?

frontend/src/tests/import-export.test.js
Altair59: all the inline `json`s are used as "context" in `diffImport` functions, which is considered as the "existing" objects and is compared with the imported objects to find the difference. in this case, should I also store these "context" objects into files or a explanation is enough (keep them inline)?

frontend/src/tests/import-export.test.js
siefkenj: There should be a `utils.js` or something like that in the test directory.

frontend/src/tests/import-export.test.js
siefkenj: Do you think it helps readability to have this JSON inline?

frontend/src/tests/import-export-data/applicants_missing_required_keys.csv
Altair59: sorry... forgot to change this one after fixing the applicants_correct.csv

frontend/src/tests/import-export.test.js
Altair59: that makes sense... moved those to a collection of import object contexts

frontend/src/tests/import-export.test.js
```diff
@@ -190,4 +196,245 @@ describe("Import/export library functionality", () => {
...
+                data: parseSpreadsheet("instructors_correct.xlsx"),
+            },
+            instructorSchema
+        );
+        expect(normalizedSpreadsheetInstructors).toMatchSnapshot();
+        // import correct instructors from JSON
+        const normalizedJsonInstructors = normalizeImport(
+            {
+                fileType: "json",
+                data: objectJSON["instructors"],
```
siefkenj: Why aren't you doing `.instructors` to access this? Also, what does `objectJSON` mean? I don't understand what data `objectJSON` is supposed to contain.

frontend/src/tests/import-export-data/import-context.js
```diff
@@ -0,0 +1,404 @@
+/*
+ * A collection of context (existing objects)
+ * used for computing difference when importing objects.
+ */
+
+export const importInstructorContext = {
```
siefkenj: `Context` is not a descriptive enough term. Come up with better names.

frontend/src/tests/import-export-data/existing-objects-for-import.js
```diff
@@ -1,17 +1,15 @@
 /*
- * A collection of context (existing objects)
+ * A collection of existing objects
  * used for computing difference when importing objects.
  */
 
-export const importInstructorContext = {
+export const existingObjectsForInstructorImport = {
```
siefkenj: I'm still unclear on this. So we have some data, and we have some modified data. We use the `diff` routine on both and expect to get something specific out. Right?

If so, we should co-locate this data. We can have an object with keys `data1` and `data2` and `expectedDiff`. Then, when one of the tests fail, someone will actually know where to look.

Altair59: do you mean instead of taking snapshots of the computed diff, I should take snapshots of a constructed object, which is:
```
{
existingData: existingInstructors,
modifiedData: normalizedInstructors,
expectedDiff: diffImport(normalizedInstructors, existingInstructors)
}
```

siefkenj: Snapshots of the diffs are okay, but you should at lease co-locate the changed and unchanged data. Note `existingData` is a bad name. It doesn't exist until you define it, but the word `existing` makes it sound like it comes from somewhere. `initialData` and `modifiedData` would be better names.

frontend/src/tests/import-export.test.js
```diff
@@ -190,4 +196,265 @@ describe("Import/export library functionality", () => {
...
+        // compute the difference between existing instructors
+        const instructorsDiff = {
+            initialData: initialInstructorData,
+            modifiedData: normalizedSpreadsheetInstructors,
+            expectedDiff: diffImport.instructors(
+                normalizedSpreadsheetInstructors,
+                initialInstructorData
+            ),
+        };
+        expect(instructorsDiff).toMatchSnapshot();
```
siefkenj: Why are you snapshotting the whole object? You should only be snapshotting the diff...

frontend/src/tests/import-export.test.js
```diff
@@ -190,4 +196,265 @@ describe("Import/export library functionality", () => {
...
+                        "instructors_missing_required_keys.xlsx"
+                    ),
+                },
+                instructorSchema
+            ).toThrow(Error)
+        );
+        // compute the difference between existing instructors
+        const instructorsDiff = {
+            initialData: initialInstructorData,
+            modifiedData: normalizedSpreadsheetInstructors,
```
siefkenj: This doesn't help. I cannot see `initialInstructorData` and `normalizedSpreadsheetInstructors` side-by-side. If this test throws an error, it will be really hard to debug. You haven't co-located the data yet...

frontend/src/tests/import-export.test.js
```diff
@@ -446,15 +384,74 @@ it("Import Ddahs from JSON/CSV/XLSX", () => {
...
     expect(normalizedJsonDdahs).toMatchSnapshot();
-    // Compute the difference with existing ddahs
-    const ddahsDiff = {
-        initialData: initialDdahData.ddahs,
-        modifiedData: normalizedJsonDdahs,
-        expectedDiff: diffImport.ddahs(normalizedJsonDdahs, initialDdahData),
-    };
+});
+
+it("Compute Instructors difference", () => {
```
siefkenj: Use the word `diff` instead of `difference`. `diff` is its own noun ;-)

frontend/src/tests/import-export-data/modified-data.js
```diff
@@ -0,0 +1,309 @@
+/*
+ * A collection of modified objects
+ * used for computing difference with initial objects.
```
siefkenj: This is much better! But modified compared to what? Think about the next student who works on this code. They come into this file and see `modifiedInstructorData`. What do they think? Where should they look next? The code and comments should guide them.

If the data were co-located, it would be much clearer. You could see the initial data and the modified data right next to it. Do you see how that would help the reader understand?

frontend/src/tests/import-export-data/diff-data.js
```diff
@@ -0,0 +1,667 @@
+/*
+ * A collection of initial objects and modified objects
+ * used for testing diff functionalities
+ */
+
+// a collection of initial instructors, applicants, positions, assignments, ddahs
+// used for computing the diff with modified objects
+export const initialObjects = {
```
siefkenj: Let's call this `reduxStoreData`

frontend/src/tests/import-export-data/diff-data.js
```diff
@@ -0,0 +1,667 @@
...
+                    description: "Marking the midterm",
+                    hours: 50,
+                },
+            ],
+        },
+    ],
+};
+
+// modified instructor data used for computing the diff with initial objects
+export const modifiedInstructorData = [
```
siefkenj: I thought you were going to make deep copies of the `reduxStoreData` and then modify the parts that you wanted changed...?

frontend/src/tests/import-export-data/diff-data.js
```diff
@@ -0,0 +1,667 @@
...
+ * A collection of initial objects and modified objects
+ * used for testing diff functionalities
+ */
+
+// a collection of initial instructors, applicants, positions, assignments, ddahs
+// used for computing the diff with modified objects
+export const initialObjects = {
+    instructors: [
+        {
+            first_name: "Henry",
```
siefkenj: Shouldn't there be `id` fields on all these objects?

frontend/src/tests/import-export-data/diff-data.js
```diff
@@ -0,0 +1,603 @@
...
+                        rate: 50,
+                    },
+                ],
+            },
+        },
+    ],
+};
+
+// modified instructor data used for computing the diff with initial objects
+export const modifiedInstructorData = reduxStoreData.instructors.map(
```
siefkenj: This is good! Move this creation part into the actual test function, and I think this PR will be ready :-)

BTW, what's the difference between the `reduxStoreData` and what is stored in all the `.json` files spread around?

Altair59: For `json`s used in export tests, for example, `assignments` are designed to reflect any possible situations how output file should look like (depending on `assignment.hours`, `assignment.wage_chunks`). For `json` files used in import tests which are spread around, they may have missing fields, redundant fields, or strange chars. All the `json`s are designed to cover some edge cases.

PR #516, opened 2021-03-12
Finish up export test

Altair59:
1. added comment for parsing File object in round trip test
2. changed reduce to for loop in parsing File object
3. added index.js for import/export util files
4. renamed util files in libs


frontend/src/api/actions/applicants.js
```diff
@@ -7,7 +7,7 @@ import {
 } from "../constants";
 import { fetchError, upsertError, deleteError } from "./errors";
 import { actionFactory, validatedApiDispatcher } from "./utils";
-import { apiGET, apiPOST } from "../../libs/apiUtils";
+import { apiGET, apiPOST } from "../../libs/apis";
```
siefkenj: I think we should make it `api-utils`. Since there is an `api` directory, it can get really confusing if we have multiple files named `api`.

frontend/src/tests/import-export.test.js
```diff
@@ -119,6 +121,7 @@ describe("Import/export library functionality", () => {
         // ROUND TRIP TEST for prepareData function
         // create instructor CSV File object
         const instructorCSV = prepareInstructorData(instructorData, "csv");
+        // fileBits[0] stores the ArrayBuffer object of exported instructors data
```
siefkenj: Why is `fileBits` an array at all? When would there be multiple items in it?

Altair59: it is because in the original code function `dataToFile` in `frontend\src\libs\import-export\data-to-file.js` makes the fileBits an array.

Altair59: according to [File API docs](https://developer.mozilla.org/en-US/docs/Web/API/File/File), it should take an array of ArrayBuffer, etc as the file content. However, since Node does not recognize `File` object and I created a shim for that, I should take `fileBits[0]` to retrieve the ArrayBuffer

siefkenj: So put this explanation in the comment. Think about future readers of the code who have not done the research you have.

PR #521, opened 2021-03-18
Rails erd version update

frankcbw:
Changed `rails-erd` to version 1.6.1


frankcbw:
Closed because of mix-up from two branches

PR #522, opened 2021-03-18
update rails erd to 1.6.1

frankcbw:
- Update `rails-erd` to version 1.6.1
- Change pre-commit hook to run rubocop within docker


siefkenj:
This was updated in a different PR

PR #523, opened 2021-03-19
Fix null undefined

### LINKED ISSUE #444: Import differentiates between `null` and `undefined`
The import diff lists things as changing from `null` to `undefined` as a change. This shouldn't be considered a change as JSON doesn't distinguish between the two.

Altair59:
fixed the issue #444 of mixing `null` and `undefined` during i**mport/export**. Now only `null` value exists after normalizing imported data and exporting data. NOTE: only `undefined` valued are converted to `null`. Empty strings or empty list are considered as valid values so will NOT be converted to `null`


Altair59:
this branch is based on #511, thus we need to merge #511 first

siefkenj:
#511 is in. You should be able to continue working on this now :-)

frontend/src/libs/schema.js
```diff
@@ -0,0 +1,110 @@
+export const instructorSchema = {
+    keys: ["first_name", "last_name", "utorid", "email"],
+    keyMap: {
```
Accelsnow: might need some comment here. It is a little bit confusing why there are multiple keys mapping to the same value "last_name", and why some of the keys are strings in double-quotes while some are not.

frontend/src/tests/utils.js
```diff
@@ -289,6 +291,28 @@ export function checkPropTypes(propTypes, data) {
...
+ */
+export function parseSpreadsheet(fileName) {
+    const workbook = XLSX.readFile(
+        __dirname + `/import-export-data/${fileName}`
+    );
+    const sheet = workbook.Sheets[workbook.SheetNames[0]];
+    let dataCSV = XLSX.utils.sheet_to_json(sheet, { header: 1 });
+    // transform to array of objects
+    const keys = dataCSV.shift();
+    dataCSV = dataCSV.map(function (row) {
```
Accelsnow: I suggest using the same way to define anonymous functions inside array functions (map, some, foreach). It is more concise to use arrow functions

frontend/src/libs/import-export/undefinedToNull.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Util functions to turn undefined to null for json
+ *
+ * @exports
+ * @param {Object} item
+ * @returns{Object}
+ */
+export function jsonUndefinedToNull(items) {
+    const newItems = items.map(function (item) {
```
Accelsnow: It is more concise to use arrow functions inside array functions like map, some and forEach.

frontend/src/libs/import-export/normalize-ddah-import.ts
```diff
@@ -0,0 +1,111 @@
...
+            "Given Name": "first_name",
+            First: "first_name",
+            "Last Name": "last_name",
+            Surname: "last_name",
+            "Family Name": "last_name",
+            Last: "last_name",
+        };
+        // We will also add `Hours #` and `Duty #` to the keymap for the number of duties in our range
+        for (let i = 0; i <= maxDuties; i++) {
+            keyMap[`Duty ${i}`] = `duty_${i}`;
```
Accelsnow: Is this part correct? So duties with numbers less than 10 will have 2 different entries? For example duty 1 will have 2 keys 'Duty 1' and 'Duty 01' with same value 'duty_1'.

frontend/src/libs/import-export/normalize-ddah-import.ts
```diff
@@ -0,0 +1,111 @@
...
+        for (const ddah of unwrapped) {
+            ret.push(ddah);
+        }
+    }
+
+    if (data.fileType === "spreadsheet") {
+        const unwrapped = data.data;
+        // Get an upper bound for the maximum number of duties that the spreadsheet might have
+        const maxDuties = Math.round(
+            Math.max(
```
Accelsnow: Why would `...unwrapped.map((row: object) => Object.keys(row).length)` be negative? if not what is the purpose of taking the max between this value and 0?

frontend/src/libs/import-export/normalize-ddah-import.ts
```diff
@@ -0,0 +1,111 @@
...
+        const rowMapper = new SpreadsheetRowMapper({
+            keys: ["position_code", "first_name", "last_name", "utorid"],
+            keyMap,
+        });
+
+        for (const row of unwrapped) {
+            const normalized: {
+                [key: string]: any;
+            } = rowMapper.formatRow(row);
+            if (normalized.utorid == null) {
```
Accelsnow: just making sure we are including the case where `normalized.utorid` is `undefined` as well but excluding the case where it is an empty string.

frontend/src/libs/import-export/undefinedToNull.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Util functions to turn undefined to null for spreadsheets
+ *
+ * @exports
+ * @param {Object} item
+ * @returns{Object}
+ */
+export function spreadsheetUndefinedToNull(items) {
+    const newItems = items.map(function (item) {
```
Accelsnow: It is more concise to use arrow functions inside array functions like map, some and forEach.

frontend/src/libs/import-export/validate.js
```diff
@@ -8,7 +8,7 @@
  * @param {*} schema
  */
 export function validate(data, schema) {
-    const { requiredKeys } = schema;
+    const { requiredKeys, keys } = schema;
     for (const item of data) {
         for (const key of requiredKeys) {
             if (item[key] == null) {
```
Accelsnow: are we including the case where it is undefined but excluding the case where it is an empty string?

Altair59: yes, and the purpose of this is to convert undefined to null.

frontend/src/libs/import-export/normalize-ddah-import.ts
Altair59: yes, that is correct. Only undefined values are converted to null. Empty strings are considered as valid values

frontend/src/libs/import-export/prepare-data.ts
```diff
@@ -90,9 +104,12 @@ export function prepareDdahDataFactory(ddahFilter: FilterFunc<Ddah>) {
         }
         return dataToFile(
             {
-                toSpreadsheet: () => prepareSpreadsheet.ddah(ddahs),
+                toSpreadsheet: () =>
+                    spreadsheetUndefinedToNull(prepareSpreadsheet.ddah(ddahs)),
                 toJson: () => ({
-                    ddahs: ddahs.map((ddah) => prepareMinimal.ddah(ddah)),
+                    ddahs: jsonUndefinedToNull(
```
siefkenj: `jsonUndefinedToNull` should be called inside the `prepareMinimal`/`prepareSpreadsheet` functions.

frontend/src/libs/schema.js
```diff
@@ -0,0 +1,110 @@
+export const instructorSchema = {
```
siefkenj: I thought I moved all the schemas to a typescript file?

frontend/src/libs/import-export/undefinedToNull.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Util functions to turn undefined to null for json
```
siefkenj: This is confusing, since JSON doesn't have `undefined`. It is already the case that `JSON.stringify` converts `undefined` to `null`. So, why do we need this function?

Also, let's make new files TypeScript by default.

frontend/src/libs/schema.js
Altair59: This is from original code. The purpose is to construct a key-mapping between imported JSON keys/ CSV Column names and actual keys of an object like `instructor`. The comment is added. All spreadsheet columns names are strings. I guess the author of this code want to include non-string keys to include situations of JSON keys.

frontend/src/tests/utils.js
Altair59: For this code, I believe it's clearer to do in this way as I am constructing an object from an array which may be hard to understand if we make it compact using arrow functions. For another part below, I believe you are correct and changed to use arrow functions.

frontend/src/libs/import-export/undefinedToNull.js
Altair59: This function is removed due to redundancy.

frontend/src/libs/import-export/normalize-ddah-import.ts
Altair59: This is correct. The purpose is to recognize both 'Duty 1' and 'Duty 01' both as a valid key-mappings to 'duty_1'.

frontend/src/libs/import-export/normalize-ddah-import.ts
Altair59: Yes this is weird... But since this code is from original code and not relating to this PR, please make a separate issue and we can look into it.

frontend/src/libs/import-export/undefinedToNull.js
Altair59: Changed to arrow functions.

frontend/src/libs/import-export/prepare-data.ts
Altair59: `spreadsheetUndefinedToNull` is moved inside `prepareSpreadsheet` functions.

frontend/src/libs/schema.js
Altair59: removed outdated schema.js

frontend/src/libs/import-export/undefinedToNull.js
Altair59: Removed this redundant function and added a comment explaining why it is not needed.

PR #528, opened 2021-03-25
Re-sending offer emails

osamaramihafez:
When an offer is re-sent, the offer should indicate any differences between it and the last offer sent.

### DIFF
--- a/backend/app/mailers/offer_mailer.rb
+++ b/backend/app/mailers/offer_mailer.rb
@@ -62,7 +62,7 @@ class OfferMailer < ApplicationMailer
                 offer.url_token
             }/view"
         @nag_count = offer.nag_count
-
+        @status_message = @offer.get_status_message
         @diff = get_prev_offer_diff
     end
 end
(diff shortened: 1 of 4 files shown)


backend/app/mailers/offer_mailer.rb
```diff
@@ -38,6 +38,14 @@ def email_nag(offer)
 
     private
 
+    def get_prev_offer_diff
+        id = @offer.assignment_id
+        latest =
+            Offer.where('assignment_id = ?', id).order('emailed_date DESC')
+                .limit(1)
+        return @offer.compute_diff(latest[0])
```
Accelsnow: I think you can omit the 'return' on this line

backend/app/views/offer_mailer/email_contract.html.erb
```diff
@@ -6,11 +6,19 @@
     </style>
   </head>
   <body>
+  <p><%= @status_message %></p>
+  <br>
   <p>Dear <%= @first_name %> <%= @last_name %></p>
   <p>
     We would like to offer you a position as a TA for <%= @position_code %> <%= @position_title %>.
   </p>
-  <p>
```
siefkenj, line 13: Why did you move this to inside if? Whitespace doesn't matter at the start of a `<p>`. Also, are the tags closed correctly?

PR #538, opened 2021-04-08
Multiple Manipulation Confirm Dialog

### LINKED ISSUE #450: Email Confirmation Dialog
The email button should have a confirmation dialog listing how many emails will be sent.

Altair59:
#450 Added a confirmation dialog showing number of offer to be emailed and their brief contents.


frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -59,13 +59,14 @@ function compareAssignment(a1: Assignment, a2: Assignment) {
     );
 }
 
-export function MultiWithdrawOfferConfirmation(props: {
+export function MultiManipulateOfferConfirmation(props: {
```
siefkenj: Let's duplicate code and have Withdraw and Confirm dialogs separate.

frontend/src/views/admin/assignments/email-assignment-confirmation.tsx
```diff
@@ -0,0 +1,118 @@
...
+        // when sorted, so we set their accessor to null (the accessor is used by react table
+        // when sorting items).
+        accessor: (data: { active_offer_status: string }) =>
+            data.active_offer_status === "No Contract"
+                ? null
+                : data.active_offer_status,
+    },
+];
+
+function compareString(str1: string, str2: string) {
```
siefkenj: Don't duplicate these functions. They should be moved to a `utils` file.

Altair59:
All actions on multiple `assignments` or `ddahs` needs a dialog to confirm. Thus, make a dialog component that can be of shared use.

frontend/src/views/admin/ddahs/utils.tsx
```diff
@@ -0,0 +1,59 @@
...
+];
+
+function compareString(str1: string, str2: string) {
+    if (str1 > str2) {
+        return 1;
+    } else if (str1 < str2) {
+        return -1;
+    }
+    return 0;
+}
```
siefkenj: This function has been duplicated again...

frontend/src/views/admin/ddahs/utils.tsx
```diff
@@ -0,0 +1,59 @@
...
+export function compareDDAH(
+    d1: ConfirmationDdahRowData,
+    d2: ConfirmationDdahRowData
+) {
+    return (
+        compareString(d1.position_code, d2.position_code) ||
+        compareString(d1.last_name, d2.last_name) ||
+        compareString(d1.first_name, d2.first_name)
+    );
+}
```
siefkenj: This probably belongs in the import/export utils file.

Altair59: I had moved all `compare` functions into `frontend/src/libs/compare-table-rows.ts` because they're only used for sorting table rows selected and displayed in the modal

frontend/src/libs/compare-table-rows.ts
```diff
@@ -0,0 +1,33 @@
...
+        compareString(
+            a1.applicant.last_name || "",
+            a2.applicant.last_name || ""
+        ) ||
+        compareString(a1.applicant.first_name, a2.applicant.first_name)
+    );
+}
+
+export function compareDDAH(
+    d1: ConfirmationDdahRowData,
```
siefkenj: Should be a `Pick<Ddah, "position_code" | "last_name" | "first_name">`. You shouldn't need to import from a view into a utility. If you've doe such a thing, it's an indication that the code isn't organized correctly.

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -39,33 +40,24 @@ const withdrawModalColumn = [
...
-        compareString(
-            a1.applicant.last_name || "",
-            a2.applicant.last_name || ""
-        ) ||
-        compareString(a1.applicant.first_name, a2.applicant.first_name)
-    );
-}
-
-export function MultiWithdrawOfferConfirmation(props: {
+export function MultiManipulateOfferConfirmation(props: {
```
siefkenj: `Manipulate` is not a descriptive word. How about `OfferConfirmationDialog`

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -39,33 +40,24 @@ const withdrawModalColumn = [
...
+    alertMsg: String;
+    confirmBtnMsg: String;
 }) {
-    const { data, visible, setVisible, withdrawOffers } = props;
+    const {
+        data,
+        visible,
+        setVisible,
+        manipulateOffers,
+        titleMsg,
```
siefkenj: `tileMsg` -> `title`

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -39,33 +40,24 @@ const withdrawModalColumn = [
...
-}
-
-export function MultiWithdrawOfferConfirmation(props: {
+export function MultiManipulateOfferConfirmation(props: {
     data: Assignment[];
     visible: boolean;
     setVisible: (visibility: boolean) => void;
-    withdrawOffers: () => void;
+    manipulateOffers: () => void;
+    titleMsg: String;
```
siefkenj: Avoid capitalized basic types. E.g., use `string`, not `String`.

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -39,33 +40,24 @@ const withdrawModalColumn = [
...
-    );
-}
-
-export function MultiWithdrawOfferConfirmation(props: {
+export function MultiManipulateOfferConfirmation(props: {
     data: Assignment[];
     visible: boolean;
     setVisible: (visibility: boolean) => void;
-    withdrawOffers: () => void;
+    manipulateOffers: () => void;
```
siefkenj: Does this typecheck? Also what does `manipulateOffers` do? Is it actually a callback? If so, call it `callback`

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -39,33 +40,24 @@ const withdrawModalColumn = [
...
+    titleMsg: String;
+    alertMsg: String;
```
siefkenj: What is an `alertMsg`? The name doesn't tell me...

Avoid abbreviating. E.g., use `alertMessage` instead. (but don't actually, because `alertMessage` is not a good name for this variable.)

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
```diff
@@ -107,11 +98,11 @@ export function MultiWithdrawOfferConfirmation(props: {
                 </Button>
                 <Button
                     onClick={() => {
-                        withdrawOffers();
+                        manipulateOffers();
                         setVisible(false);
                     }}
                 >
-                    Withdraw {data?.length} Offers
+                    {confirmBtnMsg}
```
siefkenj: No abbreviations.

siefkenj: This isn't a "message" it's just text that shows up...

frontend/src/views/admin/assignments/manipulate-assignment-confirmation.tsx
siefkenj: After reading the code, it looks like `alertMsg` is actually `body`.

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -136,37 +191,85 @@ function OfferActionButtons(props) {
...
-            <MultiWithdrawOfferConfirmation
+            <MultiManipulateOfferConfirmation
+                data={selectedAssignments}
+                visible={showCreateConfirmation}
+                setVisible={setShowCreateConfirmation}
+                manipulateOffers={createOffers}
+                titleMsg="Creating Multiple Offers"
+                alertMsg={`You are creating the following ${selectedAssignments.length} offers:`}
+                confirmBtnMsg={`Create ${selectedAssignments.length} Offers`}
+            />
```
siefkenj, line 132: We're getting a lot of code in this file...can we split the buttons into their own file and have a `<ConfirmWithDialogActionButton />` which is a component returning `<React.Fragment><ActionButton .../><ConfirmationDialog ... /></React.Fragment>`?

frontend/src/libs/compare-table-rows.ts
Altair59: The `DDAH`s have `position_code`, `last_name`, and `first_name` nested in `assignment.position` or `assignment.applicant`. Using the `Pick` would require some implementation or libraries to support nested `Pick`. which I dont think it's much easier to understand than the original interface `ConfirmationDdahRowData`. Are we going to use Pick in this case?

Altair59: Another easier solution I think I can go is to modify the compare DDAH method itself

siefkenj: Then just define a new type `{position_code: string, last_name: string, first_name: string}`

frontend/src/views/admin/assignments/action-buttons.tsx
```diff
@@ -0,0 +1,63 @@
...
+    FaBan,
+    FaCheck,
+    FaUserTimes,
+    FaUserClock,
+    FaUserPlus,
+} from "react-icons/fa";
+import { ActionButton } from "../../../components/action-buttons";
+
+export function ActionButtons(props: {
+    actionPermitted: { [Key: string]: boolean };
```
siefkenj: I was suggesting you have one object per action with a button/dialog grouped, not one object for buttons and one for dialogs.

frontend/src/views/admin/assignments/confirm-with-dialog-action-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+        actionConfirmation,
+        buttonText,
+        selectedAssignments,
+        dialogVisible,
+        dialogSetVisible,
+        actionCallback,
+        dialogTitle,
+        dialogBody,
+        dialogConfirmation,
+    } = props;
```
siefkenj: You can move this right into the function definition. I.e., use destructuring there instead of temporarily assigning to `props`.

frontend/src/views/admin/assignments/confirm-with-dialog-action-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+    buttonEnable: boolean;
+    actionConfirmation: () => void;
+    buttonText: string;
+    selectedAssignments: Assignment[];
+    dialogVisible: boolean;
+    dialogSetVisible: Function;
+    actionCallback: Function;
+    dialogTitle: string;
+    dialogBody: string;
+    dialogConfirmation: string;
```
siefkenj: There are way too many props here. See how few you can reduce it to.

`dialogConfirmation` <- bad name

frontend/src/views/admin/assignments/confirm-with-dialog-action-button.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { ActionButton } from "../../../components/action-buttons";
+import { OfferConfirmationDialog } from "./offer-confirmation-dialog";
+import { Assignment } from "../../../api/defs/types";
+
+export function ConfirmWithDialogActionButton(props: {
+    buttonIcon: Object;
+    buttonEnable: boolean;
+    actionConfirmation: () => void;
+    buttonText: string;
```
siefkenj: What button?

frontend/src/views/admin/assignments/confirm-with-dialog-action-button.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { ActionButton } from "../../../components/action-buttons";
+import { OfferConfirmationDialog } from "./offer-confirmation-dialog";
+import { Assignment } from "../../../api/defs/types";
+
+export function ConfirmWithDialogActionButton(props: {
+    buttonIcon: Object;
+    buttonEnable: boolean;
+    actionConfirmation: () => void;
```
siefkenj: What's the difference between this and the `actionCallback`?

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -165,53 +172,85 @@ function ConfirmWithDialogActionButton(props) {
...
+                buttonEnable={actionPermitted.canWithdraw}
+                buttonText="Withdraw Offer"
+                selectedAssignments={selectedAssignments}
+                dialogVisible={showWithdrawConfirmation}
+                dialogSetVisible={setShowWithdrawConfirmation}
+                actionCallback={withdrawOffers}
+                dialogTitle="Withdrawing Multiple Offers"
+                dialogBody={`You are withdrawing from the following ${selectedAssignments.length} offers:`}
+                dialogConfirmation={`Withdraw ${selectedAssignments.length} Offers`}
+            />
```
siefkenj, line 137: We're getting closer, but this is still not right. I didn't mean for you to invent a new abstract `ConfirmWithDialogAction` component. I meant for you to create 7 concrete components which repeat the needed code from `ConfirmWithDialogAction`.

frontend/src/views/admin/assignments/accept-offer-button-with-dialog.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { ActionButton } from "../../../components/action-buttons";
+import { OfferConfirmationDialog } from "./offer-confirmation-dialog";
+import { Assignment } from "../../../api/defs/types";
+import { FaCheck } from "react-icons/fa";
+
+export function AcceptOfferButtonWithDialog(props: {
```
siefkenj: This is good! :-).

Do you think it's better to have them in separate files or all in one?

frontend/src/views/admin/assignments/accept-offer-button-with-dialog.tsx
```diff
@@ -0,0 +1,35 @@
...
+    selectedAssignments: Assignment[];
+    dialogVisible: boolean;
+    dialogSetVisible: Function;
+    actionCallback: Function;
+}) {
+    return (
+        <React.Fragment>
+            <ActionButton
+                icon={<FaCheck />}
+                onClick={props.actionButtonOnClick}
```
siefkenj: Rather than accessing props all the time, destructure directly in the function arguments.

frontend/src/views/admin/assignments/accept-offer-button-with-dialog.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { FaCheck } from "react-icons/fa";
+
+export function AcceptOfferButtonWithDialog(props: {
+    actionButtonEnable: boolean;
+    actionButtonOnClick: () => void;
+    selectedAssignments: Assignment[];
+    dialogVisible: boolean;
+    dialogSetVisible: Function;
+    actionCallback: Function;
+}) {
```
siefkenj: This type is large enough it should be given a name and only defined once.

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -59,22 +59,41 @@ function OfferActionButtons(props) {
...
+        false
+    );
     const [
         showWithdrawConfirmation,
         setShowWithdrawConfirmation,
     ] = React.useState(false);
+    const [showEmailConfirmation, setShowEmailConfirmation] = React.useState(
+        false
+    );
+    const [showNagConfirmation, setShowNagConfirmation] = React.useState(false);
```
siefkenj: Handle all of this in the buttons themselves.

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -59,22 +59,41 @@ function OfferActionButtons(props) {
...
+        false
+    );
 
+    function confirmOfferCreate() {
+        if (selectedAssignments?.length > 1) {
+            setShowCreateConfirmation(true);
+        } else {
+            createOffers();
+        }
+    }
```
siefkenj: Have this handled in the `ActionButton`

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -120,53 +171,53 @@ function OfferActionButtons(props) {
...
-                visible={showWithdrawConfirmation}
-                setVisible={setShowWithdrawConfirmation}
-                withdrawOffers={withdrawOffers}
+            <CreateOfferButtonWithDialog
+                actionButtonEnable={actionPermitted.canCreate}
+                actionButtonOnClick={confirmOfferCreate}
+                selectedAssignments={selectedAssignments}
+                dialogVisible={showCreateConfirmation}
+                dialogSetVisible={setShowCreateConfirmation}
+                actionCallback={createOffers}
```
siefkenj: Why is there a `callback` and an `onClick`? (Also, don't invent a new name when `onClick` will do). Create a closure that performs the correct action and pass it as the callback. No need for an `onClick`. The `CreateOfferButtonWithDialog` will handle showing the dialog if needed.

frontend/src/views/admin/assignments/accept-offer-button-with-dialog.tsx
Altair59: Probably all in in one file! Because to destructure props in function arguments would require defining a interface to recognize props type, which should be re-used

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -120,53 +113,35 @@ function OfferActionButtons(props) {
...
-            >
-                Set as Rejected
-            </ActionButton>
-            <MultiWithdrawOfferConfirmation
-                data={selectedAssignments}
-                visible={showWithdrawConfirmation}
-                setVisible={setShowWithdrawConfirmation}
-                withdrawOffers={withdrawOffers}
+            <CreateOfferButtonWithDialog
+                actionButtonEnable={actionPermitted.canCreate}
```
siefkenj: ->`disabled`, and make this an optional prop

frontend/src/views/admin/assignments/offer-button-with-dialog.tsx
```diff
@@ -0,0 +1,251 @@
...
+    selectedAssignments: Assignment[];
+    actionCallback: Function;
+};
+
+export function CreateOfferButtonWithDialog({
+    actionButtonEnable,
+    selectedAssignments,
+    actionCallback,
+}: ButtonWithDialogProps) {
+    const [showCreateConfirmation, setShowCreateConfirmation] = React.useState(
```
siefkenj: The code in all of these is very similar. You can just call the variable `showConfirmation` in each one.

frontend/src/views/admin/assignments/offer-button-with-dialog.tsx
```diff
@@ -0,0 +1,251 @@
...
+            >
+                Create Offer
+            </ActionButton>
+            <OfferConfirmationDialog
+                data={selectedAssignments}
+                visible={showCreateConfirmation}
+                setVisible={setShowCreateConfirmation}
+                callback={actionCallback}
+                title="Creating Multiple Offers"
+                body={`You are creating the following ${selectedAssignments.length} offers:`}
```
siefkenj: End with `.` instead of `:`

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -120,53 +113,35 @@ function OfferActionButtons(props) {
...
+                selectedAssignments={selectedAssignments}
+                actionCallback={createOffers}
```
siefkenj: -> `callback`

frontend/src/views/admin/assignments/offer-confirmation-dialog.tsx
```diff
@@ -107,11 +109,11 @@ export function MultiWithdrawOfferConfirmation(props: {
                 </Button>
                 <Button
                     onClick={() => {
-                        withdrawOffers();
+                        callback();
```
siefkenj: We should await the callback and set a spinner state while the callback is executing. The dialog shouldn't close until the callback is finished. You can look at some of the other dialogs for an example of this.

frontend/src/views/admin/assignments/offer-actions.js
Altair59: Why de we need to set it as an optional one?

siefkenj: It's just nicer that way.

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -61,39 +61,39 @@ function ConfirmWithDialogActionButtons(props) {
         setOfferForAssignmentRejected,
     } = props;
 
-    function createOffers() {
+    async function createOffers() {
         for (const assignment of selectedAssignments) {
-            offerForAssignmentCreate(assignment);
+            await offerForAssignmentCreate(assignment);
```
siefkenj: No. We do not want to wait one by one. We wan them to all happen in parallel. You can use `Promise.all` for this

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -114,34 +114,34 @@ function ConfirmWithDialogActionButtons(props) {
...
             <CreateOfferButtonWithDialog
-                actionButtonEnable={actionPermitted.canCreate}
+                actionButtonDisable={!actionPermitted.canCreate}
                 selectedAssignments={selectedAssignments}
-                actionCallback={createOffers}
+                callback={createOffers}
             />
             <WithdrawOfferButtonWithDialog
-                actionButtonEnable={actionPermitted.canWithdraw}
+                actionButtonDisable={!actionPermitted.canWithdraw}
```
siefkenj: -> `disabled`. There is no point in having `actionButton` be part of this prop name.

frontend/src/views/admin/assignments/offer-actions.js
```diff
@@ -59,49 +61,40 @@ function OfferActionButtons(props) {
...
+        return selectedAssignments.map((assignment) =>
+            offerForAssignmentCreate(assignment)
+        );
     }
+
     function withdrawOffers() {
-        for (const assignment of selectedAssignments) {
-            offerForAssignmentWithdraw(assignment);
-        }
+        return selectedAssignments.map((assignment) =>
```
siefkenj: You need to return a `Promise.all`. Otherwise `await withdrawOffers()` won't do anything.

frontend/src/views/admin/assignments/offer-button-with-dialog.tsx
```diff
@@ -0,0 +1,240 @@
...
+                title="Creating Multiple Offers"
+                body={`You are creating the following ${selectedAssignments.length} offers.`}
+                confirmation={`Create ${selectedAssignments.length} Offers`}
+            />
+        </React.Fragment>
+    );
+}
+
+export function WithdrawOfferButtonWithDialog({
+    disabled = true,
```
siefkenj: The default for a disabled attribute should always be `false`.

frontend/src/views/admin/assignments/offer-confirmation-dialog.tsx
```diff
@@ -59,13 +51,40 @@ function compareAssignment(a1: Assignment, a2: Assignment) {
...
+        setVisible,
+        callback,
+        title,
+        body,
+        confirmation,
+    } = props;
+
+    const [inProgress, setInProgress] = React.useState(false);
+
+    function executeCallback() {
```
siefkenj: Make this an async function and use `await`.

frontend/src/views/admin/assignments/offer-confirmation-dialog.tsx
```diff
@@ -59,13 +51,40 @@ function compareAssignment(a1: Assignment, a2: Assignment) {
...
+        setInProgress(true);
+        Promise.all(callback()).then(() => {
```
siefkenj: `Promise.all` should be moved to the `callback` as previously described.

siefkenj:
Good job!

PR #539, opened 2021-04-09
Implement posting-tests

### LINKED ISSUE #543: Fetch all applications associated with a posting
`/admin/postings/:posting_id/applications` need to be implemented and tested (todo in `posting-tests.js`)

### LINKED ISSUE #544: Disable creating a posting-position with a position associated with a different session than the posting
As title, creating a `posting-position` with a position from a different session than the posting should not be allowed.
(The related test case is skipped for now in `posting-test.js`)

### LINKED ISSUE #545: Deleting a posting that has associated posting-positions causes error
Deleting a posting should also delete all associated posting-positions. (The related test case is skipped in `posting-test.js`)

### LINKED ISSUE #546: Setting `custom_questions` in postings to Object type causes error when fetching
Setting `custom_questions` in postings to serializable Object type should be allowed and the exact same objects should be returned when fetching `custom_questions` from postings. (Related test case is skipped in `posting-tests.js`)

frankcbw:
Implementing `posting-tests`


frankcbw:
1. `"Fetch a survey for a posting"` has problems: 
   - error msg: `no implicit conversion of nil into Array`
2. `"Deleting a posting also deletes all associated posting_positions"` failed
   - deleting not allowed if the `posting.id` is referenced as a foreign key in the posting_position table


3. ` "Cannot create a posting_position with a position associated with a different session than the posting"` failed
   - It should have a status `error` but got `success` instead

frankcbw:
- Two tests are skipped due to backend issues. Should be unskipped after fixing issues #544 and #545 
- One test is left as `todo`. Should be implemented after fixing issue #543

frankcbw:
#546 is resolved

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,354 @@
...
+        resp = await apiPOST("/admin/postings", {
+            ...posting,
+            session_id: session.id,
+        });
+        expect(resp).toHaveStatus("success");
+        checkPropTypes(postingPropTypes, resp.payload);
+        expect(resp.payload.id).not.toBeNull();
+        Object.assign(posting, resp.payload);
+
+        const newPosting2 = {
```
Altair59: you may want to change the naming of `newPosting2`, as there's only one new `Posting` for this test

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,354 @@
...
-    );
-    it.todo(
-        "Can create two posting_positions with the same position_id for different postings"
-    );
+    // `/admin/postings/${posting.id}/applications` need to be implemented
+    it.todo("Fetch all applications associated with a posting");
+
+    it("Fetch a survey for a posting", async () => {
+        resp = await apiGET(`/admin/postings/${posting.id}/survey`);
+        expect(resp).toHaveStatus("success");
```
Altair59, line 221: you may want to also check `checkPropTypes` on the `survey` object, and probably by first defining a correct `survey ` prop type in `frontend\src\api\defs\prop-types.ts`

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,354 @@
...
+    });
+
+    it("Survey for a posting includes questions related to each PostingPosition", async () => {
+        resp = await apiGET(`/admin/postings/${posting.id}/survey`);
+        expect(resp).toHaveStatus("success");
+        let survey = resp.payload;
+        let postingPosQuestions = survey.pages.filter(
```
Altair59: you may want to use `const`...

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,354 @@
...
+            (p) => p.name === "preferences_page"
+        )[0].elements[0].rows;
+        resp = await apiGET(`/admin/postings/${posting.id}/posting_positions`);
+        expect(resp).toHaveStatus("success");
+        expect(resp.payload.length).toEqual(postingPosQuestions.length);
+    });
+
+    it.skip("Cannot create a posting_position with a position associated with a different session than the posting", async () => {
+        const newSession = await addSession(api);
```
Altair59: you may only need just one another new `session` globally, and re-use it on this test as well as previous tests involving different session

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,354 @@
...
+            position_id: position.id,
+        });
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET(`/admin/postings/${posting.id}/posting_positions`);
+        expect(
+            resp.payload.filter((p) => p.position_id === position.id).length
+        ).toEqual(0);
+    });
+
+    it.skip("Deleting a posting also deletes all associated posting_positions", async () => {
```
Altair59, line 371: why is this test with `skip`

frankcbw: Because there is an issue need to be fixed at the backend. Skipped for now.

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,356 @@
...
+            ...posting,
+            availability: 3,
+        });
+        expect(resp).toHaveStatus("error");
+    });
+
+    it("Fetch all postings for a session", async () => {
+        resp = await apiGET(`/admin/sessions/${session.id}/postings`);
+        expect(resp).toHaveStatus("success");
+        checkPropTypes(PropTypes.arrayOf(postingPropTypes), resp.payload);
```
siefkenj, line 142: You know what postings should be here. You should check for them explicitly. Also, does any session have more than one posting? You should make sure there's a session with more than one posting before checking.

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,356 @@
...
+    it("Survey for a posting includes questions related to each PostingPosition", async () => {
+        resp = await apiGET(`/admin/postings/${posting.id}/survey`);
+        expect(resp).toHaveStatus("success");
+        const survey = resp.payload;
+        const postingPosQuestions = survey.pages.filter(
+            (p) => p.name === "preferences_page"
+        )[0].elements[0].rows;
+        resp = await apiGET(`/admin/postings/${posting.id}/posting_positions`);
+        expect(resp).toHaveStatus("success");
+        expect(resp.payload.length).toEqual(postingPosQuestions.length);
```
siefkenj: This isn't a sufficient check...we really want to know if the `position_code` for each `posting_position` shows up. First we should have multiple posting positions in the session. Second, just a text search of the JSON should be fine. That wy there's not fragile unpacking of the survey format (which may change in the future)

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,356 @@
...
+            newPostingPos
+        );
+
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET(
+            `/admin/postings/${newPosting.id}/posting_positions`
+        );
+        // Verify a new posting_position is created successfully for the new posting
+        expect(
+            resp.payload.filter((p) => p.position_id === position.id).length
```
siefkenj: You can use `.find` instead of `.filter`

frontend/src/tests/posting-tests.js
```diff
@@ -1,39 +1,356 @@
...
+        resp = await apiPOST(
+            `/admin/postings/${posting.id}/posting_positions`,
+            newPostingPos
+        );
+        expect(resp).toHaveStatus("success");
+        resp = await apiGET(`/admin/postings/${posting.id}/posting_positions`);
+
+        // The previous POST request should not create a new posting_position and instead it upserts to an existing one
+        expect(
+            resp.payload.filter((p) => p.position_id === position.id).length
```
siefkenj, line 335: `.find` instead of `.filter`

frontend/src/tests/posting-tests.js
frankcbw: Yes. The pre-seeded `session` should contain two postings at this point.

frontend/src/tests/posting-tests.js
frankcbw: Here I want to check explicitly that there is only one `posting_position` that has `.position_id === position.id`. Using `.find` can only make sure there exists such one but not guarantee it's the only one, right?

siefkenj: That's right. `.find` will only find the first one

PR #541, opened 2021-04-09
Make React Table columns flex-span aligned

Accelsnow:
Make the table header and table content aligned, make them span all available widths while being resizeable.


frontend/src/components/filter-table/advanced-filter-table.tsx
```diff
@@ -185,31 +137,36 @@ export function AdvancedFilterTable({
...
-                                    <FixedSizeList
-                                        height={height}
-                                        itemCount={table.rows.length}
-                                        itemSize={30}
-                                        width={table.totalColumnsWidth}
-                                        ref={scrollRef}
-                                        style={{
-                                            overflow: "visible",
-                                        }}
+                    {table.rows.map((row) => {
```
siefkenj: Keep the `renderRow` as a its own function.

frontend/package.json
```diff
@@ -33,14 +32,15 @@
...
         "react-bootstrap-typeahead": "^5.1.4",
         "react-copy-to-clipboard": "^5.0.3",
-        "react-custom-scrollbars": "^4.2.1",
         "react-dom": "^17.0.1",
         "react-icons": "^4.1.0",
         "react-notification-system-redux": "^2.0.1",
         "react-redux": "^7.2.2",
         "react-router-dom": "^5.2.0",
         "react-scripts": "4.0.3",
+        "react-scrollbars-custom": "^4.0.25",
```
siefkenj, line 41: Why 4.0 and not 4.2?

frontend/src/views/admin/positions/position-list.js
```diff
@@ -316,7 +316,6 @@ export function ConnectedPositionsList({
         {
             Header: generateHeaderCell("Hours/Assignment"),
             accessor: "hours_per_assignment",
-            maxWidth: 64,
```
siefkenj: Why did you take away all these max widths?

Accelsnow: Many of these columns had very narrow width where the column name or the content itself was hardly visible (half of the word is cut off), so I removed them so they can fit their content.

siefkenj: For columns that are displaying a single number, you should keep them skinny. There's no point in making them wide, and it makes the data harder to understand when they're wide.

frontend/package.json
Accelsnow: This is a new library. The original library was not well maintained (last updated 2017) and it did not have enough functionality to achieve our goal. One of the main obstacles was that the old library, when creating the div for the scroll bars, does not give them a className, thus making it impossible to override their default styles or aligning them with the table. This library's latest version is [4.0.25](https://www.npmjs.com/package/react-scrollbars-custom), which includes functionalities like verticle/horizontal only, flexible scrollbar width and properly tagged div containers.

siefkenj: Ah. I didn't notice that it had a different name!

frontend/src/views/admin/positions/position-list.js
Accelsnow: roger! I will make the numeric columns a fixed width then.

Accelsnow:
During typescript transformation, I found that there are 3 places where typescript failed to recognize the `...args` argument passage (line [117](https://github.com/uoft-tapp/tapp/pull/541/files#diff-35150bf1e175655228882dbb23b4d21aa46e35a8f6a4de5ee234e3eb77c2321dR142), [141](https://github.com/uoft-tapp/tapp/pull/541/files#diff-35150bf1e175655228882dbb23b4d21aa46e35a8f6a4de5ee234e3eb77c2321dL139), [156](https://github.com/uoft-tapp/tapp/pull/541/files#diff-35150bf1e175655228882dbb23b4d21aa46e35a8f6a4de5ee234e3eb77c2321dL153)). Without ts-ignore, typescript reports `TS2556: Expected 1 arguments, but got 0 or more.` when calling the upsert functions. This might be caused by typescript not recognizing the underlying implementation of validatedApiDispatcher (which takes `...args` with type `ArgType extends unknown[]`. I have not yet found a "good" solution to this without using `// @ts-ignore`.

frontend/src/components/filter-table/advanced-filter-table.tsx
```diff
@@ -148,16 +151,15 @@ export function AdvancedFilterTable({
         [table]
     );
 
-    const handleScroll = React.useCallback(
-        ({ target }) => {
-            const { scrollTop } = target;
+    const handleScroll = ({ scrollTop, scrollLeft }: any) => {
```
siefkenj: Avoid `any` if you can.

frontend/src/components/filter-table/filter-table.css
```diff
@@ -74,58 +84,76 @@
...
 
 .thead,
 .hidden-row-warning {
-    border-bottom: 1px solid rgba(167, 182, 194, 0.4);
-    box-shadow: 0px 3px 8px -3px rgba(167, 182, 194, 0.5);
     user-select: none;
+    overflow-x: scroll;
 }
+
+::-webkit-scrollbar {
```
siefkenj: This concerns me...we shouldn't need prefixed css properties...

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -46,7 +47,7 @@ export function ApplicantCell(props) {
  * @param {*} { original }
  * @returns
  */
-export function StatusCell({ row }) {
+export function StatusCell({ row }: any) {
```
siefkenj: Look elsewhere in the code for examples. This can be typed.

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -12,18 +12,19 @@ import { FaSearch } from "react-icons/fa";
...
  * @param {*} props
  * @returns
  */
-export function ApplicantCell(props) {
+export function ApplicantCell(props: any) {
     const title = `Edit ${"" + props.column.Header}`;
     const { upsertApplicant, field, editable } = props;
     const applicant = props.row.original || props.row._original;
-    async function onChange(newVal) {
+    async function onChange(newVal: any) {
```
siefkenj: `any` -> `Applicant[keyof Applicant]`

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -12,18 +12,19 @@ import { FaSearch } from "react-icons/fa";
...
+import { generateHeaderCell } from "../../../components/table-utils";
 
 /**
  * A cell that renders editable applicant information
  *
  * @param {*} props
  * @returns
  */
-export function ApplicantCell(props) {
+export function ApplicantCell(props: any) {
```
siefkenj: No `any`. Look elsewhere in the code for examples of how to type this.

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -161,56 +165,55 @@ export function ConnectedOfferTable(props) {
...
             },
             {
-                Header: "Status",
+                Header: generateHeaderCell("Status"),
                 id: "status",
                 // We want items with no active offer to appear at the end of the list
                 // when sorted, so we set their accessor to null (the accessor is used by react table
                 // when sorting items).
-                accessor: (data) =>
+                accessor: (data: { active_offer_status: string }) =>
```
siefkenj, line 228: `data: Assignment`

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -132,11 +134,12 @@ export function ConnectedOfferTable(props) {
...
     const columns = React.useMemo(() => {
         // Bind an `ApplicantCell` to a particular field
-        function generateApplicantCell(field) {
-            return (props) => (
+        function generateApplicantCell(field: any) {
+            return (props: any) => (
                 <ApplicantCell
                     field={field}
-                    upsertApplicant={(...args) =>
+                    upsertApplicant={(...args: any) =>
```
siefkenj: We shouldn't need to destructure `args` here. We know the type of `upsertApplicant`.

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -132,11 +134,12 @@ export function ConnectedOfferTable(props) {
...
+                        // @ts-ignore
```
siefkenj: I've thought typescipt was giving an error for no reason many times. Almost always, I was wrong...

Any valid use of `// @ts-ignore` should be accompanied by documentation about why that's really the correct solution. In this case, I don't think it's right.

frontend/src/components/filter-table/filter-table.css
Accelsnow: this is deprecated from when I was trying to implement with native scrollbars. will be removed

frontend/src/views/admin/offertable/index.tsx
Accelsnow: This part I need some further investigation. The code works as-is and breaks when I remove any of the `...` (which makes sense, `...` makes many arguments, without `...` there is only one argument); therefore it is confusing why ts is complaining about this. It might be some underlying babysitting done by javascript to magically make incorrect code work as intended.

frontend/src/views/admin/offertable/index.tsx
Accelsnow:
```
TS2367: This condition will always return 'false' since the types '"accepted" | "rejected" | "withdrawn" | "provisional" | "pending" | "no_offer" | null' and '"No Contract"' have no overlap.
```
The Assignment type lists all possible values for active_offer_status; "No Contract" is not one of them, thus TS complains about the condition check below always being false

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -13,21 +13,39 @@ import { formatDownloadUrl, capitalize, formatDate } from "../../../libs/utils";
...
 /**
  * A cell that renders editable applicant information
  *
  * @param {*} props
  * @returns
  */
-export function ApplicantCell(props: any) {
+export function ApplicantCell(props: {
+    column: { Header: string };
+    row: { original: Assignment; _original?: Assignment };
```
siefkenj: We don't need `_original` any more. That was from react table 6, which is now completely gone :-)

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -13,21 +13,39 @@ import { formatDownloadUrl, capitalize, formatDate } from "../../../libs/utils";
...
+    upsertApplicant: (payload: {
+        [p: string]: Applicant[keyof Applicant];
+        id: number;
+    }) => Promise<RawApplicant>;
+    value: string;
+    field: string;
```
siefkenj: `keyof Applicant`, I suppose?

frontend/src/views/admin/offertable/index.tsx
```diff
@@ -13,21 +13,39 @@ import { formatDownloadUrl, capitalize, formatDate } from "../../../libs/utils";
...
  *
  * @param {*} props
  * @returns
  */
-export function ApplicantCell(props: any) {
+export function ApplicantCell(props: {
+    column: { Header: string };
+    row: { original: Assignment; _original?: Assignment };
+    upsertApplicant: (payload: {
+        [p: string]: Applicant[keyof Applicant];
```
siefkenj: `[p: keyof Applicant]: Applicant[p]`

frontend/src/views/admin/offertable/index.tsx
siefkenj: I see...does a new type get created for the data? If so, we should use that type.

frontend/src/views/admin/offertable/index.tsx
Accelsnow: I changed it to `Partial<Applicant>` as it is the payload type defined by the actual upsertApplicant function.

frontend/src/views/admin/offertable/index.tsx
Accelsnow: I am unsure about the exact logic here. Under what circumstances would "No Contract" be set as a status value for an assignment? I thought before that No Contract was a nice way of displaying null in the table to the users, and thus it should never be the actual status value in an Assignment object returned by the backend.

frontend/src/components/filter-table/advanced-filter-table.tsx
```diff
@@ -42,19 +41,22 @@ import "./filter-table.css";
...
-    setSelected,
-}: {
-    columns: any[];
-    data: any[];
-    filterable?: boolean | null;
-    selected?: any[];
-    setSelected?: Function;
-}) {
+export function AdvancedFilterTable(
+    this: any,
```
siefkenj, line 45: What does it mean to pass a `this` object? It makes me uncomfortable...

siefkenj:
Fixed via #615

PR #552, opened 2021-04-15
added recent activity date col for DDAH table

### LINKED ISSUE #549: Recent Activity Date missing from DDAH view
A recent activity column should be added to the DDAH table.

Altair59:
#549 added recent activity date col for DDAH table

### DIFF
--- a/frontend/src/views/admin/ddah-table/index.tsx
+++ b/frontend/src/views/admin/ddah-table/index.tsx
@@ -25,6 +25,7 @@ export interface RowData {
     first_name: string;
     total_hours: number | null;
     status: string | null;
+    recent_activity_date: string;
     issues: string | null;
     issue_code: "hours_mismatch" | "missing" | null;
 }
@@ -52,6 +53,35 @@ export function getReadableStatus(ddah: Pick<Ddah, "status">) {
         default:
             return "Unsent";
     }
+}
+
+function isIsoDateString(dateString: string | null) {
+    if (dateString == null) return false;
+    if (!/\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z/.test(dateString))
+        return false;
+    const date = new Date(dateString);
+    return date.toISOString() === dateString;
+}
+
+function getRecentActivityDate(ddah: Ddah) {
+    const validDates = [
+        ddah.accepted_date,
+        ddah.approved_date,
+        ddah.emailed_date,
+        ddah.revised_date,
+    ]
+        .filter((dateString) => {
+            return isIsoDateString(dateString);
+        })
+        // dateString || 0 is required to avoid
+        // typescript complaining about potentially passing null to Date constructor,
+        // which cannot happen (although it complains) due to previous filtering
+        .map((dateString) => new Date(dateString || 0).getTime());
+    if (validDates.length === 0) return "";
+    const recentActivityDate = new Date(Math.max.apply(null, validDates));
+    return `${recentActivityDate.toLocaleDateString(
+        "en-CA"
+    )} ${recentActivityDate.toLocaleTimeString("en-CA")}`;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


frontend/src/views/admin/ddah-table/index.tsx
```diff
@@ -54,6 +55,33 @@ export function getReadableStatus(ddah: Pick<Ddah, "status">) {
...
+
+function getRecentActivityDate(ddah: Ddah) {
+    const validDates = [
+        ddah.accepted_date,
+        ddah.approved_date,
+        ddah.emailed_date,
+        ddah.revised_date,
+    ]
+        .filter((dateString) => {
+            return isIsoDateString(dateString);
```
siefkenj: Just do `filter(date => date)`. There's no need to validate.

frontend/src/views/admin/ddah-table/index.tsx
```diff
@@ -54,6 +55,33 @@ export function getReadableStatus(ddah: Pick<Ddah, "status">) {
     }
 }
 
+function isIsoDateString(dateString: string | null) {
```
siefkenj: Get rid of this function.

frontend/src/views/admin/ddah-table/index.tsx
```diff
@@ -54,6 +55,33 @@ export function getReadableStatus(ddah: Pick<Ddah, "status">) {
...
+        ddah.revised_date,
+    ]
+        .filter((dateString) => {
+            return isIsoDateString(dateString);
+        })
+        // dateString || 0 is required to avoid
+        // typescript complaining about potentially passing null to Date constructor,
+        // which cannot happen (although it complains) due to previous filtering
+        .map((dateString) => new Date(dateString || 0).getTime());
+    if (validDates.length === 0) return "";
```
siefkenj: Never leave off `{  }` on an if statement.

frontend/src/views/admin/ddah-table/index.tsx
```diff
@@ -54,6 +55,33 @@ export function getReadableStatus(ddah: Pick<Ddah, "status">) {
...
+        ddah.emailed_date,
+        ddah.revised_date,
+    ]
+        .filter((dateString) => {
+            return isIsoDateString(dateString);
+        })
+        // dateString || 0 is required to avoid
+        // typescript complaining about potentially passing null to Date constructor,
+        // which cannot happen (although it complains) due to previous filtering
+        .map((dateString) => new Date(dateString || 0).getTime());
```
siefkenj: What? Dates can already be compared numerically without a call to `getTime`

frontend/src/views/admin/ddah-table/index.tsx
```diff
@@ -54,6 +55,33 @@ export function getReadableStatus(ddah: Pick<Ddah, "status">) {
...
+    if (validDates.length === 0) return "";
+    const recentActivityDate = new Date(Math.max.apply(null, validDates));
```
siefkenj: How about `new Date(Math.max(...[<all the dates>].filter(date => date).map(date => new Date(date))))`

frontend/src/views/admin/ddah-table/index.tsx
```diff
@@ -25,6 +25,7 @@ export interface RowData {
     first_name: string;
     total_hours: number | null;
     status: string | null;
+    recent_activity_date: string;
```
siefkenj: Should be nullable if there is no recent activity.

frontend/src/views/admin/ddah-table/index.tsx
Altair59: But Typescript will complain that `Argument of type 'Date' is not assignable to parameter of type 'number'` when using `Math.max`, which is why I convert `Date`s to numeric values using `getTime`

siefkenj: In that case, you can cast them all to numbers (use `+` as a prefix to cast), or you can retype `Max`. E.g.,

```
let x = new Date();
let y = new Date();

let z =(Math.max as (...values: any[]) => number)(x,y);
```

Issue #551, opened 2021-04-15
Update DDAH to match the new collective agreement


siefkenj:
The new collective agreement requires the DDAH be divided into the categories
  * meetings
  * training
  * preparation
  * contact time
  * marking/grading
  * other duties

PR #557, opened 2021-04-19
Update Ddahs for new Collective Agreement

siefkenj:
The new collective agreement mandates that DDAH duties are grouped into 6 specific categories. This PR updates the rendering and storage of DDAH duties to be prefixed with `<category>:` and then rendered appropriately.


siefkenj:
Fixes #551

PR #558, opened 2021-04-23
added mock data JSON and  button for mock data loading

frankcbw:
- Added mock data files in JSON format for 1000 applicants, 100 positions, and 1000 assignments under `/frontend/src/mock_data`. 
- Added a button to set up a mock session and to load the JSON files in order.


frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+            contractTemplates.push(currContractTemplate);
+            setNow(50);
+            // upsert any missing instructors
+            setStage("Instructors");
+            for (let instructor of mockInstructors) {
+                if (!instructors.some((i) => i.utorid === instructor.utorid)) {
+                    const newInstructor = await dispatch(
+                        upsertInstructor(instructor)
+                    );
+                    instructors.push(newInstructor);
```
siefkenj: This shouldn't be needed. When the instructors update the selector should automatically update.

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+                    data: mockPositionData,
+                },
+                positionSchema
+            ).map((p) =>
+                prepareFull.position(p, {
+                    instructors,
+                    contractTemplates,
+                })
+            );
+            for (let p of data) {
```
siefkenj: `let`->`const`

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+
+            setStage("Contract Template");
+            const currContractTemplate = await dispatch(
+                upsertContractTemplate(mockContractTemplate)
+            );
+            contractTemplates.push(currContractTemplate);
+            setNow(50);
+            // upsert any missing instructors
+            setStage("Instructors");
+            for (let instructor of mockInstructors) {
```
siefkenj: Use `const` unless you're going to mutate the variable.

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+                })
+            );
+            for (let a of data) {
+                await dispatch(upsertAssignment(a));
+                count++;
+                setNow(Math.round((count / total) * 100));
+            }
+        } catch (error) {
+            console.error(error);
+            return error;
```
siefkenj: Why are you returning the errors?

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+            }
+            setInProgress(true);
+            await setupMockSession();
+            await setupMockPositions();
+            await setupMockApplicants();
+            await setupMockAssignments();
+            setInProgress(false);
+        } catch (error) {
+            setInProgress(false);
+            return;
```
siefkenj: No need for a return.

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+                setDisable(true);
+                return;
+            }
+            setInProgress(true);
+            await setupMockSession();
+            await setupMockPositions();
+            await setupMockApplicants();
+            await setupMockAssignments();
+            setInProgress(false);
+        } catch (error) {
```
siefkenj, line 194: A `finally` would be a better place to `setInProgress(false)`

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+            setInProgress(false);
+            return;
+        }
+    }
+
+    return (
+        <span
+            title={"One-click loads all mock data JSON files"}
```
siefkenj: String attributes can be set directly

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+            className="mock-button"
+        >
+            <Button onClick={clickHandler}>Load Mock</Button>
+            <Modal show={disable} onHide={() => setDisable(false)} size="md">
+                <Modal.Header closeButton>
+                    <Modal.Title>{"Mock session already exists"}</Modal.Title>
```
siefkenj: You don't need a string here. You can type directly in jsx.

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+                </Modal.Header>
+                <Modal.Body>
+                    {"Please delete the mock session first before reloading."}
```
siefkenj: What does this mean?

frontend/src/views/dev_frame/load-mock-data.js
```diff
@@ -0,0 +1,227 @@
...
+ * This component only renders when `process.env.REACT_APP_DEV_FEATURES` is truthy.
+ *
+ * @export
+ * @returns {React.ElementType}
+ */
+function LoadMockButton() {
+    const [disable, setDisable] = React.useState(false);
+    const [inProgress, setInProgress] = React.useState(false);
+    const [stage, setStage] = React.useState("");
+    const [now, setNow] = React.useState(0);
```
siefkenj: `now` -> `progress`

frontend/src/views/dev_frame/load-mock-data.js
frankcbw: Yes that's what I originally thought too... but I checked and it did not update `instructors`

frontend/src/views/dev_frame/load-mock-data.js
frankcbw: A simple `console.error(error)` in the `catch`-block and moving `setInProgress(false);` into a `finally`-block ?

frontend/src/views/dev_frame/load-mock-data.js
frankcbw: If users have clicked the mock button, there will be a mock session already... so I set up a modal to remind that they have to delete the mock session (manually go to the session table and delete it) before they can click the button again...

frontend/src/views/dev_frame/load-mock-data.js
frankcbw: This is supposed to be the tooltip content, as I tried to be consistent with the other two components, `ActiveUserButton` and `ToggleMockApi`.

frankcbw: The tooltip content might need to be changed tho

frontend/src/views/dev_frame/load-mock-data.js
siefkenj: Let's randomly generate the session name based on a timestamp. That way there won't be conflicts.

Actually, it'd be really nice to have a "use existing session"/"new session" dropdown to decide where the data goes.

frontend/src/views/dev_frame/load-mock-data.js
siefkenj: I suspect that's a bug, then...

frontend/src/views/dev_frame/load-mock-data.js
frankcbw: Adding to existing sessions directly using the mock data may cause conflicts because it doesn't check diff/changed compared with existing data so that there could be duplicate position codes, assignments, and even applicants... But I think technically we could choose to compute diff first if using an existing session...

siefkenj: No need. If it errors, it errors. This is a debug tools. We're going to assume the devs can understand the errors.

frontend/src/views/dev_frame/load-mock-data.js
siefkenj: When it changes, then put the brackets back.

siefkenj:
Merging for now. We'll fix remaining issues later.

Issue #474, opened 2021-01-15
[backend] potential typo in applicantion config


Accelsnow:
In line 58 of `backend/config/application.rb`, the default_url_options host is set to `'${confit.base_url}'`. This may potentially be a typo that will cause a bug (it likely should be `'${config.base_url}'`). There is also currently no test case that detects this bug.

PR #571, opened 2021-05-06
Fix skipped posting tests

siefkenj:
Two postings tests were skipped because of errors in the backend. The errors are fixed and the tests are enabled.


siefkenj:
Fixes #474

Issue #590, opened 2021-05-26
Completing Application Tests 


ghost:
Complete current TODOs in frontend's application-tests.js

Public route tests. 
- [x] Get survey.js posting data through public route
- [x] Survey.js posting data is pre-filled based on prior applicant/application data
- [x] Can submit survey.js data via the public postings route
- [x] When submitting survey.js data an applicant and application are automatically created they don't exist
- [x] When submitting survey.js data an applicant and application are updated if they already exist
- [ ] Even if a different utorid is submitted via survey.js data the active_user's utorid is used
- [x] When submitting survey.js data cannot add a position_preference for a position not listed in the posting
- [ ] When submitting survey.js data attached files are stored on disk rather than as base64 strings in the database
- [ ] Can submit a jpg/png file as a 'transcript' for an application; the resulting file can be retrieved
- [ ] Can submit a pdf file as a 'transcript' for an application; the resulting file can be retrieved
- [ ] Can submit and retrieve multiple files as a 'transcript' for an application
- [ ] Can submit and retrieve attachments for a new application
- [ ] Can submit and retrieve attachments for an updated application

Admin route tests.
- [ ] Create an application
- [ ] Update an application

ghost:
Update from Zoom meeting with @mertkaya1033.

Upon prior completion of database seeder for posting / response data (which both people can utilize), Mert will work on test cases when submitting survey.js data: "automatic creation of applicant/application data if DNE", "automatic update of applicant/applicant data if already exist". I'll be working on "verifying active_user's utorid is used" and "disallowing addtion of position_preference for a position not listed in posting".

PR #591, opened 2021-05-26
test for submiting survey.js data

ghost:
Reference to issue #590.


backend/Gemfile
```diff
@@ -7,7 +7,7 @@ ruby '2.6.6'
 
 gem 'rails', '~> 6.1.1'
 gem 'pg', '>= 0.18', '< 2.0'
-gem 'puma', '~> 3.12'
+gem 'puma', '~> 4.3'
```
alex-kozin: Are these updates necessary for this PR? I think we shouldn't need to update it here? Let me know otherwise!

frontend/src/tests/application-tests.js
```diff
@@ -23,9 +114,6 @@ export function applicationsTests({ apiGET, apiPOST }) {
         it.todo(
             "Even if a different utorid is submitted via survey.js data the active_user's utorid is used"
         );
-        it.todo(
-            "When submitting survey.js data cannot add a position_preference for a position not listed in the posting"
-        );
```
alex-kozin, line 28: Sorry, what was the reason we deleted this one?

frontend/src/tests/utils.js
```diff
@@ -174,8 +175,8 @@ function _ensurePath(path) {
  * @param {string} url The api un-prefixed url route (e.g. "/sessions")
  * @returns
  */
-export async function apiGET(url) {
-    url = API_URL + _ensurePath(url);
+export async function apiGET(url, omitPrefix = false) {
+    url = omitPrefix ? URL + _ensurePath(url) : API_URL + _ensurePath(url);
```
alex-kozin: I think this have been changed on master too, did you rebase the latest one?

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,110 @@
...
+    let resp;
+    let admin;
+    let surveyData;
+
+    const taOnlyUser = {
+        utorid: "matthewc",
+        roles: ["ta"],
+    };
+
+    const postingData = {
```
alex-kozin, line 22: Do we expect this to be reused in other tests too? If so, should we put it in the seeded data?

Also, if this is used in other tests - how is it different from `posting` global variable above?

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,110 @@
...
+        it("Can submit survey.js data via the public postings route", async () => {
+            //Create a new posting
+            resp = await apiPOST(
+                `/admin/sessions/${session.id}/postings`,
+                postingData
+            );
+            expect(resp).toHaveStatus("success");
+            Object.assign(posting, resp.payload);
+            checkPropTypes(postingPropTypes, posting);
+            expect(posting.id).not.toBeNull();
```
alex-kozin, line 56: if id is not returned it would be `undefined` I think? Maybe `expect(posting.id).toBeDefined()` would work better, what do you think?

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,110 @@
...
+                },
+            };
+
+            //Submit survey.js data
+            resp = await apiPOST(
+                `/public/postings/${posting.url_token}/submit`,
+                surveyData,
+                true
+            );
+            expect(resp).toHaveStatus("success");
```
alex-kozin, line 101: Do we want to also `expect(resp.payload).toStrictEqual()` the expected payload?

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,110 @@
...
+            Object.assign(posting, resp.payload);
+            checkPropTypes(postingPropTypes, posting);
+            expect(posting.id).not.toBeNull();
+
+            //Set position for posting
+            resp = await apiPOST(
+                `/admin/postings/${posting.id}/posting_positions`,
+                { hours: 20, num_positions: 2, position_id: position.id }
+            );
+            expect(resp).toHaveStatus("success");
```
alex-kozin: If this one is the same for the whole test set, should it be in the `beforeAll` or the seeded data?

alex-kozin (commented):
Just some questions :sweat_smile: Otherwise great job! :rocket:

frontend/src/tests/application-tests.js
ghost: The seeder's to come in a different PR as I can't combine two issues in one PR.
Right now, upon submission of postingData, we store as a global.

frontend/src/tests/application-tests.js
ghost: According to Jason, verification like so at this stage doesn't confirm anything meaningful, and since application data is verified in another test case, I'm  just confirming success here.

frontend/src/tests/utils.js
ghost: From what I see, this is the latest on master as well. What change are you referring to?

frontend/src/tests/application-tests.js
ghost: `not.toBeNull()` is the method used for `id`s in all other test suites.

frontend/src/tests/application-tests.js
ghost: Hmm thought I pushed back in. Did it again.

backend/Gemfile
ghost: The master already has these updated, so when getting the changes from master, it automatically changed the versions.

frontend/src/tests/application-tests.js
ghost: As said before, I'm following up with a PR for seeder as I can't combine two issues in one. This is temporary.

backend/Gemfile
alex-kozin: Oh, hmmm :thinking: I think it should not be shown in your PR if it is on master? Did you fetch the upstream and rebase from your fork?

frontend/src/tests/utils.js
alex-kozin: (please see above :slightly_smiling_face:  )

frontend/src/tests/application-tests.js
alex-kozin: @developerjinchun I understand it's used elsewhere, but do you think it will catch the case where id does not exist?

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,110 @@
...
+
+            //Create and switch to a ta only user
+            resp = await apiPOST("/debug/users", taOnlyUser);
+            expect(resp).toHaveStatus("success");
```
mertkaya1033: Even though you are creating a user with only TA permissions, you are not setting the active user to the taOnlyUser. Because this test is for the posting route, we should be testing if someone with only ta permissions can submit to a posting. This can be done by making a POST request to "/debug/active_user".

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,110 @@
...
+            };
+
+            //Submit survey.js data
+            resp = await apiPOST(
+                `/public/postings/${posting.url_token}/submit`,
+                surveyData,
+                true
+            );
+            expect(resp).toHaveStatus("success");
+
```
mertkaya1033, line 102: After you are done with the tests, you should make sure to change active_user to its default "admin" user.

siefkenj:
You need to pull from master so that `Gemfile` and `package-lock.json` don't appear in your diff.

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,116 @@
...
 import { databaseSeeder } from "./setup";
 
 export function applicationsTests({ apiGET, apiPOST }) {
-    beforeAll(async () => {
-        await databaseSeeder.seed({ apiGET, apiPOST });
-    }, 30000);
+    let session;
+    let position;
+    let posting = {};
+    let resp;
```
siefkenj: Define this inside each function that needs it. There are potential race conditions otherwise.

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,116 @@
...
     // These tests set data through the `/public/postings` route,
     // but read data through the `/api/v1/admin` route.
     describe("Public route tests", () => {
+        beforeAll(async () => {
+            await databaseSeeder.seed({ apiGET, apiPOST });
+            session = databaseSeeder.seededData.session;
+            position = databaseSeeder.seededData.position;
+            resp = await apiGET(`/debug/active_user`);
+            expect(resp).toHaveStatus("success");
+            admin = {
```
siefkenj: -> `defaultUser` or `adminUser`

siefkenj: Also, why didn't you just do `adminUser = resp.payload`?

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,116 @@
...
+        });
+
         it.todo("Get survey.js posting data through public route");
         it.todo(
             "Survey.js posting data is pre-filled based on prior applicant/application data"
         );
-        it.todo("Can submit survey.js data via the public postings route");
+
+        it("Can submit survey.js data via the public postings route", async () => {
+            //Create a new posting
```
siefkenj: Spaces before you start a comment.

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,116 @@
...
+            );
+            expect(resp).toHaveStatus("success");
+            Object.assign(posting, resp.payload);
+            checkPropTypes(postingPropTypes, posting);
+            expect(posting.id).not.toBeNull();
+
+            //Set position for posting
+            resp = await apiPOST(
+                `/admin/postings/${posting.id}/posting_positions`,
+                { hours: 20, num_positions: 2, position_id: position.id }
```
siefkenj: Setting these fields is fine. You could also leave them out, because they have default values.

frontend/src/tests/application-tests.js
```diff
@@ -1,19 +1,116 @@
...
+            );
+            expect(resp).toHaveStatus("success");
+
+            //Create and switch to a ta only user
+            resp = await apiPOST("/debug/users", taOnlyUser);
```
siefkenj, line 66: I think Mert's PR includes utility functions for doing this switch. Are you planning on using those when his PR is merged?

frontend/src/tests/application-tests.js
ghost: Currently `session` and `position` are used as global variables and defined in `beforeAll` through seeders. To clarify, are you saying they should be defined from calling the seeder in each case that needs it?

frontend/src/tests/application-tests.js
ghost: They can be switched if needed.

siefkenj:
Great job!

PR #594, opened 2021-05-26
omitPrefix in apiGET and apiPOST functions

mertkaya1033:
This omitting prefix functionality is required when testing public routes that are used in the frontend and do not include the API_URL prefix.


ghost (commented):
Seems like `omitPrefix` is by default set to `false`. So just to confirm, the function's intended default behaviour (without the optional parameter) is that it still shouldn't omit the prefix?

mertkaya1033:
Yes, since all of the tests include the API_URL (i.e. /api/v1) prefix, omitting this by default by setting it to true would cause conflicts with previous tests.

PR #596, opened 2021-05-28
[hotfix] Test api omit prefix

mertkaya1033:
There is a minor bug in the omitPrefix functionality where the whole URL prefix (including "http://backend:3000") was being omitted. This is not the expected behaviour as the apiGET and apiPOST will fail with ECONNREFUSED error when `omitPrefix === true`.

### DIFF
--- a/frontend/src/tests/utils.js
+++ b/frontend/src/tests/utils.js
@@ -175,7 +175,7 @@ function _ensurePath(path) {
  * @returns
  */
 export async function apiGET(url, omitPrefix = false) {
-    url = omitPrefix ? _ensurePath(url) : API_URL + _ensurePath(url);
+    url = omitPrefix ? API_URL + _ensurePath(url) : _ensurePath(url);
     let resp = null;
     try {
         resp = await axios.get(url);
@@ -202,7 +202,7 @@ export async function apiGET(url, omitPrefix = false) {
  * @returns
  */
 export async function apiPOST(url, body = {}, omitPrefix = false) {
-    url = omitPrefix ? _ensurePath(url) : API_URL + _ensurePath(url);
+    url = omitPrefix ? API_URL + _ensurePath(url) : _ensurePath(url);
     let resp = null;
     try {
         resp = await axios.post(url, body);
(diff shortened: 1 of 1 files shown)


frontend/src/tests/utils.js
```diff
@@ -160,6 +160,7 @@ export { expect, test, it, describe, beforeAll };
 
 /** URL prefix for making API calls from inside a docker image */
 export const API_URL = "http://backend:3000/api/v1";
```
alex-kozin, line 162: Couldn't find why `API_URL` is exported, it seems to not be used outside of this file?

frontend/src/tests/utils.js
```diff
@@ -175,7 +176,7 @@ function _ensurePath(path) {
  * @returns
  */
 export async function apiGET(url, omitPrefix = false) {
-    url = omitPrefix ? _ensurePath(url) : API_URL + _ensurePath(url);
+    url = omitPrefix ? URL + _ensurePath(url) : API_URL + _ensurePath(url);
```
alex-kozin, line 179: Good catch! What do you think of something like

```
const BASE_URL = "http://backend:3000";
....
export async function apiGET(url, omitPrefix = false) {
    url = `${BASE_URL}${omitPrefix ? '' : '/api/v1'}${_ensurePath(url)}`;
    ...
```

Or alternatively:

```
const BASE_URL = "http://backend:3000";
const API_ROUTE = '/api/v1';
....
export async function apiGET(url, omitPrefix = false) {
    let urlParts = [BASE_URL];
    if (!omitPrefix) {
        urlParts.push(API_ROUTE);
    }
    urlParts.push(_ensurePath(url));
    url = urlParts.concat();
    ...
```

alex-kozin: I prefer the first one, but even better would be to have `addPrefix`, instead of `omitPrefix` (omit = false is double negation), but that's a style choice that is consistent with the real method after all.

siefkenj: Look into the standard `URL` object before you reinvent the wheel too much ;-)

alex-kozin: I forgot of its existence :joy: Something like `url = new URL(omitPrefix ? '' : 'api/v1', 'https://localhost:8000').href;`? Looks a lot better indeed!

PR #598, opened 2021-05-29
Mock data button

frankcbw:
New Updates:
- Can upload data by batches


frontend/src/mock_data/contract_template.json
```diff
@@ -0,0 +1,4 @@
+{
```
siefkenj: Why are you adding all this singleton data? We have pluralized versions of all these files already in the repo.

frankcbw: this was the old original JSON file... will be removed in the next commit

frankcbw: that's weird because all those singleton data were added when I first started this PR... not from my recent commit

siefkenj: Maybe you need to merge upstream again?

frankcbw: I think it's because I didn't create a new branch from master. Instead, I reopened the previously merged PR... Should I close this one and create a new PR?

frankcbw: I think it's the renaming issue on GIT again... I suppose the mock data JSON files have been renamed to their pluralized forms. Then git added them as untracked files, instead of replacing the original singleton files

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -123,6 +122,27 @@ export function SeedDataMenu({
         }
     }, [dropdownVisible, fetchSessions]);
 
+    async function progressDispatch(
```
siefkenj, line 123: We need a comment explaining what this function does

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -123,6 +122,27 @@ export function SeedDataMenu({
...
+        dispatchAction: Function,
+        dispatchArray: any[],
```
siefkenj, line 125: We can use a generic for `dispatchAction` and then `Parameters<...>[]` for `dispatchArray`.


Here's a type for it courtesy of stack overflow

```typescript
// Modified from https://stackoverflow.com/questions/51851677/how-to-get-argument-types-from-function-in-typescript/51851844
type FirstArgumentType<F extends Function> = F extends (...args: infer A) => any ? A[0] : never;

async function progressDispatch<F extends Function> (dispatchAction: F, dispatchArray: FirstArgumentType<F>[]) { ... }
```

In general, if you have an `any` type, it means you've given up on typescript and so you don't get any of its benefits.

siefkenj:
Closing for now, since this PR isn't being worked on.

Issue #597, opened 2021-05-28
Instructor Permission Tests are not implemented


alex-kozin:
We need to write the following tests:

- [x] fetch Ddahs
- [x] fetch Ddahs a position associated with self
- [x] fetch Ddahs an assignment associated with self
- [x] cannot fetch Ddahs for assignment not associated with self
- [x] create a Ddah for an assignment associated with self
- [x] update a Ddah for an assignment associated with self
- [x] cannot set approved_date/accepted_date/revised_date/emailed_ate/signature for a Ddah associated with self
- [x] cannot create a Ddah for an assignment not associated with self
- [x] cannot update a Ddah for an assignment not associated with self

PR #599, opened 2021-05-31
Add more tests for instructor permissions to fetch ddahs

### LINKED ISSUE #526: Instructor permission test should also test correct data visibility
PR #477 is getting bloated; therefore this section will not be implemented in this PR. After this PR is merged a new PR will be created to implement tests for data visibility.

### LINKED ISSUE #603: Tests: revisit test decisions for instructor permissions once the UI is more mature
There were several instructor permission tests deleted. They were testing querying Ddahs by assignment id and by position id. It was decided best to delete these tests until the UI is more developed to see if they will be useful at all.

The reasoning was that the admin view currently fetches everything on a per-session basis and later on additional filters may be applied. If instructor view follows the same pattern - we would have no need to query Ddahs on a per-position or per-assignment basis.

The goal of this issue is to revisit this decision when the instructor UI is built out to see if we need these tests (and corresponding functionality) or not.

alex-kozin:
More tests for fetching DDAHs for instructor. Related to issue #597

### DIFF
--- a/frontend/src/tests/instructor-permission-test.js
+++ b/frontend/src/tests/instructor-permission-test.js
@@ -476,28 +476,12 @@ export function instructorsPermissionTests(api) {
     });
 
     describe("Ddah permissions", () => {
-        let position;
-        let instructor;
+        let position = databaseSeeder.seededData.positions[3];
         let assignment;
         let assignmentInstructorCantAccess;
 
         beforeAll(async () => {
             await restoreDefaultUser();
-
-            // Get our instructor
-            let resp = await apiGET(`/admin/instructors`);
-            expect(resp).toHaveStatus("success");
-            instructor = resp.payload.find(
-                (instructor) => instructor.utorid === instructorUser.utorid
-            );
-            expect(instructor).toBeDefined();
-
-            // Fetch the position related to our instructor
-            resp = await apiGET(`/admin/sessions/${session.id}/positions`);
-            position = resp.payload.find((position) =>
-                position.instructor_ids.find((id) => id === instructor.id)
-            );
-            expect(position).toBeDefined();
 
             // Create another assignment (with no DDAH)
             // for position associated with our instructor
@@ -507,29 +491,14 @@ export function instructorsPermissionTests(api) {
                 start_date: "2019-09-02T00:00:00.000Z",
                 end_date: "2019-12-31T00:00:00.000Z",
             };
-            resp = await apiPOST("/admin/assignments", newAssignment);
+            let resp = await apiPOST("/admin/assignments", newAssignment);
             expect(resp).toHaveStatus("success");
             expect(resp.payload).toBeDefined();
             assignment = resp.payload;
 
-            // Fetch the position related to other instructor
-            resp = await apiGET(`/admin/instructors`);
-            expect(resp).toHaveStatus('success');
-            expect(resp.payload).toBeDefined();
-            const otherInstructor = resp.payload.find(
-                (instructor) =>
-                    instructor.utorid ===
-                    databaseSeeder.seededData.instructors[4].utorid
-            );
-            expect(otherInstructor).toBeDefined();
-            resp = await apiGET(`/admin/sessions/${session.id}/positions`);
-            const otherPosition = resp.payload.find((position) =>
-                position.instructor_ids.find((id) => id === otherInstructor.id)
-            );
-            expect(otherPosition).toBeDefined();
-
             // Create another assignment (with no DDAH)
             // for position associated with other instructor
+            const otherPosition = databaseSeeder.seededData.positions[4];
             const otherAssignment = {
                 note: "",
                 applicant_id: databaseSeeder.seededData.applicant.id,
@@ -621,7 +590,7 @@ export function instructorsPermissionTests(api) {
             expect(resp.payload).toEqual(expect.objectContaining(updatedDdah));
         });
 
-        // This test indeed fails and allows to set these fields 0w0
+        // This test indeed fails and allows to set these fields, see Issue #608
         it.skip("cannot set approved_date/accepted_date/revised_date/emailed_date/signature for a Ddah associated with self", async () => {
             // Fetch the DDAH related to instructor's assignment
             let resp = await apiGET(`/instructor/sessions/${session.id}/ddahs`);
@@ -656,12 +625,13 @@ export function instructorsPermissionTests(api) {
             expect(resp.payload).toEqual(expect.objectContaining(originalDdah));
 
             resp = await apiGET(`/instructor/sessions/${session.id}/ddahs`);
+            // still a success, but does not change the underlying data
             expect(resp).toHaveStatus("success");
             const newDdah = resp.payload.find(
                 (ddah) => ddah.assignment_id === assignment.id
             );
             expect(newDdah).toEqual(expect.objectContaining(originalDdah));
-        }); // still a success, but does not change the underlying data
+        });
 
         it("cannot create a Ddah for an assignment not associated with self", async () => {
             const ddahToCreate = {
--- a/frontend/src/tests/instructor-permission-test.js
+++ b/frontend/src/tests/instructor-permission-test.js
@@ -1,11 +1,4 @@
-import {
-    it,
-    expect,
-    beforeAll,
-    describe,
-    checkPropTypes,
-    errorPropTypes,
-} from "./utils";
+import { it, expect, beforeAll, checkPropTypes, errorPropTypes } from "./utils";
 import { databaseSeeder } from "./setup";
 
 /**
@@ -22,7 +15,6 @@ export function instructorsPermissionTests(api) {
     // eslint-disable-next-line
     const { apiGET, apiPOST } = api;
     let session = null;
-    let assignment;
     let instructorUser;
     let defaultUser;
     let existingContractTemplateId;
@@ -52,56 +44,6 @@ export function instructorsPermissionTests(api) {
             defaultUser
         );
         expect(respSwitchBackUser).toHaveStatus("success");
-    }
-
-    /**
-     * Updates a position to include instructor with `instructorId` and
-     * creates a DDAH for one of the assignments realted to that
-     * instructor.
-     *
-     * @param instructorId: int - the unique Id of an instructor
-     *
-     * @returns {Promise<Ddah>}
-     */
-    async function createDdahWithFixedDuties(instructorId) {
-        // We first need to update position to include our instructor
-        const existingPosition = databaseSeeder.seededData.positions[0];
-        const positionWithInstructor = {
-            ...existingPosition,
-            instructor_ids: [...existingPosition.instructor_ids, instructorId],
-        };
-        let resp = await apiPOST(`/admin/positions`, positionWithInstructor);
-        expect(resp).toHaveStatus("success");
-
-        // We then proceed to create a DDAH for that position
-        resp = await apiGET(`/instructor/sessions/${session.id}/assignments`);
-        expect(resp).toHaveStatus("success");
-        expect(resp.payload.length).toBeGreaterThan(0);
-        assignment = resp.payload[0];
-        const newDdah = {
-            assignment_id: assignment.id,
-            duties: [
-                {
-                    order: 2,
-                    hours: 25,
-                    description: "marking:Marking the midterm",
-                },
-                {
-                    order: 1,
-                    hours: 4,
-                    description: "training:Initial training",
-                },
-                {
-                    order: 3,
-                    hours: 40,
-                    description: "contact:Running tutorials",
-                },
-            ],
-        };
-
-        resp = await apiPOST(`/admin/ddahs`, newDdah);
-        expect(resp).toHaveStatus("success");
-        return resp.payload;
     }
 
     beforeAll(async () => {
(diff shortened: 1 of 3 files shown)


frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+            const newAssignment = {
+                note: "",
+                applicant_id: databaseSeeder.seededData.applicant.id,
+                position_id: position.id,
+                start_date: "2019-09-02T00:00:00.000Z",
+                end_date: "2019-12-31T00:00:00.000Z",
+            };
+            resp = await apiPOST("/admin/assignments", newAssignment);
+            expect(resp).toHaveStatus("success");
+            assignment = resp.payload;
```
mertkaya1033, line 497: Maybe you should add a check if the assignment is defined since you are assuming that the assignment is defined in the future. It is fine without it but you cannot really over-test it :)

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+
+            // Fetch the position related to other instructor
+            resp = await apiGET(`/admin/instructors`);
+            const otherInstructor = resp.payload.find(
```
mertkaya1033: Again, maybe you should add the check if resp is a success since you cannot really over-test these kinds of stuff.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
-        resp = await apiGET(`/admin/instructors`);
-        expect(resp).toHaveStatus("success");
-        const instructorId = resp.payload.find(
-            (instructor) => instructor.utorid === instructorUser.utorid
-        )?.id;
-        expect(instructorId).toBeDefined();
-        const newDDAH = await createDdahWithFixedDuties(instructorId);
+        beforeEach(async () => {
+            await switchToInstructorOnlyUser();
+        });
```
mertkaya1033, line 516: I like how you are making sure your active user is instructor only before each test!

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
-    });
+        it("fetch Ddahs associated with self only", async () => {
+            // Our instructor should only be able to fetch ddahs associated with them
+            const resp = await apiGET(
+                `/instructor/sessions/${session.id}/ddahs`
+            );
+            expect(resp).toHaveStatus("success");
+
+            // Only one DDAH is intially seeded for this instructor
+            expect(resp.payload).toHaveLength(1);
```
mertkaya1033, line 526: Even though this implementation should be fine, I think you should not assume that only one DDAH is initially seeded for the instructor unless you are seeding the instructor within the test. This is because someone may need to seed more DDAH for this particular instructor in the future and might run into problems with this test.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+                    {
+                        order: 1,
+                        hours: 20,
+                        description: "other:Facilitating workshops",
+                    },
+                ],
+            };
+            let resp = await apiPOST(`/instructor/ddahs`, ddahToCreate);
+            expect(resp).toHaveStatus("error");
+            expect(resp.message).toStrictEqual(
```
mertkaya1033: This message might change in the future as even a change on cases will make this test fail. Just expecting it to respond with error should be fine in my opinion.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+                    {
+                        order: 1,
+                        hours: 20,
+                        description: "other:Facilitating workshops",
+                    },
+                ],
+            };
+            resp = await apiPOST(`/instructor/ddahs`, ddahToUpdate);
+            expect(resp).toHaveStatus("error");
+            expect(resp.message).toStrictEqual(
```
mertkaya1033: See the comment above.

mertkaya1033 (changes requested):
This is a great PR only with a single minor issue.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -238,10 +189,10 @@ export function instructorsPermissionTests(api) {
 
     it("can't update instructors except for self (i.e. active user)", async () => {
         // we use the default user's admin privilege here to fetch an unrelated instructor
-        let respFetchInst = await apiGET(`/instructor/instructors`);
+        const respFetchInst = await apiGET(`/instructor/instructors`);
```
siefkenj: `let resp = await ...` and reuse `resp` unless you need to keep the response object around for some reason...

frontend/src/tests/instructor-permission-test.js
```diff
@@ -409,7 +362,7 @@ export function instructorsPermissionTests(api) {
 
     it("fetch contract templates", async () => {
         await switchToInstructorOnlyUser();
-        let resp = await apiGET(
+        const resp = await apiGET(
```
siefkenj: No sense in changing all of these to `const`. It's just cluttering the PR.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+        beforeAll(async () => {
+            await restoreDefaultUser();
+
+            // Get our instructor
+            let resp = await apiGET(`/admin/instructors`);
+            expect(resp).toHaveStatus("success");
+            instructor = resp.payload.find(
+                (instructor) => instructor.utorid === instructorUser.utorid
+            );
+            expect(instructor).toBeDefined();
```
siefkenj: Isn't this instructor already in the seed data?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+
+            // Fetch the position related to our instructor
+            resp = await apiGET(`/admin/sessions/${session.id}/positions`);
+            position = resp.payload.find((position) =>
+                position.instructor_ids.find((id) => id === instructor.id)
+            );
+            expect(position).toBeDefined();
```
siefkenj: The position is in the seed data too, right?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+
+            // Create another assignment (with no DDAH)
+            // for position associated with our instructor
+            const newAssignment = {
+                note: "",
```
siefkenj: just leave off `note`

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+            const otherInstructor = resp.payload.find(
+                (instructor) =>
+                    instructor.utorid ===
+                    databaseSeeder.seededData.instructors[4].utorid
+            );
+            expect(otherInstructor).toBeDefined();
+            resp = await apiGET(`/admin/sessions/${session.id}/positions`);
+            const otherPosition = resp.payload.find((position) =>
+                position.instructor_ids.find((id) => id === otherInstructor.id)
+            );
```
siefkenj: already in the seed data?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+
+            // Only one DDAH is intially seeded for this instructor
+            expect(resp.payload).toHaveLength(1);
+
+            // We get duties returned sorted in ascending order, so
+            // we need to sort the seeded ones before comparing
+            const sortedSeededDuties = databaseSeeder.seededData.ddahs[0].duties.sort(
+                (first, second) => first.order - second.order
+            );
+            expect(resp.payload[0].duties).toEqual(sortedSeededDuties);
```
siefkenj: Better to save the DDAH as a new variable rather than access the payload directly.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+                    {
+                        order: 2,
+                        hours: 50,
+                        description: "other:Lecture support",
+                    },
+                ],
+            };
+            let resp = await apiPOST(`/instructor/ddahs`, ddahToCreate);
+            expect(resp).toHaveStatus("success");
+            expect(resp.payload).toEqual(expect.objectContaining(ddahToCreate));
```
siefkenj, line 556: How does this work? Isn't `resp.payload` an array? We have a `toContainObject` extension to expect for just this case..

frontend/src/tests/instructor-permission-test.js
siefkenj: Agreed. Don't test the error messages.

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+                "Cannot create a DDAH without an assignment_id OR improper permissions to access/create a DDAH"
+            );
+        });
+
+        it("cannot update a Ddah for an assignment not associated with self", async () => {
+            // First we add the DDAH to the assignment
+            // associated with another instructor
+            await restoreDefaultUser();
+            const ddahToCreate = {
+                assignment_id: foreignAssignment.id,
```
siefkenj: `foreignAssignment` is a strange name...

frontend/src/tests/instructor-permission-test.js
```diff
@@ -516,51 +469,254 @@ export function instructorsPermissionTests(api) {
...
+                    {
+                        order: 1,
+                        hours: 30,
+                        description: "contact:Practice",
+                    },
+                ],
+            };
+            let resp = await apiPOST(`/admin/ddahs`, ddahToCreate);
+            expect(resp).toHaveStatus("success");
+            expect(resp.payload.duties).toStrictEqual(ddahToCreate.duties);
```
siefkenj: We've already done this check in another test, so you can just assume it's working.

frontend/src/tests/setup.js
```diff
@@ -66,6 +66,18 @@ class DatabaseSeeder {
...
                     utorid: "millerm",
                 },
+                {
+                    last_name: "Lucas",
+                    first_name: "George",
+                    email: "george.lucas@utoronto.ca",
+                    utorid: "lucasg",
+                },
+                {
+                    last_name: "Bellford",
```
siefkenj: Maybe `Bell`? We don't want to be accidentally associated with Jordan Belfort...

frontend/src/tests/setup.js
```diff
@@ -173,6 +201,58 @@ class DatabaseSeeder {
                     position_code: "MAT235H1F",
                     applicant_utorid: "weasleyr",
                 },
+                {
+                    ddah_seed_id: 1,
```
siefkenj: prefix with `_` for fields that aren't "real"

frontend/src/tests/instructor-permission-test.js
alex-kozin: While I don't like this approach either, that is similar to what's done [here](https://github.com/uoft-tapp/tapp/blob/fd163fd2b23401d63931250d1da9f6212634c416/frontend/src/tests/ddah-tests.js#L91) for the ddah tests. I see before it used `.toContain()` and I like that better too. I am not sure that's the direction we will be going in though. @siefkenj do you have a preference?

frontend/src/tests/instructor-permission-test.js
alex-kozin: In this case `resp.payload` is an object and that was the reason why the test was failing! `toContainObject` would not work with objects.

For `.toEqual` Jest allows its matchers to be passed in the function. It can match a literal or match based on Jest's `expect.matcher()` expressions. In this case we are matching any object that recursively has the properties of the `ddahToCreate`!

frontend/src/tests/instructor-permission-test.js
siefkenj: I think you should fetch all DDAHs as admin and filter that list to contain only those for that instructor. Then check the list the instructor retrieves is the same.

frontend/src/tests/instructor-permission-test.js
alex-kozin: Any suggestions what it could be? :slightly_smiling_face:

frontend/src/tests/instructor-permission-test.js
alex-kozin: It is, but without the id, I need the id to supply later!

frontend/src/tests/instructor-permission-test.js
alex-kozin: Yes, but I need its id, which I don't have from the seeded data alone

frontend/src/tests/instructor-permission-test.js
alex-kozin: As with the other position, I need the id which I can't get from seeded data alone

frontend/src/tests/instructor-permission-test.js
alex-kozin: That's similar to [this approach here](https://github.com/uoft-tapp/tapp/blob/fd163fd2b23401d63931250d1da9f6212634c416/frontend/src/tests/instructor-permission-test.js#L116), so I assumed that would be ok for one-time use, will change.

frontend/src/tests/instructor-permission-test.js
alex-kozin: Ok, will delete. Out of curiosity, wouldn't that be consistent with testing our assumptions (like Mert suggested with payload to be defined before `.find` it)?

frontend/src/tests/instructor-permission-test.js
alex-kozin: It's a great idea! For the implementation though, we don't have an endpoint to get all DDAHs per instructor? That means we will have to fetch all positions, filter them per instructor, then get all assignments for these positions and then get their respective DDAHs? That would complicate the test a lot imo, even though that will yield the most predictable outcome regardless of the seeded data.

That also conflicts with the idea that we "know the seeded data" and can rely on it unless I am missing something of course

frontend/src/tests/instructor-permission-test.js
siefkenj: No...that's not how it works. Seed data should always be an exact duplicate of what you'd get if you hit the endpoint.

frontend/src/tests/instructor-permission-test.js
siefkenj: `assignmentInstructorCantAccess`

frontend/src/tests/instructor-permission-test.js
alex-kozin: Oh, are you saying I can force the id of seeded instructor? I thought the database would insert it with whatever id it wants?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -524,43 +475,242 @@ export function instructorsPermissionTests(api) {
...
+                        description: "other:Additional duties",
+                    },
+                ],
+            };
+            resp = await apiPOST(`/instructor/ddahs`, updatedDdah);
+            expect(resp).toHaveStatus("success");
+            expect(resp.payload).toEqual(expect.objectContaining(updatedDdah));
+        });
+
+        // This test indeed fails and allows to set these fields 0w0
```
alex-kozin: Include issue # here

frontend/src/tests/setup.js
```diff
@@ -173,6 +201,58 @@ class DatabaseSeeder {
                     position_code: "MAT235H1F",
                     applicant_utorid: "weasleyr",
                 },
+                {
+                    _ddah_seed_id: 1,
```
alex-kozin: make these the same - `temp_id`

frontend/src/tests/setup.js
```diff
@@ -431,5 +514,37 @@ async function seedDatabaseForInstructors(
         resp = await apiPOST(`/admin/assignments`, assignment);
         expect(resp).toHaveStatus("success");
         Object.assign(assignment, resp.payload);
```
alex-kozin, line 516: Checkout this - we should have ids, don't forget to delete fake ids too, util function might be handy

frontend/src/tests/setup.js
```diff
@@ -431,5 +514,49 @@ async function seedDatabaseForInstructors(
...
+            assignment_id,
+        };
+        let resp = await apiPOST(`/admin/ddahs`, newDdah);
+        expect(resp).toHaveStatus("success");
+    }
+
+    // Cleanup
+    for (const assignment of seeded.assignments) {
+        if (assignment._temp_id) {
+            const { temp_id, ...cleanAssignment } = assignment;
```
alex-kozin, line 553: @alex-kozin Fix this - delete all properties with an underscore

PR #605, opened 2021-06-06
Add more DDAH tests

alex-kozin:
This PR adds additional test coverage for DDAH functionality (related to Issue #606)


frontend/src/tests/ddah-tests.js
```diff
@@ -414,5 +507,57 @@ export function ddahsEmailAndDownloadTests(api) {
...
+        let resp = await apiPOST(`/admin/ddahs`, ddahWithAllDutyTypes);
+        expect(resp).toHaveStatus("success");
+
+        const ddahHtml = (
+            await axios.get(
+                `${BACKEND_BASE_URL}/public/ddahs/${ddah.url_token}`
+            )
+        ).data;
+
+        expect(ddahHtml).toMatchSnapshot();
```
alex-kozin: Opted for a snapshot here to make sure this test fails every time HTML is updated - this way the snapshot will always be checked for containing the categories

frontend/src/tests/api.test.js
```diff
@@ -148,7 +146,4 @@ describe("Mock API tests", () => {
     describe("User permissions tests", () => {
         userPermissionsTests(mockAPI);
     });
-    describe("DDAH tests", () => {
-        ddahTests({ apiGET, apiPOST });
-    });
```
alex-kozin, line 153: These tests were failing for Mock API - deleted, since we don't intend to support all use cases for Mock API

frontend/src/tests/ddah-tests.js
siefkenj: I don't like this. It's not going to be reliable, and the html will likely change often. Please check the html for the duties.

alex-kozin: The reason why I think it is better as a snapshot is because we want to make sure all categories are properly displayed, not just contained in the HTML. I can change it to a simple matcher, if you prefer that of course, but if the styles make them not render properly or HTML changes -  I thought we would want to revisit the snapshot and make sure categories are visible. We could possibly use Jest's snapshot serializer to display a more meaningful message.

For the matcher option, would you like a simple `.stringContaining()` or anything else for the check? :slightly_smiling_face:

siefkenj: `stringContaining` would be fine. We don't want to reply on by-hand inspection. That's a recipe for disaster.

alex-kozin:
Travis is fed up with checking my pull request :joy:

frontend/src/tests/ddah-tests.js
```diff
@@ -67,7 +67,7 @@ export function ddahTests(api) {
             ],
         };
 
-        let resp = await apiPOST(`/admin/ddahs`, newDdah);
+        const resp = await apiPOST(`/admin/ddahs`, newDdah);
```
mertkaya1033: I see that you have changed the resp object to be a constant every time you make an API request. Even though this is not incorrect, it might be more beneficial to make resp a global variable declared with let and used in every test. This will also allow you to make more than 1 API request within a single test :).

mertkaya1033 (approved):
Everything seems great.

frontend/src/tests/ddah-tests.js
alex-kozin: I believe that was my local linter that changed these :smile:

Usually, I try to stick to using `const` whenever possible, and `let` only if I know the variable will change. This helps to avoid some cases where I expect that variable to hold some value before it's reassigned in that test. Kind of similar to [this thread](https://softwareengineering.stackexchange.com/questions/278652/how-much-should-i-be-using-let-vs-const-in-es6).

I believe it's also a common practice / style decision made in many es6 codebases, just like using global variables is a code smell usually (obviously it might be warranted for some cases like the test suite we have). Of course, it's a style matter though!

siefkenj: Let's put them back as `let`. This change doesn't have anything to do with this PR

frontend/src/tests/ddah-tests.js
```diff
@@ -406,13 +513,80 @@ export function ddahsEmailAndDownloadTests(api) {
...
+            "Marking/Grading",
+            "Contact Time",
+            "Preparation",
+            "Meetings",
+            "Notes",
+            "Other duties",
+        ];
+
+        for (const expectedString of expectedStrings) {
+            expect(ddahHtml).toEqual(expect.stringContaining(expectedString));
```
siefkenj, line 589: The labels are not the important part. It's the duty descriptions which are important. Currently, if you make `duties` an empty array, your test will still pass.

alex-kozin: :open_mouth:  Thanks for pointing that out!

siefkenj:
Great job!

Issue #608, opened 2021-06-11
Setting emailed_date for a Ddah through instructor/ddahs route should not be allowed


alex-kozin:
Please see `"cannot set approved_date/accepted_date/revised_date/emailed_date/signature for a Ddah associated with self"` test in `instructor-permission-test` test file. It is skipped in PR #599 and should pass once this issue is resolved

siefkenj:
@alex-kozin Did the merged PR fix this issue? Can we close it?

PR #612, opened 2021-06-18
Fix DDAH permissions

alex-kozin:
This PR fixes Issue #608 by restricting which parameters are permitted in ddah controllers for instructors. It also reduces the number of ways one could modify existing ddah by forcing to go through `/accept` `/approve` and `/email` routes instead.

Additionally, some tests are fixed

### DIFF
--- a/backend/app/controllers/api/v1/admin/ddahs_controller.rb
+++ b/backend/app/controllers/api/v1/admin/ddahs_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
-class Api::V1::Admin::DdahsController < ApplicationController # GET /sessions/:session_id/ddahs
+class Api::V1::Admin::DdahsController < ApplicationController
+    # GET /sessions/:session_id/ddahs
     def index
         render_success Ddah.by_session(params[:session_id])
     end
@@ -32,18 +33,25 @@ class Api::V1::Admin::DdahsController < ApplicationController # GET /sessions/:s
 
     # GET /assignments/:id/ddah
     def show_by_assignment
+        # Because we are being routed through `/assignments`, the param is :id
+        # and not :assignment_id
         assignment = Assignment.find(params[:id])
         render_success assignment.ddah
     end
 
     # POST /assignments/:id/ddah
     def upsert_by_assignment
+        # We'll mangle the `params` variable in place and fall back to the
+        # usual `create` function
+        # Because we are being routed through `/assignments`, the param is :id
+        # and not :assignment_id
         params[:assignment_id] = params[:id]
         create
     end
 
     # POST /ddahs
     def create
+        # look up the DDAH first by assignment ID, otherwise, create a new DDAH
         @ddah =
             if params[:assignment_id]
                 Assignment.find_by(id: params[:assignment_id]).ddah ||
@@ -80,10 +88,11 @@ class Api::V1::Admin::DdahsController < ApplicationController # GET /sessions/:s
     def delete
         find_ddah
         render_on_condition(
-            object: @ddah,
-            condition: proc { @ddah.destroy! },
-            error_message: "Could not delete ddah '#{@ddah.id}'."
-        )
+          object: @ddah,
+          condition: proc { @ddah.destroy! },
+          error_message:
+            "Could not delete ddah '#{@ddah.id}'."
+      )
     end
 
     # POST /ddahs/:ddah_id/email
@@ -125,9 +134,11 @@ class Api::V1::Admin::DdahsController < ApplicationController # GET /sessions/:s
         template_file = "#{contract_dir}/ddah-signature-list.html"
 
         # Verify that the template file is actually contained in the template directory
-        unless Pathname.new(template_file).realdirpath.to_s.starts_with?(
-                   contract_dir.to_s
-               )
+        unless Pathname
+                   .new(template_file)
+                   .realdirpath
+                   .to_s
+                   .starts_with?(contract_dir.to_s)
             raise StandardError, "Invalid contract path #{template_file}"
         end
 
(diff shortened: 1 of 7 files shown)


backend/app/controllers/api/v1/admin/ddahs_controller.rb
```diff
@@ -115,18 +115,7 @@ def find_ddah
...
-            :assignment_id,
-            :approved_date,
-            :accepted_date,
-            :revised_date,
-            :emailed_date,
-            :signature
-        ).permit!
-    end
-
-    def duty_params
```
alex-kozin, line 128: Here and below `ddah_params` are unused in individual controllers

backend/app/controllers/api/v1/admin/ddahs_controller.rb
```diff
@@ -115,18 +115,7 @@ def find_ddah
     end
 
     def ddah_params
-        params.slice(
-            :assignment_id,
-            :approved_date,
-            :accepted_date,
-            :revised_date,
-            :emailed_date,
-            :signature
```
alex-kozin, line 124: We want to only allow `assignment_id`, the rest is updated in the DdahService or thorugh individual `accept`, `approve`, and `email` routes.

backend/app/services/ddah_service.rb
```diff
@@ -119,14 +119,7 @@ def normalize_duty(duty:)
     end
 
     def ddah_params
-        @params.slice(
-            :assignment_id,
-            :approved_date,
-            :accepted_date,
-            :revised_date,
-            :emailed_date,
-            :signature
```
alex-kozin, line 128: We don't want to permit these to be updated through the service, since individual routes implement functionality to accept, sign, and email ddahs.

frontend/src/tests/ddah-tests.js
```diff
@@ -272,20 +272,21 @@ export function ddahTests(api) {
...
         expect(resp.payload.approved_date).toBeFalsy();
 
-        // Updating a signed DDAH should remove the signature
-        resp = await apiPOST(`/admin/ddahs`, ddah);
+        // Sign the DDAH
+        resp = await apiPOST(
+            `/public/ddahs/${ddah.url_token}/accept`,
+            { signature: "My Sig" },
+            true
+        );
```
alex-kozin, line 280: Signing the DDAHs should be done through a special route, and not through POSTing to the `admin/ddahs` route.

frontend/src/tests/ddah-tests.js
```diff
@@ -272,20 +272,21 @@ export function ddahTests(api) {
...
         expect(resp).toHaveStatus("success");
-        ddah = resp.payload;
-        resp = await apiPOST(`/admin/ddahs`, {
-            ...ddah,
-            signature: "My Sig",
-            accepted_date: new Date(),
-        });
+        // Get the signed DDAH since it is not returned from public route
+        resp = await apiGET(`/admin/assignments/${newAssignment.id}/ddah`);
```
alex-kozin, line 284: I suppose the reason that the `/public` route does not return the newly updated record is security-related?

frontend/src/tests/instructor-permission-test.js
```diff
@@ -600,37 +600,36 @@ export function instructorsPermissionTests(api) {
...
             expect(resp).toHaveStatus("success");
-            const newDdah = resp.payload.find(
-                (ddah) => ddah.assignment_id === assignment.id
-            );
-            expect(newDdah).toEqual(expect.objectContaining(originalDdah));
+            const originalDdahWithUpdatedDuties = {
+                ...originalDdah,
+                duties: updatedDuties,
+            };
+            expect(resp.payload).toEqual(originalDdahWithUpdatedDuties);
```
alex-kozin, line 632: Here we are doing pretty much the same test as before, but also checking that duties updates are still permitted whereas other updates (dates and signature) are not.

backend/app/models/ddah.rb
```diff
@@ -2,7 +2,7 @@
 
 class Ddah < ApplicationRecord
     belongs_to :assignment
-    has_many :duties, dependent: :destroy
+    has_many :duties, -> { order(:order) }, dependent: :destroy
```
alex-kozin, line 5: The other syntax did not seem to work, so I used the one suggested [here](https://stackoverflow.com/questions/18284606/deprecated-warning-for-rails-4-has-many-with-order). As a bonus, it's compatible with Rails 4

backend/docs/erd.svg
```diff
@@ -62,7 +62,7 @@
 <text text-anchor="start" x="288" y="-621.63" font-family="Arial Italic" font-size="10.00" fill="#999999">string ∗</text>
 </g>
 <!-- m_ActiveStorage::Blob&#45;&gt;m_ActiveStorage::Attachment -->
-<g id="edge19" class="edge">
+<g id="edge22" class="edge">
```
alex-kozin: These changes happened after re-migrating with duties ordering, not sure if they are significant, but maybe still worth merging so this file doesn't show up in changes on every setup?

frontend/src/tests/ddah-tests.js
```diff
@@ -272,27 +272,28 @@ export function ddahTests(api) {
...
+        // Updating a signed DDAH should remove the signature
         resp = await apiPOST(`/admin/ddahs`, {
             id: signedDdah.id,
             duties: [
                 ...ddah.duties,
                 {
-                    order: 10,
-                    description: "meeting:Hot-pocket eating contest",
+                    order: 11,
+                    description: "meeting:Hot-pocket eating contest 2.0",
```
alex-kozin, line 296: These changes actually change duties and so should discard the signature

backend/docs/erd.svg
siefkenj: schema didn't change. Re-checout this file from master and commit. Then this diff will go away.

siefkenj:
Great job!

PR #613, opened 2021-06-18
Prefill survey data based on previous application(s)

siefkenj:
Resolves some of items in #590

### DIFF
--- a/backend/app/services/posting_service.rb
+++ b/backend/app/services/posting_service.rb
@@ -92,18 +92,19 @@ class PostingService
                     :"offers.position_start_date",
                     :"offers.position_end_date"
                 )
-            if !offer_history.blank?
+            unless offer_history.blank?
                 data[:previous_department_ta] = true
                 data[:previous_university_ta] = true
                 # Technically we cannot infer this, but if they've TAed for the department
                 # before, we don't care about other history
                 data[:previous_other_university_ta] = false
                 data[:previous_experience_summary] =
-                    offer_history.map { |(course, hours, start_date, end_date)|
+                    offer_history
+                        .map do |(course, hours, start_date, end_date)|
                         "#{course} (#{hours} hours) from #{
                             start_date.strftime('%b %Y')
                         } to #{end_date.strftime('%b %Y')}"
-                    }.join '; '
+                    end.join '; '
             end
 
             # Some information rarely changes from application to application.
@@ -124,8 +125,7 @@ class PostingService
                 data[:program] = last_application_data[:program]
             end
             if last_application_data[:program_start]
-                data[:program_start] =
-                    last_application_data[:program_start]
+                data[:program_start] = last_application_data[:program_start]
             end
         end
 
--- a/backend/app/services/posting_service.rb
+++ b/backend/app/services/posting_service.rb
@@ -118,15 +118,11 @@ class PostingService
             application_service =
                 ApplicationService.new application: last_application
             last_application_data = application_service.prefilled_data
-            if last_application_data[:department]
-                data[:department] = last_application_data[:department]
-            end
-            if last_application_data[:program]
-                data[:program] = last_application_data[:program]
-            end
-            if last_application_data[:program_start]
-                data[:program_start] = last_application_data[:program_start]
-            end
+            data.merge! last_application_data.slice(
+                            :department,
+                            :program,
+                            :program_start
+                        ).compact
         end
 
         data
--- a/backend/app/services/posting_service.rb
+++ b/backend/app/services/posting_service.rb
@@ -65,7 +65,8 @@ class PostingService
         applicant = Applicant.find_by(utorid: utorid)
         return { utorid: utorid } unless applicant
 
-        {
+        # This is the basic data we can get if the applicant already exists
+        data = {
             utorid: utorid,
             student_number: applicant.student_number,
             first_name: applicant.first_name,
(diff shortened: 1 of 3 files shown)


backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,71 @@ def prefill(user:)
...
             last_name: applicant.last_name,
             email: applicant.email,
             phone: applicant.phone
         }
+
+        existing_application = Application.find_by(posting: @posting)
+        if existing_application
+            application_service =
+                ApplicationService.new application: existing_application
+            data.merge! application_service.prefilled_data
```
alex-kozin, line 82: Do I understand correctly that we are 

1) Creating a new application by providing existing application (from our search by posting) as a parameter to the `ApplicationService`'s `new` method,
2) then merging the prefilled data of newly created application with the `data` described above (that we got from the applicant data)
3) And that happens only if such an application existed for a posting?

And the reason for doing that is so we could extract some posting-specific data and applicant-related data? How did it work before? Was it just prefilling the utorid?

backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,71 @@ def prefill(user:)
...
+        else
+            # Find out if they have previous contracts in the database.
+            # This will tell us information about whether the applicant has been previously employed
+            offer_history =
+                Assignment.joins(:applicant, :offers, :position).where(
+                    applicant: applicant, "offers.status": 'accepted'
+                ).order("offers.position_start_date": :ASC).pluck(
```
alex-kozin, line 89: Pretty cool ORM stuff! Do I understand correctly that `pluck` is like SELECT for specific columns?

backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,71 @@ def prefill(user:)
...
+                    :"positions.position_code",
+                    :"offers.hours",
+                    :"offers.position_start_date",
+                    :"offers.position_end_date"
+                )
+            unless offer_history.blank?
```
alex-kozin, line 95: What is the difference between `blank?` and `empty?` I tried looking it up, but it was kind of confusing, do I understand correctly that for arrays these would be equivalent?

backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,71 @@ def prefill(user:)
...
+                data[:previous_department_ta] = true
+                data[:previous_university_ta] = true
+                # Technically we cannot infer this, but if they've TAed for the department
+                # before, we don't care about other history
+                data[:previous_other_university_ta] = false
```
alex-kozin, line 100: I don't really know where it's used so far, but just curious if we have to specify this value at all (would it be `true` if we didn't? or would it be an error if we didn't?)

backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,71 @@ def prefill(user:)
...
+            # Some information rarely changes from application to application.
+            # For example, the program of study/department/program start.
+            # We retrieve this information from the most recently completed application
+            # if it's available.
+            last_application =
+                Application.joins(:applicant).where(applicant: applicant).order(
+                    updated_at: :DESC
+                ).first
+            application_service =
+                ApplicationService.new application: last_application
```
alex-kozin, line 119: Would it work fine if `last_application` couldn't be found? i.e. if it is the first application? Or is it impossible when `prefill` method is called? :slightly_smiling_face:

backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,71 @@ def prefill(user:)
...
+            last_application_data = application_service.prefilled_data
+            if last_application_data[:department]
+                data[:department] = last_application_data[:department]
+            end
+            if last_application_data[:program]
+                data[:program] = last_application_data[:program]
+            end
+            if last_application_data[:program_start]
+                data[:program_start] = last_application_data[:program_start]
+            end
```
alex-kozin: I don't know how `merge` works above, but would that be a similar use case here? Again, sorry if it's a dumb question, I'm still ramping up on Ruby :slightly_smiling_face:

backend/app/services/posting_service.rb
siefkenj: We aren't actually creating an a new application here. We're looking 1) at a current application, and if there is none 2) at a past application. Then we extract relevant information from those applications.

backend/app/services/posting_service.rb
siefkenj: Yes. That's exactly what it does. You can try it out in the rails console and see the query that it produces :-)

backend/app/services/posting_service.rb
siefkenj: `blank?` is more general than `empty?`. For example, `nil` has `blank?` but not `empty?`

backend/app/services/posting_service.rb
siefkenj: If we leave it off, it will be an unanswered question. Since it requires an answer, the user would have to interact with that page before moving on. Since UofT experience is way more important than any other experience, setting a default of `false` will prevent clutter in the applications.

backend/app/services/posting_service.rb
siefkenj: Yes. If you look at `ApplicationService.prefill`, a `nil` application will produce `{}` as a return value.

backend/app/services/posting_service.rb
siefkenj: Yeah. This might be cleaned up by using `merge` instead...

backend/app/services/posting_service.rb
alex-kozin: Oh I see, so we're prefilling false, makes sense!

backend/app/services/posting_service.rb
```diff
@@ -65,14 +65,67 @@ def prefill(user:)
...
+        data = {
             utorid: utorid,
             student_number: applicant.student_number,
             first_name: applicant.first_name,
             last_name: applicant.last_name,
             email: applicant.email,
             phone: applicant.phone
         }
+
+        existing_application = Application.find_by(posting: @posting)
```
siefkenj, line 78: @alex-kozin This line turns out to be a huge bug...I wasn't finding the application by the applicant, I was just picking the first application for the posting...

PR #625, opened 2021-07-01
Fallback to null and clear state if deleted session is active.

### LINKED ISSUE #564: [Frontend UI] Contract template and Position tables to be reinitialized after deleting sessions
Both contract template and position tables need to be reinitialized with Redux after deleting a session. (Currently, it requires manually refreshing the page)

alex-kozin:
Solves #564. We are now falling back to the `null` active session when deleting a currently active session. Additionally, we clear all session-related data. A small refactor so we actually do that on any `setActiveSession(null)` call. Confirmed this should not break anything in other parts of the app.

Visually, as soon as the active session gets deleted, we fall back to the default "unselected" session and the tables UI is cleared of anything related to the deleted session.

### DIFF
--- a/frontend/src/api/actions/init.ts
+++ b/frontend/src/api/actions/init.ts
@@ -253,12 +253,17 @@ export function initFromStage(
     };
 }
 
-export function clearSessionDependentData(): ThunkAction<Promise<void>, RootState, void, AnyAction> {
+export function clearSessionDependentData(): ThunkAction<
+    Promise<void>,
+    RootState,
+    void,
+    AnyAction
+> {
     return async (dispatch, getState) => {
         dispatch(fetchApplicantsSuccess([]));
         dispatch(fetchAssignmentsSuccess([]));
         dispatch(fetchContractTemplatesSuccess([]));
         dispatch(fetchApplicationsSuccess([]));
         dispatch(fetchPositionsSuccess([]));
-    }
+    };
 }
--- a/frontend/src/api/actions/init.ts
+++ b/frontend/src/api/actions/init.ts
@@ -253,13 +253,20 @@ export function initFromStage(
     };
 }
 
+/**
+ * Clear all session-specific store data: applicants, assignments,
+ * contract templates, applications, and positions.
+ *
+ * @export
+ * @returns an async function that handles all the API calls.
+ */
 export function clearSessionDependentData(): ThunkAction<
     Promise<void>,
     RootState,
     void,
     AnyAction
 > {
-    return async (dispatch, getState) => {
+    return async (dispatch) => {
         dispatch(fetchApplicantsSuccess([]));
         dispatch(fetchAssignmentsSuccess([]));
         dispatch(fetchContractTemplatesSuccess([]));
--- a/frontend/src/api/actions/init.ts
+++ b/frontend/src/api/actions/init.ts
@@ -229,7 +229,7 @@ export function initFromStage(
             // Before fetching session-related data, the existing data
             // should be cleared. It will be re-fetched via the appropriate routes,
             // but clearing now will prevent excess re-renders as data streams in.
-            dispatch(clearSessionDependentData());
+            await dispatch(clearSessionDependentData());
 
             // `fetchActions` array contains all the fetch API calls that need to be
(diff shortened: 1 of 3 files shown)


frontend/src/components/sessions.js
```diff
@@ -128,6 +140,16 @@ export function ConnectedSessionsList(props) {
                 show={!!sessionToDelete}
                 onHide={() => setSessionToDelete(null)}
                 onDelete={async () => {
+                    if (sessionToDelete === activeSession) {
+                        await Promise.all([
```
siefkenj: We have an init session that should be used instead. Look into `initFromStage`. You shouldn't be directly dispatching these fetch successes from UI code.

frontend/src/api/actions/init.ts
```diff
@@ -97,7 +97,7 @@ export function initFromStage(
     stage: InitStages,
     options = { startAfterStage: false }
 ): ThunkAction<Promise<void>, RootState, void, AnyAction> {
-    const startAfterStage = +!!options.startAfterStage;
+    const startAfterStage = options.startAfterStage ? 1 : 0;
```
alex-kozin, line 100: Thought it would be easier to read without sacrificing the guaranteed type conversion

frontend/src/api/actions/init.ts
```diff
@@ -233,11 +229,7 @@ export function initFromStage(
             // Before fetching session-related data, the existing data
             // should be cleared. It will be re-fetched via the appropriate routes,
             // but clearing now will prevent excess re-renders as data streams in.
-            dispatch(fetchApplicantsSuccess([]));
-            dispatch(fetchAssignmentsSuccess([]));
-            dispatch(fetchContractTemplatesSuccess([]));
-            dispatch(fetchApplicationsSuccess([]));
-            dispatch(fetchPositionsSuccess([]));
+            await dispatch(clearSessionDependentData());
```
alex-kozin, line 232: Awaiting here to make sure we clear state before we populate it again

frontend/src/components/sessions.js
```diff
@@ -128,6 +135,9 @@ export function ConnectedSessionsList(props) {
                 show={!!sessionToDelete}
                 onHide={() => setSessionToDelete(null)}
                 onDelete={async () => {
+                    if (sessionToDelete === activeSession) {
+                        await dispatch(setActiveSession(null));
```
alex-kozin, line 139: We only need to clear the active session if we are deleting the active session.

frontend/src/api/actions/init.ts
```diff
@@ -181,12 +181,8 @@ export function initFromStage(
         if (shouldRunStage("setActiveUserRole")) {
             // When the role is changed, data should be cleared immediately.
             // It will be re-fetched via the appropriate routes.
-            dispatch(fetchApplicantsSuccess([]));
-            dispatch(fetchAssignmentsSuccess([]));
-            dispatch(fetchContractTemplatesSuccess([]));
-            dispatch(fetchApplicationsSuccess([]));
+            dispatch(clearSessionDependentData());
             dispatch(fetchInstructorsSuccess([]));
```
alex-kozin, line 185: Do we need to await these? Does `setActiveUserRole` change state?

siefkenj: I don't think instructors depend on a session. There is no need to clear them.

frontend/src/api/actions/init.ts
```diff
@@ -260,3 +252,18 @@ export function initFromStage(
         await Promise.all(asyncActions);
     };
 }
+
+export function clearSessionDependentData(): ThunkAction<
```
siefkenj, line 263: Missing doc string

frontend/src/api/actions/init.ts
alex-kozin: Sorry, not sure I understand. That's not session dependent data indeed, but I suppose we are clearing instructors in `setActiveUserRole` because we might be changing the permissions and shouldn't have them in a state anymore?

siefkenj: Ah. I see. This is not a line you added :-)

siefkenj:
Good work!

PR #628, opened 2021-07-12
Make headers sticky for all matrix questions

### LINKED ISSUE #618: Postings UI/UX Issues
This is a tracking issue for things that come up when testing the postings functionality.

alex-kozin:
Related to #618 For a better UX we need the header with choices of position preferences to be sticky. This behavior will apply to all matrix (tables) questions. Video to show the new behavior:

https://user-images.githubusercontent.com/23097023/125226229-a5d76900-e29e-11eb-9849-9a6449e9d7e1.mp4

### DIFF
--- a/frontend/src/views/public/postings/survey.css
+++ b/frontend/src/views/public/postings/survey.css
@@ -12,3 +12,11 @@ the active button to be bright. */
 .sv_main.sv_main .sv-boolean__switch {
     background-color: var(--primary) !important;
 }
+
+.table.sv_q_matrix thead th, .table.sv_q_matrix thead td {
+    position: sticky;
+    top: 0;
+    border-top: 0;
+    z-index: 1;
+    background: white;
+}


frontend/src/views/public/postings/survey.css
```diff
@@ -11,4 +11,12 @@ the active button to be bright. */
 
 .sv_main.sv_main .sv-boolean__switch {
     background-color: var(--primary) !important;
+}
+
+.table.sv_q_matrix thead th, .table.sv_q_matrix thead td {
+    position: sticky;
+    top: 0;
+    border-top: 0;
+    z-index: 1;
```
alex-kozin, line 20: This is required as the table cells below the headers are [set to be on top by default](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/Stacking_without_z-index).

frontend/src/views/public/postings/survey.css
```diff
@@ -11,4 +11,12 @@ the active button to be bright. */
 
 .sv_main.sv_main .sv-boolean__switch {
     background-color: var(--primary) !important;
+}
+
+.table.sv_q_matrix thead th, .table.sv_q_matrix thead td {
```
alex-kozin, line 16: We could specify the `position_preferences` name directly, but I thought it would be reasonable to make a sticky header for all table questions instead.

siefkenj:
This looks great. I love the preview video!

PR #632, opened 2021-07-23
Change header for instructor view to be dynamic - display positions for current session

alex-kozin:
Change the instructor header view to be dynamic and display all positions for selected session. Under each position there are two tabs - TA Information (old component still) and DDAHs (not implemented in this PR). Please, see screenshot for an example:

Landing view (as before):
image

Position view - TA Information is selected by default when a position is selected:
image

Position view - DDAHs view is not implemented yet
image


frontend/src/views/instructor/routes/index.tsx
```diff
@@ -1,27 +1,26 @@
...
+            <Route
+                path="/:positionId"
+                render={({match}) => {
+                    const positionId = match.params.positionId;
+                    return <Redirect to={`/${positionId}/assignments`} />;
+                }}
+            />
+            <Route path="/:positionId/assignments">
                 <InstructorAssignmentsView />
             </Route>
```
alex-kozin, line 22: This `InstructorAssignmentsView` component seems to not render, and I have no idea why, was thinking it's redux's `shouldComponentUpdate()`, but wrapping in `withRouter` did not help. Any thoughts @siefkenj ?

PR #633, opened 2021-07-26
 finish file attachment test cases for application

ghost:
Reference to issue #590


frontend/src/tests/application-tests.js
```diff
@@ -6,20 +6,72 @@ import {
...
     let session;
     let position;
     let posting = {};
     let adminUser;
-    let surveyData;
+
+    const BACKEND_BASE_URL = "http://backend:3000";
+
+    // Returns hashes rather than expecting their equality as possible fail is untracable back to the test case
+    async function getMD5Hashes(url_token, file_type) {
```
alex-kozin: Do we want a better dosctring for these? I am not sure what `file_type` I can provide based on the current comment. The `url_token` here is the token to get on the `/public/files/` path, right?

frontend/src/tests/application-tests.js
```diff
@@ -6,20 +6,72 @@ import {
...
+        },
+    };
 
     const taOnlyUser = {
         utorid: "matthewc",
         roles: ["ta"],
     };
 
-    const postingData = {
+    let postingData = {
```
alex-kozin: Why this change? You do not reassign this object as far as I can tell?

frontend/src/tests/application-tests.js
```diff
@@ -37,6 +89,13 @@ export function applicationsTests({ apiGET, apiPOST }) {
             let resp = await apiGET(`/debug/active_user`);
             expect(resp).toHaveStatus("success");
             adminUser = resp.payload;
+            // Create taonlyuser
```
alex-kozin: Maybe 'user with TA permissions only'?

frontend/src/tests/application-tests.js
```diff
@@ -118,8 +151,11 @@ export function applicationsTests({ apiGET, apiPOST }) {
         );
         it.skip("When submitting survey.js data cannot add a position_preference for a position not listed in the posting", async () => {
```
alex-kozin: This is still skipped?

frontend/src/tests/application-tests.js
```diff
@@ -37,6 +89,13 @@ export function applicationsTests({ apiGET, apiPOST }) {
             let resp = await apiGET(`/debug/active_user`);
             expect(resp).toHaveStatus("success");
             adminUser = resp.payload;
+            // Create taonlyuser
+            resp = await apiPOST("/debug/users", taOnlyUser);
+            expect(resp).toHaveStatus("success");
+
+            surveyWithoutTranscript.answers.position_preferences = {
+                [position.position_code]: 3,
+            };
```
alex-kozin: I am not sure I understand the purpose of this change in the `beforeAll` ? Is that the check on line 159?

frontend/src/tests/application-tests.js
```diff
@@ -142,19 +178,289 @@ export function applicationsTests({ apiGET, apiPOST }) {
...
+                surveyWithTranscript,
+                true
+            );
+            expect(resp).toHaveStatus("success");
+
+            resp = await apiPOST("/debug/active_user", adminUser);
+            expect(resp).toHaveStatus("success");
+
+            resp = await apiGET(`/admin/sessions/${session.id}/applications`);
+            let url_token = resp.payload[0].documents[0].url_token;
```
alex-kozin, line 571: Would be nice to assert this exists before proceeding?

frontend/src/tests/application-tests.js
```diff
@@ -6,20 +6,72 @@ import {
...
+import md5 from "md5";
 
 export function applicationsTests({ apiGET, apiPOST }) {
     let session;
     let position;
     let posting = {};
     let adminUser;
-    let surveyData;
+
+    const BACKEND_BASE_URL = "http://backend:3000";
```
alex-kozin: (nb) This can be potentially be imported from `utils` file if we export the `URL` const.

frontend/src/tests/application-tests.js
```diff
@@ -142,19 +178,289 @@ export function applicationsTests({ apiGET, apiPOST }) {
...
+            let resp = await apiPOST("/debug/active_user", taOnlyUser);
+            expect(resp).toHaveStatus("success");
+
+            let str = fs.readFileSync(
+                path.resolve(__dirname, "./image-data/dummy.txt"),
+                {
+                    encoding: "base64",
+                }
+            );
+            let txt_str = "data:text/plain;base64," + str;
```
alex-kozin, line 547: What is the significance of this prepend? Is that the default format of the files that are processed and stored on the backend? My main worry is that if someone submits a regular file that does not have this metadata, are we guaranteed to receive a file with this metadata after they upload?

alex-kozin (commented):
Good job! I had several questions, but looking good otherwise

frontend/src/tests/application-tests.js
mertkaya1033: It will be removed with my PR.

frontend/src/tests/application-tests.js
ghost: This case tests using a `txt` file. When a user submits a text file, the part is prepended automatically into the string.

ghost:
@siefkenj Do you have any thoughts or comments, or can this be merged?

siefkenj:
> @siefkenj Do you have any thoughts or comments, or can this be merged?

I will get to this when I can. The start of the term is a bit crazy...

siefkenj:
Closing for now, since this PR isn't being worked on.

PR #642, opened 2021-08-02
Instructor DDAH View

siefkenj:
This PR adds an editable instructor DDAH view to the instructors view. It only shows the DDAHs for the actively selected position (available through the `activePositionSelector`). You can edit and create new DDAHs. Email functionality is not yet present.


frontend/src/components/header.js
```diff
@@ -72,7 +73,11 @@ export function Header(props) {
 
     // filters the routes to include only the current route, then maps all of that route's subroutes to BootstrapNavItems
     const availableSubroutes = routes
-        .filter((route) => route.route.substring(1) === mainRoute)
+        .filter(
+            (route) =>
+                route.route.substring(1) === mainRoute ||
+                fullRoute.startsWith(route.route)
```
alex-kozin, line 79: This is unrelated to the current PR, but it would be nice to have a dynamic header component, where matching rules would either come as props or be defined differently, because right now we are stuck with routes `/42/assignments/...` instead of `/positions/42/assignments` for that exact reason

frontend/src/views/instructor/header.tsx
```diff
@@ -16,10 +16,8 @@ import {
 export function InstructorHeader() {
     const positions = useSelector(positionsSelector);
     const routes = positions.map((position) => ({
-        route: `/${position.id}`,
-        name:
-            `${position.position_code} ${position.position_title}` ||
-            `Position #${position.id}`,
+        route: `/positions/${position.id}`,
+        name: `${position.position_code}`,
```
alex-kozin, line 20: My Typescript (a newer one because I still can't switch back to the one we're using) complained position title can be `null`, this helps with that as I suppose the code is always defined? In our types it's still a `string` so it can technically be empty, do we still want a fallback?

frontend/src/views/instructor/store/reducers.ts
```diff
@@ -0,0 +1,27 @@
...
+} from "./constants";
+import { createReducer } from "redux-create-reducer";
+
+interface InstructorUIState {
+    selectedDdahIds: number[];
+    activePositionId: number | null;
+}
+
+// initialize the state of offer table
+const initialState: InstructorUIState = {
```
alex-kozin, line 13: I just love typed Redux :heart:

frontend/src/views/instructor/store/actions.ts
```diff
@@ -0,0 +1,31 @@
...
+    payload: data,
+});
+
+export const setActivePositionId = (data: number | null) => ({
+    type: SET_INSTRUCTOR_ACTIVE_POSITION,
+    payload: data,
+});
+
+// selectors
+export const instructorUISelector = (state: RootState) => state.ui.instructor;
```
alex-kozin, line 21: Do we want a `null` fallback value here?

frontend/src/views/instructor/routes/index.tsx
```diff
@@ -1,21 +1,46 @@
...
+}
 
 export function InstructorRoutes() {
     return (
         <Switch>
             <Route exact path="/">
                 <InstructorLanding />
             </Route>
-            <Route path="/:positionId/assignments">
+            <Route path="/positions/:position_id/assignments">
```
alex-kozin, line 31: Will this route work with the header we currently have?

frontend/src/views/instructor/ddahs/ddah-editor.tsx
```diff
@@ -0,0 +1,509 @@
...
+import { stringToNativeType } from "../../../libs/urls";
+
+type Category =
+    | "note"
+    | "prep"
+    | "training"
+    | "meeting"
+    | "contact"
+    | "marking"
+    | "other";
```
alex-kozin, line 27: Can we make this a global type? I am sure it would be useful in other files too. Additionally, we could utilize stronger type with string templates, e.g.:

```
type Category = 'meeting' | 'prep' | ... ;
type DescriptionText = string;
....
export interface RawDuty {
    order: number;
    hours: number;
    description: `${Category}:${DescriptionText}`;
};
```

frontend/src/views/instructor/ddahs/ddah-editor.tsx
```diff
@@ -0,0 +1,509 @@
...
+                    <Alert variant="secondary">
+                        <FaInfoCircle className="mr-2" />
+                        {categoryInformation["note"].helpText}
+                    </Alert>
+                )}
+                <ul>
+                    <DutyList
+                        category={"note"}
+                        duties={dutiesByCategory["note"] || []}
+                        showEmptyDutyString={false}
```
alex-kozin, line 367: I see this prop is only used once, is there a use-case where we'd want to use the default `true`?

frontend/src/components/header.js
siefkenj: This change already fixes it. I changed all the routes to `/positions/42/assignments`

frontend/src/views/instructor/ddahs/ddah-editor.tsx
siefkenj: We want it `true` most of the time. Just not for `notes`

frontend/src/views/instructor/header.tsx
siefkenj: The title doesn't matter to the instructor. They know what course they're teaching :-). And the position code is always non-null

frontend/src/views/instructor/routes/index.tsx
siefkenj, line 31: With the fixed header code, it works!

frontend/src/views/instructor/store/actions.ts
siefkenj: No. It should never be able to be null.

frontend/src/views/instructor/routes/index.tsx
```diff
@@ -1,21 +1,46 @@
...
+/**
+ * React component that will update the active position id in the
+ * Redux store whenever it is rendered. This component is meant to be the first
+ * component in a route that depends on an active position.
+ */
+function UpdateActivePosition() {
+    const params = useParams<{ position_id?: string }>();
+    const dispatch = useThunkDispatch();
+    const rawPositionId = +(params.position_id || NaN);
+    const position_id = Number.isFinite(rawPositionId) ? rawPositionId : null;
```
alex-kozin, line 18: Can we change it to be something more readable perhaps? For example,

```
const rawPositionId = params.position_id;
const position_id = typeof rawPositionId === 'number' ? rawPositionId : null;
```

or 

```
const rawPositionId = params.position_id;
const position_id = isNan(rawPositionId) ? null : rawPositionId;
```

frontend/src/views/instructor/store/reducers.ts
siefkenj: It's pretty cool when it actually works!

frontend/src/views/instructor/header.tsx
```diff
@@ -16,10 +16,8 @@ import {
 export function InstructorHeader() {
     const positions = useSelector(positionsSelector);
     const routes = positions.map((position) => ({
-        route: `/${position.id}`,
-        name:
-            `${position.position_code} ${position.position_title}` ||
-            `Position #${position.id}`,
+        route: `/positions/${position.id}`,
```
alex-kozin, line 19: Will this route work with the header we currently have?

frontend/src/views/instructor/routes/index.tsx
siefkenj: Well, it starts out as a string. `isNan` could be used, but that lets `Infinity` through...

PR #652, opened 2021-08-09
Add columns to the TA Information table

alex-kozin:
This PR adds UTORid and DDAH columns to the TA Information table. The DDAH column has buttons to either Create or View existing DDAH through the modals that are the same for the DDAH table too.

https://user-images.githubusercontent.com/23097023/129055078-277e0b6c-e2f7-48f8-a94b-dac09ceda7ea.mp4

In the future, it would be nice to put the modal-related logic in Redux store and put the modals in the root of the page. This way we could display them from anywhere. This would allow us to package this modal display with the view/edit cell component and reduce repetition and complexity.

### DIFF
--- a/frontend/src/views/instructor/assignments/assignments-table.tsx
+++ b/frontend/src/views/instructor/assignments/assignments-table.tsx
@@ -7,12 +7,42 @@ import { assignmentsSelector } from "../../../api/actions";
 import { AdvancedFilterTable } from "../../../components/filter-table/advanced-filter-table";
 import { Assignment } from "../../../api/defs/types";
 import { ddahsSelector } from "../../../api/actions/ddahs";
+import { Button } from "react-bootstrap";
 
-export function InstructorAssignmentsTable() {
+interface RowData {
+    id: number;
+    applicant: {
+        first_name: string | null;
+        last_name: string | null;
+        utorid: string;
+        email: string | null;
+    };
+    ddah: {
+        id?: number;
+        assignment: {
+            id: number;
+        };
+    } | null;
+    hours: number;
+    active_offer_status: string | null;
+    active_offer_recent_activity_date: string | null;
+}
+
+export function InstructorAssignmentsTable({
+    position_id,
+    onView,
+    onCreate,
+    onEmail,
+}: {
+    position_id: number;
+    onView?: (ddah_id: number) => any;
+    onCreate?: (assignment_id: number) => any;
+    onEmail?: () => any;
+}) {
     const activePosition = useSelector(activePositionSelector);
     const allAssignments = useSelector(assignmentsSelector);
     const allDDAHs = useSelector(ddahsSelector);
-    const rowData = useMemo(() => {
+    const rowData: RowData[] = useMemo(() => {
         const currentAssignments = allAssignments.filter(
             (assignment) => assignment.position.id === activePosition?.id
         );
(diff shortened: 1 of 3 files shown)


frontend/src/views/instructor/assignments/assignments-table.tsx
```diff
@@ -6,22 +6,104 @@ import { formatDate } from "../../../libs/utils";
...
+        }));
+
+        return assignmentsWithDDAH;
+    }, [activePosition, allAssignments, allDDAHs]);
 
     if (!activePosition) {
         return <h4>No position currently selected</h4>;
     }
 
+    function ViewOrCreateCell({
```
alex-kozin, line 64: Would be really nice to create a separate component for this logic

frontend/src/views/instructor/assignments/index.tsx
```diff
@@ -50,7 +90,69 @@ export function InstructorAssignmentsView() {
...
+                />
+                <DdahPreviewModal
+                    ddah={previewDdah}
+                    show={previewVisible}
+                    onHide={() => setPreviewVisible(false)}
+                    onEdit={async (newDdah: Ddah) => {
+                        await dispatch(upsertDdah(newDdah));
+                    }}
+                />
+                <DdahPreviewModal
```
alex-kozin, line 141: If these modals were global - we could display them from anywhere using Redux store - a pattern I've seen used successfully before if we decide to do so. This would allow us to package this modal display with the view/edit cell component and reduce repetition and complexity

frontend/src/views/instructor/assignments/index.tsx
```diff
@@ -50,7 +90,69 @@ export function InstructorAssignmentsView() {
...
+                    position_id={activePosition?.id || -1}
+                    onView={(ddah_id) => {
+                        const ddah = ddahs.find((d) => d.id === ddah_id);
+                        if (ddah) {
+                            setPreviewDdah(ddah);
+                            setPreviewVisible(true);
+                        }
+                    }}
+                    onCreate={(assignment_id) => {
+                        const assignment = assignments.find(
```
alex-kozin, line 106: Tried wrapping these in a HOC but since we're using so many hooks it was not possible. If the `setPreview` and other display functions were handled with stateful logic (without the need to select ddahs) - we could abstract this out.

frontend/src/views/instructor/assignments/assignments-table.tsx
```diff
@@ -6,22 +6,104 @@ import { formatDate } from "../../../libs/utils";
...
-            allAssignments.filter(
-                (assignment) => assignment.position.id === activePosition?.id
-            ),
-        [activePosition, allAssignments]
-    );
+    const allDDAHs = useSelector(ddahsSelector);
+    const rowData: RowData[] = useMemo(() => {
+        const currentAssignments = allAssignments.filter(
+            (assignment) => assignment.position.id === activePosition?.id
+        );
```
alex-kozin, line 48: We could definitely make a selector for this

frontend/src/views/instructor/assignments/assignments-table.tsx
```diff
@@ -6,22 +6,104 @@ import { formatDate } from "../../../libs/utils";
...
+
+        const assignmentsWithDDAH = currentAssignments.map((assignment) => ({
+            ...assignment,
+            ddah:
+                allDDAHs.find((ddah) => ddah.assignment.id === assignment.id) ||
```
alex-kozin, line 53: Same here, a selector would be helpful

alex-kozin:
The problem with merging this now is that it will be confusing for next maintainers to figure out why we have duplicate code for the modal. Since there is no time left to refactor the modal (and have it work with Redux), we are closing this PR and splitting other changes in #659 and #660.

PR #656, opened 2021-08-15
Instructor view: render comprehensible changes to the DDAHs on import

alex-kozin:
This PR fixes several problems with the import UI for DDAHs in the instructor view:

1. Duties of the modified/created DDAHs are no longer a single string - they are rendered as a list with visual cues to see what will be the difference after import:

image
image

2. The import modal clears the name of the file on hitting 'Cancel' and preserves the filename along with changes when closed.
TODO: clear duties diff on 'Cancel' - component updates before unmounting

### DIFF
--- a/frontend/src/components/import-button.js
+++ b/frontend/src/components/import-button.js
@@ -53,7 +53,7 @@ export function ImportDialog({
 
     const withFileContentsReset = (actionHandler) => () => {
         setFileContents(null);
-        setFileArrayBuffer(null)
+        setFileArrayBuffer(null);
         actionHandler();
     };
 
--- a/frontend/src/components/import-button.js
+++ b/frontend/src/components/import-button.js
@@ -53,6 +53,7 @@ export function ImportDialog({
 
     const withFileContentsReset = (actionHandler) => () => {
         setFileContents(null);
+        setFileArrayBuffer(null)
         actionHandler();
     };
 
@@ -184,7 +185,7 @@ export function ImportDialog({
             <Modal.Footer>
                 <Button
                     variant="secondary"
-                    onClick={withFileContentsReset(withLabelReset(onCancel))}
+                    onClick={withLabelReset(withFileContentsReset(onCancel))}
                 >
                     Cancel
                 </Button>
--- a/frontend/src/components/import-button.js
+++ b/frontend/src/components/import-button.js
@@ -49,12 +49,12 @@ export function ImportDialog({
     const withLabelReset = (actionHandler) => () => {
         setFileInputLabel(DEFAULT_LABEL);
         actionHandler();
-    }
+    };
 
     const withFileContentsReset = (actionHandler) => () => {
         setFileContents(null);
         actionHandler();
-    }
+    };
 
     // When we are processing we want to set a spinner button
     // in the dialog as well as communicate to our parent
@@ -182,7 +182,10 @@ export function ImportDialog({
             </Modal.Body>
 
             <Modal.Footer>
-                <Button variant="secondary" onClick={withFileContentsReset(withLabelReset(onCancel))}>
+                <Button
+                    variant="secondary"
+                    onClick={withFileContentsReset(withLabelReset(onCancel))}
+                >
                     Cancel
                 </Button>
                 <Button variant="primary" onClick={_onConfirm}>
--- a/frontend/src/components/import-button.js
+++ b/frontend/src/components/import-button.js
@@ -46,8 +46,13 @@ export function ImportDialog({
 
     let formElement = null;
 
-    const withLabelReset = (actionHandler) => async () => {
+    const withLabelReset = (actionHandler) => () => {
         setFileInputLabel(DEFAULT_LABEL);
+        actionHandler();
+    }
+
+    const withFileContentsReset = (actionHandler) => () => {
+        setFileContents(null);
         actionHandler();
     }
 
@@ -177,7 +182,7 @@ export function ImportDialog({
             </Modal.Body>
 
             <Modal.Footer>
-                <Button variant="secondary" onClick={withLabelReset(onCancel)}>
+                <Button variant="secondary" onClick={withFileContentsReset(withLabelReset(onCancel))}>
                     Cancel
                 </Button>
                 <Button variant="primary" onClick={_onConfirm}>
--- a/frontend/src/components/import-button.js
+++ b/frontend/src/components/import-button.js
@@ -44,6 +44,13 @@ export function ImportDialog({
     const [fileContents, setFileContents] = React.useState(null);
     const [inProgress, _setInProgress] = React.useState(false);
 
+    let formElement = null;
+
+    const withLabelReset = (actionHandler) => async () => {
+        setFileInputLabel(DEFAULT_LABEL);
+        actionHandler();
+    }
+
     // When we are processing we want to set a spinner button
     // in the dialog as well as communicate to our parent
     // that we are in the midst of processing. Therefore, we
@@ -68,7 +75,7 @@ export function ImportDialog({
         if (onFileChange instanceof Function) {
             onFileChange(fileContents);
         }
-    }, [fileContents, onFileChange]);
+    }, [fileContents, onFileChange, formElement]);
 
     // Wrap the <input type="file" /> in an effect that parses the file
     React.useEffect(() => {
@@ -170,7 +177,7 @@ export function ImportDialog({
             </Modal.Body>
 
             <Modal.Footer>
-                <Button variant="secondary" onClick={onCancel}>
+                <Button variant="secondary" onClick={withLabelReset(onCancel)}>
                     Cancel
                 </Button>
                 <Button variant="primary" onClick={_onConfirm}>
--- a/frontend/src/views/admin/ddahs/import-export.tsx
+++ b/frontend/src/views/admin/ddahs/import-export.tsx
@@ -353,8 +353,7 @@ const DialogContent = React.memo(function DialogContent({
         if (newItems.length === 0 && modifiedDiffSpec.length === 0) {
             dialogContent = (
                 <Alert variant="warning">
-                    No difference between imported applicants and those already
-                    on the system.
+                    No difference between imported and existing DDAHs.
                 </Alert>
             );
         } else {
--- a/frontend/src/views/instructor/ddahs/import.tsx
+++ b/frontend/src/views/instructor/ddahs/import.tsx
@@ -1,6 +1,5 @@
 import React from "react";
 import { Alert } from "react-bootstrap";
-import { FaArrowRight } from "react-icons/fa";
 import { useSelector } from "react-redux";
 import { assignmentsSelector, applicantsSelector } from "../../../api/actions";
 import { ddahsSelector, upsertDdahs } from "../../../api/actions/ddahs";
@@ -10,6 +9,7 @@ import {
     Applicant,
     MinimalDdah,
     RawDuty,
+    Duty,
 } from "../../../api/defs/types";
 import { ImportActionButton } from "../../../components/import-button";
 import { DiffSpec, diffImport, getChanged } from "../../../libs/diffs";
@@ -23,10 +23,7 @@ enum DiffType {
     Deleted = "DELETED",
 }
 
-type MinimalDuty = {
-    description: string;
-    hours: number;
-};
+type MinimalDuty = Omit<Duty, "order">;
 
 interface DdahUpdate {
     updatedDdah: Ddah;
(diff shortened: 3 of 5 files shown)


frontend/src/views/instructor/ddahs/style.css
```diff
@@ -39,3 +39,41 @@
...
+}
+
+.instructor-ddah-import-duties-diff-list .duty-created:after {
+    content: " (new)";
+    color: green;
+    font-weight: bold;
+}
+
+.instructor-ddah-import-duties-diff-list .duty-updated {
+    list-style-type: "\270F";
```
siefkenj: Put in the actual character instead of the escape sequence. That way it's easier to understand what's going on. Also, prefer black and white characters over color emojis.

frontend/src/components/import-button.js
```diff
@@ -68,7 +80,7 @@ export function ImportDialog({
         if (onFileChange instanceof Function) {
             onFileChange(fileContents);
         }
-    }, [fileContents, onFileChange]);
+    }, [fileContents, onFileChange, formElement]);
```
siefkenj, line 84: What? You cannot track regular local variables. Only things set by react hooks can be listed here...

frontend/src/views/admin/ddahs/import-export.tsx
```diff
@@ -353,8 +353,7 @@ const DialogContent = React.memo(function DialogContent({
         if (newItems.length === 0 && modifiedDiffSpec.length === 0) {
             dialogContent = (
                 <Alert variant="warning">
-                    No difference between imported applicants and those already
-                    on the system.
+                    No difference between imported and existing DDAHs.
```
siefkenj, line 356: Good find!

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
+import React from "react";
+import { Alert } from "react-bootstrap";
+import { FaArrowRight } from "react-icons/fa";
```
siefkenj: Don't use this. Just use a unicode arrow

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+    Unchanged = "UNCHANGED",
+    Created = "CREATED",
+    Updated = "UPDATED",
+    Deleted = "DELETED",
+}
+
+type MinimalDuty = {
+    description: string;
+    hours: number;
+};
```
siefkenj: This is `Omit<Duty, "order">`, right?

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+    oldDuty?: MinimalDuty;
+    newDuty?: MinimalDuty;
+    status: DiffType;
+}
+
+function getDutiesDiff(oldDuties: RawDuty[], newDuties: RawDuty[]) {
+    let changes: DutyDiff[] = [];
+    let dutiesChanged = false;
+
+    oldDuties.forEach((oldDuty) => {
```
siefkenj: Prefer for loops over `forEach`. They're much faster.

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+                oldDutyDeleted = false;
+            } else if (
+                oldDuty.description === newDuty.description &&
+                oldDuty.hours !== newDuty.hours
+            ) {
+                changes.push({ oldDuty, newDuty, status: DiffType.Updated });
+                newDuties.splice(index, 1);
+                oldDutyDeleted = false;
+                dutiesChanged = true;
+            }
```
siefkenj, line 62: I'm confused about what exactly is going on...it looks like this is the same as a map with two filters?

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+        if (oldDutyDeleted) {
+            changes.push({ oldDuty, status: DiffType.Deleted });
+            dutiesChanged = true;
+        }
+    });
+
+    // The new duties that are left were just created - add them
+    newDuties.forEach((duty) =>
+        changes.push({ newDuty: duty, status: DiffType.Created })
+    );
```
siefkenj, line 75: `changes = changes.concat(newDuties.map(...))` or `changes.pusht(...newDuties.map(...))` if you're a fan of mutability.

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+    // Do not return changes if everything is unchanged
+    return dutiesChanged ? changes : [];
+}
+
+function dutiesDiffList(changes: DutyDiff[]) {
```
siefkenj: A react component that doesn't start with a capital?

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+    return (
+        <ul>
+            {changes.map(({ oldDuty, newDuty, status }) => {
+                switch (status) {
```
siefkenj: This makes me unhappy...Make a `DutyItem` component which contains this logic instead. I don't like seeing switch statement logic nested deep inside the JSX.

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+                            </li>
+                        );
+                    case DiffType.Deleted:
+                        return (
+                            <li className="duty-deleted">
+                                {oldDuty!.description} - {oldDuty!.hours}h
+                            </li>
+                        );
+                    default:
+                        return null;
```
siefkenj: You should never get here, right? You should be able to do `null as never` or something and that will make typescript check that you never reach this point in the code.

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+function DdahsList({ ddahs }: { ddahs: Omit<Ddah, "id">[] }) {
+    return (
+        <ul>
+            {ddahs.map(({ assignment: { applicant }, duties }, i) => (
+                <li>
+                    DDAH for {applicant.first_name} {applicant.last_name} (
+                    {applicant.utorid})
+                    <ul className="instructor-ddah-import-new-ddah-duties-list">
+                        {duties.map((duty) => (
+                            <li key={`ddah-new-${i}`}>
```
siefkenj: Why the elaborate key?

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+                                {duty.description} - {duty.hours}h
```
siefkenj: Let's wrap these in spans so we can style them separately

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+            {ddahUpdates.map(({ oldDdah, updatedDdah }, i) => {
+                const dutiesDiff = getDutiesDiff(
+                    oldDdah.duties,
+                    updatedDdah.duties
+                );
+                const {
+                    assignment: {
+                        applicant: { first_name, last_name, utorid },
+                    },
+                } = updatedDdah;
```
siefkenj: Don't do deep de-structuring.

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+                return (
+                    <li key={`ddah-diff-${i}`}>
+                        DDAH for {first_name} {last_name} ({utorid})
+                        {dutiesDiffList(dutiesDiff)}
```
siefkenj: Why aren't you using a JSX component? You're bypassing any optimizations that React chooses to make when you do it this way.

frontend/src/views/instructor/ddahs/import.tsx
```diff
@@ -0,0 +1,319 @@
...
+    if (newDdahs.length === 0 && ddahUpdates.length === 0) {
+        return (
+            <Alert variant="warning">
+                No difference between imported and existing DDAHs.
+            </Alert>
+        );
+    }
+
+    return (
+        <>
```
siefkenj: Write out `React.Fragment`. It makes it easier for the autocomplete.

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: So far for loops are faster, just because they are older, but `forEach` loops are easier to read in my opinion, especially when we don't need the iterator. And if we trade a tiny performance increase bound by the number of duties, I feel it would make sense if the code is easier to reason about. Let me know your thoughts though!

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: Not too sure I understand why, but ok, I'll change it. I don't have a preference either way.

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: I was thinking of some examples when we might want to "undo" some updates on import, which is not a feature we are actively looking to implement as I understand. So for now no particular reason, do you want me to change that to just `i`?

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: You mean `description` and `hours`?

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: I am not too sure I follow, which map and filters are you thinking of? :slightly_smiling_face:

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: I do prefer the `forEach` here, is there a reason why you'd prefer `map`? I understand it's more functional-like style, but out of these two:

```
changes = changes.concat(newDuties.map(duty => ({ newDuty: duty, status: DiffType.Created })))

newDuties.forEach((duty) =>
        changes.push({ newDuty: duty, status: DiffType.Created })
    );
```

The second one reads better to me. Let me know what you think!

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: If I take out the `switch` here, the whole component will contain of a `<ul>` followed by a `map`. The `switch` here is one level nested, so I was thinking it is fine? Let me know what you think :slightly_smiling_face:

frontend/src/views/instructor/ddahs/import.tsx
siefkenj: I disagree on both accounts. `forEach` loops are slower because they involve transferring execution to another function repeatedly. I also don't think they contribute to readability :-P

frontend/src/views/instructor/ddahs/import.tsx
siefkenj: You're pushing and splicing...try to rewrite this as a map and see how it looks.

frontend/src/views/instructor/ddahs/import.tsx
siefkenj: Well, you can do https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap if you prefer, but I like to avoid mutability where I can.

frontend/src/views/instructor/ddahs/import.tsx
siefkenj: Nope. Change it.

frontend/src/views/instructor/ddahs/import.tsx
siefkenj: Yes, please.

frontend/src/views/instructor/ddahs/import.tsx
siefkenj: Yes

frontend/src/views/instructor/ddahs/import.tsx
alex-kozin: I am not sure how to do that unfortunately, the `map` needs every old duty to be mapped to something that is not `undefined` (or else the types break) and for old duties that are deleted - they are "mapped" to `undefined`. That is why it's a loop, and not a map :)

frontend/src/components/import-button.js
siefkenj: This still seems to be an issue...

Issue #623, opened 2021-06-30
[UI] Money is rounded incorrectly


siefkenj:
When session rates are displayed (e.g. in the sessions table), they are rounded eagerly, making "46.70" show up as "46.7". Money should be rounded to two decimal places, always.

siefkenj:
Fixed by #658

PR #658, opened 2021-08-17
Rates are always displayed in 2 decimal places

mertkaya1033:
This PR was made for issue #623.

### DIFF
--- a/backend/app/services/wage_chunk_formatting_service.rb
+++ b/backend/app/services/wage_chunk_formatting_service.rb
@@ -17,7 +17,7 @@ class WageChunkFormattingService
                 start_date = I18n.l chunk[:start_date].to_date, format: :long
                 end_date = I18n.l chunk[:end_date].to_date, format: :long
                 hours = chunk[:hours]
-                rate = chunk[:rate]
+                rate = '%.2f' % chunk[:rate]
 
                 "#{hours} hours at $#{rate}/hour from #{start_date} to #{
                     end_date
--- a/backend/app/services/wage_chunk_formatting_service.rb
+++ b/backend/app/services/wage_chunk_formatting_service.rb
@@ -17,7 +17,7 @@ class WageChunkFormattingService
                 start_date = I18n.l chunk[:start_date].to_date, format: :long
                 end_date = I18n.l chunk[:end_date].to_date, format: :long
                 hours = chunk[:hours]
-                rate = '%.2f' % chunk[:rate]
+                rate = format('%.2f', chunk[:rate])
 
                 "#{hours} hours at $#{rate}/hour from #{start_date} to #{
                     end_date
--- a/frontend/src/components/editable-cell.tsx
+++ b/frontend/src/components/editable-cell.tsx
@@ -70,8 +70,6 @@ export function EditableCell<T extends HasId, Field extends keyof T = keyof T>({
     editable?: boolean;
 }): React.ReactElement {
     const title = `Edit ${"" + column.Header}`;
-    const isDate = type === "date";
-    const isMoney = type === "money";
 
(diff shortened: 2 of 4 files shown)


backend/app/services/wage_chunk_formatting_service.rb
```diff
@@ -17,7 +17,7 @@ def pay_period_description
                 start_date = I18n.l chunk[:start_date].to_date, format: :long
                 end_date = I18n.l chunk[:end_date].to_date, format: :long
                 hours = chunk[:hours]
-                rate = chunk[:rate]
+                rate = '%.2f' % chunk[:rate]
```
alex-kozin: Just to clarify, this is to display money with 2 decimals in the PDFs?

alex-kozin (commented):
Very clean, I like it! :rocket:

backend/app/services/wage_chunk_formatting_service.rb
mertkaya1033: Yes. It does work but I don't understand why Ruby Formatting tests fail.

frontend/src/libs/utils.ts
```diff
@@ -187,3 +187,14 @@ export function guessActiveSession(sessions: Session[]): Session | null {
 
     return null;
 }
+
+/**
+ * Formats the input "number" string to 2 decimal places.
+ * @param {string} moneyString
+ */
+export function formatMoney(moneyString: string): string {
+    if (!moneyString) {
```
siefkenj: What if `moneyString == 0`?

I think it would be better to do an `isFinite` check on the result of `parseFloat`

frontend/src/components/editable-cell.tsx
```diff
@@ -81,6 +83,12 @@ export function EditableCell<T extends HasId, Field extends keyof T = keyof T>({
     }
 
     const formattedValue = isDate ? ("" + value).slice(0, 10) : value || "";
+    let displayedValue = isDate ? formatDate("" + value) : value;
+
+    if (isMoney) {
+        displayedValue = value != null ? formatMoney("" + value) : "";
+        type = "number";
```
siefkenj: I don't like overriding input params. You should save the value in a new variable, e.g. `inputType`.

frontend/src/components/editable-cell.tsx
```diff
@@ -81,6 +83,12 @@ export function EditableCell<T extends HasId, Field extends keyof T = keyof T>({
     }
 
     const formattedValue = isDate ? ("" + value).slice(0, 10) : value || "";
+    let displayedValue = isDate ? formatDate("" + value) : value;
```
siefkenj: We're getting complicated here. Probably better to use a switch statement to handle all cases at once.

frontend/src/components/editable-cell.tsx
```diff
@@ -81,6 +83,12 @@ export function EditableCell<T extends HasId, Field extends keyof T = keyof T>({
     }
 
     const formattedValue = isDate ? ("" + value).slice(0, 10) : value || "";
```
siefkenj: What's the difference between `formattedValue` and `displayValue`? If they're both needed, then there should be some comments explaining what they're doing.

backend/app/services/wage_chunk_formatting_service.rb
siefkenj: Look at your error from Rubocop. You can google it to see how the code should be modified.

frontend/src/libs/utils.ts
mertkaya1033: `moneyString` has the type string, so it cannot be 0. However, I do agree that `isFinite` check should be used.

PR #683, opened 2022-06-13
Adding application seed data

elaine-huynh:
- Auto-generated application data (e.g., program, department, YIP, GPA, position preferences) using the existing user and applicant mock data as a basis
- Added button in dev-frame to upsert application data


frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -145,7 +151,7 @@ export function SeedDataMenu({
         setProgress(100);
     }
 
-    async function seedUsers(limit = 1000) {
+    async function seedUsers(limit = 1053) {
```
siefkenj: I think we can cut this down to 600. That should speed things up but also be realistic.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +269,70 @@ export function SeedDataMenu({
         }
     }
 
+    async function seedApplications(limit=1000) {
```
siefkenj: 400 to match the users

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +269,70 @@ export function SeedDataMenu({
...
+        setStage("Applications");
+        setProgress(0);
+
+        if (!targetSession) {
+            throw new Error("Need a valid session to continue");
+        }
+
+        count = 0;
```
siefkenj: Always use `let` or `const` when creating variables. eslint should be giving you a bunch of warnings in the console about this.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +269,70 @@ export function SeedDataMenu({
...
+        // Keep track of the original active user so we can swap back
+        const initialUser = await dispatch(fetchActiveUser());
+
+        // Get this session's posting token:
+        let url_token = "";
+        await apiGET(`/admin/sessions/${targetSession.id}/postings`).then((resp: any) => {
+            url_token = resp[0].url_token;
+        })
+
+        for (const a of seedData.applications.slice(0, limit)) {
```
siefkenj: Avoid single-letter variables. Make it descriptive.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +269,70 @@ export function SeedDataMenu({
...
+
+        count = 0;
+        total = seedData.applications.length;
+
+        // Keep track of the original active user so we can swap back
+        const initialUser = await dispatch(fetchActiveUser());
+
+        // Get this session's posting token:
+        let url_token = "";
+        await apiGET(`/admin/sessions/${targetSession.id}/postings`).then((resp: any) => {
```
siefkenj: You're mixing promise formats.


```
const resp = await apiGET(...);
let url_token = resp[0].url_token;
```

I'm not happy about the hardcoded `0`. How do you know that that is the correct posting (someone may have created a posting and the clicked the seed button).

Also, work hard to avoid using `any`. It is an escape hatch for extreme times... In this case `resp: RawPosting[]`, right?

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +269,70 @@ export function SeedDataMenu({
...
+            url_token = resp[0].url_token;
+        })
+
+        for (const a of seedData.applications.slice(0, limit)) {
+            const currUser = {"utorid": a.utorid, "roles": ["admin", "instructor", "ta"]};
+            await dispatch(debugOnlySetActiveUser(currUser, {skipInit: true}));
+
+            const newApp = {
+                "utorid": a.utorid,
+                "student_number": a.student_number.toString(),
```
siefkenj: Student number is already a string. If it weren't use `""+a.student_number` or `String(a.student_number)`. Those two function will never error, but `a.student_number.toString()` will error if `a.student_number` is null or undefined.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +269,70 @@ export function SeedDataMenu({
...
+                "previous_department_ta": a.previous_department_ta,
+                "previous_university_ta": a.previous_university_ta,
+                "program_start": a.program_start,
+                "previous_other_university_ta": a.previous_other_university_ta,
+                "position_preferences": a.position_preferences,
+            };
+
+            if (a.gpa) {
+                newApp.gpa = a.gpa;
+            }
```
siefkenj: Why is this here? Use `{"gpa": a.gpa || 0}` in your original assignment.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,126 @@ export function SeedDataMenu({
...
+                { answers: newApp },
+                true
+            );
+
+            count++;
+            setProgress(Math.round((count / total) * 100));
+        }
+
+        await dispatch(
+            debugOnlySetActiveUser({
```
siefkenj, line 348: You should just be able to do

```
dispatch(debugOnlyActiveUser(initialUser))
```

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,126 @@ export function SeedDataMenu({
...
+
+        if (!targetSession) {
+            throw new Error("Need a valid session to continue");
+        }
+
+        let count = 0;
+        let total = seedData.applications.length;
+
+        for (const application of seedData.applications.slice(0, limit)) {
+            if (application.instructor_preferences) {
```
siefkenj, line 369: I prefer early exit rather than putting everything in an if. e.g.

```
if (!application.instructor_preferences) {
    continue;
}
```

then everything else doesn't need to be indented :-)

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,126 @@ export function SeedDataMenu({
...
+                        (currApplication) =>
+                            currApplication.applicant.utorid ===
+                            application.utorid
+                    ) || null;
+
+                if (!targetApplication) {
+                    throw new Error("No application found for " + application);
+                }
+
+                for (const position of application.instructor_preferences) {
```
siefkenj, line 381: `position` should be called `preference`, right?

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,126 @@ export function SeedDataMenu({
...
+                    const targetPosition =
+                        positions.find(
+                            (currPosition) =>
+                                currPosition.position_code ===
+                                position.position_code
+                        ) || null;
```
siefkenj, line 387: We've got cubic behaviour here...it might be worth it to create a has of position_code -> position so that this lookup can be more efficient.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,126 @@ export function SeedDataMenu({
...
+                            "No position found for " + position.position_code
+                        );
+                    }
+
+                    await apiPOST(`/instructor/instructor_preferences`, {
+                        preference_level: position.preference_level,
+                        comment: position.comment,
+                        application_id: targetApplication.id,
+                        position_id: targetPosition.id,
+                    });
```
siefkenj, line 400: I feel weird about this loop. I think you should create an array of what needs to be upserted, and then, in a separate loop upsert the data via async calls.

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,128 @@ export function SeedDataMenu({
...
+        let total = seedData.applications.length;
+
+        // Keep track of the original active user so we can swap back
+        const initialUser = await dispatch(fetchActiveUser());
+
+        // Get this session's posting token:
+        const resp = await apiGET(
+            `/admin/sessions/${targetSession.id}/postings`
+        );
+        let url_token = resp[0].url_token;
```
siefkenj, line 307: You should check and throw an error early if `resp` has zero length

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,128 @@ export function SeedDataMenu({
...
+                applications.find(
+                    (currApplication) =>
+                        currApplication.applicant.utorid === application.utorid
+                ) || null;
+
+            if (!targetApplication) {
+                throw new Error("No application found for " + application);
+            }
+
+            for (let preference of application.instructor_preferences) {
```
siefkenj: Use `const`

frontend/src/views/dev_frame/load-mock-data.tsx
```diff
@@ -263,6 +275,128 @@ export function SeedDataMenu({
...
+                preferences.push({
+                    preference_level: preference.preference_level,
+                    comment: preference.comment,
+                    application_id: targetApplication.id,
+                    position_id: targetPosition.id,
+                });
+            }
+        }
+
+        for (let preference of preferences) {
```
siefkenj: use `const`

PR #687, opened 2022-08-16
Matching bug fixes

### LINKED ISSUE #686: Issue tracker for matching
This issue is to track bugs and to-dos for the matching component of TAPP.

### High Priority
- [x] Whole page reloads when assigning someone, changing their hours, then pressing "enter" in the dialog box
- [x] Pressing "enter" in the filter text box reloads the page
- [x] Pressing "enter" in the position filter box reloads the page
- [x] If all sorts are removed, there are no applicants displayed
- [x] There's weird behavior with the sorting. It appears to be unstable. If an irrelevant sort item (e.g., "hours owed" when there is no sub appt data available) the items reorder, but they shouldn't.
- [ ] Local storage
- [x] #695

### Mid Priority
- [ ] If a course doesn't have a default number of hours to assign, it assigns them zero hours. That is weird. Please use the following algorithm:
    - [ ] If a course has no default hours or the default is zero assignment and there are no others assigned/staged, automatically open the "adjust hours" dialog when they are assigned.
    - [ ] If there are any assigned/staged assignments and there is no default hours assignment or the default is zero, make the default the mode of the assigned/staged hours value.
    - [ ] If there is a non-zero default assignment, use that.
- [ ] Applicant Grid Items
    - [ ] The pills should all be Buttons so that they are keyboard navigable. Enter/space should open the context menu.
    - [x] #696
- [ ] Table View
    - [x] The AdvancedFilterTable in the table view is not scaled to be the height of the page
    - [x] The default column sizes in the table view should be set better (E.g., YIP, GPA, TA Preference should be skinny) 
    - [x] A null GPA should not show up as "0", it should be blank.
    - [ ] There is no way to view the application from the table view; this is part of a larger issue where the "table view" doesn't actually have much functionality tied to it (i.e., you can't perform actions like starring, hiding, assigning through this view)
- [x] Position List
    - [x] It's hard to tell what course is currently active in the left bar. The bigger left border is cute, but it is confusing because the width of the bar also represents progress, so it is being used for two different things.
    - [x] The yellow progress bar is hard to see.
- [x] #688
    - [ ] The Import Appt. Data dialog should, by default, explain the acceptable data format
    - [x] The import dialogs don't close when you "click away" from them; all the dialogs need the "X" button to close them
- [ ] The Finalize Changes dialog should have checkboxes (default with all rows checked) and only those checked assignments should be exported.

### Low Priority
- [x] The Filter Applicants dialog has a scroll bar when it shouldn't need one.
- [ ] There appears to be weirdness in the horizontal scrollbar in the matching view. Both the horizontal and vertical scrollbars always appear.
- [x] The "Showing X/Y applicants" should be pushed to the bottom of the screen. It should have a title  attribute explaining how to show hidden applicants if there are indeed any hidden applicants
- [x] Import Appt. Data should be  Import Sub. Appt. Data
- [ ] Not matching-specific: In the application dialog, the preferences should be sorted by preference level (highest to lowest) and then by course code alphabetically.
- [ ] Position and applicant lists seem to resize in width when applicant list is empty/all filtered out

elaine-huynh:


frontend/src/views/admin/matching/applicant-view/grid/modals.tsx
```diff
@@ -67,7 +67,7 @@ export function AdjustHourModal({
                 <Modal.Title>Update Hours</Modal.Title>
             </Modal.Header>
             <Modal.Body>
-                <Form>
+                <Form onSubmit={(e) => e.preventDefault()}>
```
siefkenj: This is not the right solution. Other dialogs do not do this. You probably shouldn't be using `Form` controls at all.

frontend/src/views/admin/matching/position-list.tsx
```diff
@@ -92,17 +92,17 @@ function PositionRow({
 
     return (
         <div
-            className="position-row noselect"
+            className={`position-row noselect ${focused ? "selected" : ""} ${
```
siefkenj: Use `classNames(...)`

PR #690, opened 2022-08-21
Matching CSS and layout fixes

siefkenj:


frontend/src/views/admin/matching/styles.css
```diff
@@ -321,35 +287,33 @@
...
 .matching-course-header {
-    padding: 0 3px;
-    position: sticky;
-    top: 0;
-    background-color: white;
-    z-index: 10;
+    padding: 5px 10px 0px 10px;
+}
+.matching-course-header h2 {
+    margin-bottom: 0;
```
elaine-huynh, line 304: I think having a bit of a margin to distinguish between the header and body helps to keep things more visually organized.

Maybe `margin-bottom: 2px` or `margin-bottom: 3px`?

frontend/src/views/admin/matching/styles.css
```diff
@@ -321,35 +287,33 @@
...
 }
 
 .matching-course-body {
-    padding: 0 3px;
-    height: 100%;
     display: flex;
     flex-direction: column;
+    overflow-y: auto;
+    flex-grow: 1;
+    row-gap: 5px;
```
elaine-huynh, line 312: `row-gap: 10px` or `row-gap: 15px` look nicer to me and promote better separation of status categories.

frontend/src/views/admin/matching/styles.css
```diff
@@ -393,12 +357,13 @@
...
+    gap: 1px;
+    flex-wrap: wrap;
 }
 
 .sort-dropdown-container > * > .dropdown-menu.show {
     max-height: 250px;
-    overflow-y: scroll;
+    overflow-y: auto;
     font-size: smaller;
 }
```
elaine-huynh, line 369: Dropdown menu font size is quite large in comparison to the rest of the view.
Please add the following and remove `.applicant-dropdown-menu` a few lines down:
```
.applicant-pill + .dropdown-menu {
    font-size: smaller;
}
```

frontend/src/views/admin/matching/styles.css
```diff
@@ -176,58 +177,23 @@
     background-color: rgba(255, 204, 22, 0.25);
```
elaine-huynh, line 177: I think the other colours could be left at 0.25, but let's try bumping up the alpha channel of `.progress.under`:
`background-color: rgba(255, 204, 22, 0.4);`

The grey-ish background for the position list makes the transparent colours harder to see. This isn't really an issue for the blue and green since the whole bar is filled in those cases, but it exacerbates the issue of yellow being not very visible when representing progress.

frontend/src/views/admin/matching/styles.css
```diff
@@ -1,73 +1,62 @@
...
     display: flex;
     flex-direction: column;
+    flex-grow: 0;
+    flex-basis: 0;
 }
 
 .search-container {
     display: flex;
     align-items: center;
-    margin-bottom: 10px;
```
elaine-huynh, line 62: Maybe 10px is a bit too big, but I feel the absence of a margin entirely isn't that appealing either. Could we compromise with `margin-bottom: 5px`?

frontend/src/views/admin/matching/styles.css
```diff
@@ -84,55 +73,67 @@
...
     cursor: pointer;
+    position: relative;
     display: flex;
-    align-items: center;
-    height: 35px;
-    margin-right: 15px;
-    margin-bottom: 5px;
     box-shadow: 3px 2px 5px -3px #7d7d7d;
-    transition: height 0.4s ease-out;
+    transition: height 0.4s ease-out, flex-grow 0.4s ease-out;
```
elaine-huynh, line 108: Do transitions work properly on your browser with this code? Looks like the animation for changing height of a position row doesn't work anymore on Chrome. If this is intended, we can remove the `transition` line entirely.

If it isn't, I think we need to debug this a bit. Here's what I have on my mind at the moment: in order for a transition to work, we need to define values for the animated property to change between (e.g., `height: 35px` in `.position-row` and `height: 65px` in `.position-row.selected` in the old code), then add the property to the `transition` line.

In the current CSS:
- `height` is no longer defined in `.position-row` nor `.position-row.selected` so having it in the `transition` line does nothing
- The transition for `flex-grow` also doesn't seem to do anything since there's no value for `flex-grow` defined in any of the classes that this class co-exists with

So... I think we need to either creatively define `height` to both `.position-row` and `.position-row.selected` in a way that it automatically computes/adjusts height based on subelements being changed, or we add back the explicit definitions for height.

elaine-huynh (changes requested):
My comments are all about CSS and mostly about spacing. I think it's important to introduce spacing so that parts of the UI are emphasized or better perceived as being separate from one another, as well as to keep grouping within sets of elements. 🙂 

Alas, styling is a very subjective topic so feel free to omit my suggestions if they don't sit well with you. 👌

frontend/src/views/admin/matching/styles.css
siefkenj: No, the transitions don't work right. Setting an explicit size is one solution, but it is a bad solution since sometimes positions have long names and a fixed size won't wrap correctly.

The best solution is probably a React component wrapper that monitors the size of the inner content and then dynamically sets the height property.

siefkenj:
@elaine-huynh I've merged the PR as is. You've pointed out several places for cosmetic cleanups. Go ahead and make those adjustments in a new PR :-)

PR #691, opened 2023-01-18
Rejected offers

### LINKED ISSUE #695: Rejected (and withdrawn) contracts show up as "assigned" in the matching view.

elaine-huynh:
This pull request fixes the issue where rejected/withdrawn offers would count towards the total number of assigned hours for positions in matching data.

This fix also adds another category/match status ("unassignable") so that admins cannot re-assign these TAs via matching and instead need to do so via the Assignments tab. Unassignable applicants are hidden by default in the applicant list.

### DIFF
--- a/frontend/src/views/admin/matching/actions.ts
+++ b/frontend/src/views/admin/matching/actions.ts
@@ -245,12 +245,16 @@ export const matchesSelector = createSelector(
             // Override with official assignments
             for (const assignment of assignmentsByApplicantId[applicant.id] ||
                 []) {
-                matchesByPositionId[assignment.position.id] = {
-                    applicant: applicant,
-                    position: assignment.position,
-                    hoursAssigned: assignment.hours || 0,
-                    status: "assigned",
-                };
+                // Check if offer was rejected/withdrawn:
+                if (assignment.active_offer_status !== "rejected" &&
+                    assignment.active_offer_status !== "withdrawn") {
+                    matchesByPositionId[assignment.position.id] = {
+                        applicant: applicant,
+                        position: assignment.position,
+                        hoursAssigned: assignment.hours || 0,
+                        status: "assigned",
+                    };
+                }
             }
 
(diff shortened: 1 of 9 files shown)


frontend/src/views/admin/matching/actions.ts
```diff
@@ -245,11 +245,22 @@ export const matchesSelector = createSelector(
             // Override with official assignments
             for (const assignment of assignmentsByApplicantId[applicant.id] ||
                 []) {
+                let hours = assignment.hours || 0;
```
siefkenj: Something's making me uncomfortable about the order of code here.

I think it would be better to have an `activeOfferStatusToStatus(...)` function. Then you can check `hours = status === "unassignable" ? 0 : assignment.hours || 0`

frontend/src/views/admin/matching/actions.ts
```diff
@@ -260,6 +267,18 @@ export const matchesSelector = createSelector(
     }
 );
 
+/**
+ * Return a MatchStatus ("assigned" or "unassignable") based on an active offer status type.
+ */
+function activeOfferStatusToStatus(active_offer_status: string | null) {
+    switch (active_offer_status) {
+        case "rejected" || "withdrawn":
```
siefkenj: I don't think this works...doesn't `"rejected" || 
"withdrawn"` evaluate to `"rejected"`? Did you test it?

Use
```
case "rejected":
case "withdrawn":
  ...
```
to get both cases mapping to the same place.

frontend/src/views/admin/matching/applicant-view/grid/grid-view.tsx
```diff
@@ -100,6 +102,13 @@ function GridSection({
             Positions > Assignments tab."
                     />
                 )}
+                {header === "Unassignable" && (
```
siefkenj: I don't like comparing against display names. Display names aren't type checked. Can you compare against the status instead?

PR #694, opened 2023-01-23
Add star and note to matching dropdown & decouple functions from dropdown code

### LINKED ISSUE #696: Star and Add Note should be added to the pill's context menus

elaine-huynh:
This PR adds "star" and "add note" options to the dropdown lists when clicking on applicant pills. 

Some matching functions have also been decoupled from the dropdown and moved to a separate directory in order to support reuse in future updates (e.g., implementation of a student-based view and adding functionality to the table view).


frontend/src/views/admin/matching/match-actions/modify-match-status.tsx
```diff
@@ -0,0 +1,109 @@
+import React from "react";
+import { useThunkDispatch } from "../../../../libs/thunk-dispatch";
+import { upsertMatch } from "../actions";
+import { ApplicantSummary, MatchableAssignment } from "../types";
+
+/**
+ * Update a match to mark an applicant as being staged-assigned to a
+ * position, with `hoursAssigned` number of hours.
+ */
+export function AssignApplicantToPosition(
```
siefkenj: You're writing a bunch of custom hooks. Hooks should start with `use`, so, e.g., this should be called `useAssignApplicantToPosition`. The same for all the others.

When hooks are used, they should be assigned to constants first-thing in your components before any possible conditional logic.

frontend/src/views/admin/matching/match-actions/modify-match-status.tsx
```diff
@@ -0,0 +1,109 @@
+import React from "react";
```
siefkenj: There's no JSX in this file, so it should be `.ts`

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
```diff
@@ -107,28 +75,62 @@ export function GridItemDropdown({
...
+                            baseMatchValues.positionCode,
+                            baseMatchValues.utorid
+                        )}
+                    >
                         Unassign from <b>{match.position.position_code}</b>
                     </Dropdown.Item>
                 </>
             )}
+            {canBeStarred && (
+                <Dropdown.Item onClick={SetStarred(match, true)}>
```
siefkenj: This is very weird. You should have a `toggleStar` action instead. This should also be put in its own component, so that starring can be worried about in a single place.

frontend/src/views/admin/matching/applicant-view/grid/grid-item/applicant-star.tsx
```diff
@@ -1,31 +1,22 @@
...
-        );
-    }, [dispatch, match]);
-
     return (
         <BsStarFill
             className={classNames("star-icon", {
                 filled: match.status === "starred",
             })}
-            onClick={updateStarred}
+            onClick={useToggleStarred(
```
siefkenj: No. Hooks must be assigned to a variable at the top of the function and then used. Don't define them in component callbacks. If this changed to conditionally render something, the hook could randomly break.

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
```diff
@@ -35,57 +41,48 @@ export function GridItemDropdown({
...
-    const hideFromPosition = React.useCallback(() => {
-        dispatch(upsertMatch({ ...baseMatchValues, hidden: true }));
-    }, [baseMatchValues, dispatch]);
+    const assignApplicant = useAssignApplicantToPosition(
+        baseMatchValues.positionCode,
+        baseMatchValues.utorid,
+        match.position.hours_per_assignment || 0
+    );
+
+    const unassignApplicant = useUnassignApplicantFromPosition(
```
siefkenj: The assign and unassign look like they should be grouped and delivered using the same hook. Do you actually mean to toggle the assignment, or do you need these as separate functions?

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
```diff
@@ -35,57 +41,48 @@ export function GridItemDropdown({
...
-            dispatch(
-                upsertMatch({
-                    utorid: match.applicant.utorid,
-                    positionCode: targetMatch.position.position_code,
-                    hidden: true,
-                })
-            );
-        }
-    }, [dispatch, match, applicantSummary.matches]);
+    const hideFromPosition = useHideFromPosition(
```
siefkenj: Seems like the hide functions can also be delivered by one hook.

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
```diff
@@ -35,57 +41,48 @@ export function GridItemDropdown({
...
+        baseMatchValues.positionCode,
+        baseMatchValues.utorid,
+        true
```
siefkenj: These extra `true` arguments are very confusing. Why are they needed? Use object literals if you have lots of arguments so that they are self documenting.

frontend/src/views/admin/matching/match-actions/edit-applicant-note.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * A modal window allowing users to view and edit notes for an applicant.
+ */
+export function ApplicantNoteModal({
+    applicantSummary,
+    show,
+    setShow,
+}: {
+    applicantSummary: ApplicantSummary;
+    show: boolean;
+    setShow: (arg0: boolean) => void;
```
siefkenj: You can name this more descriptively than `arg0`.

frontend/src/views/admin/matching/match-actions/edit-applicant-note.tsx
```diff
@@ -0,0 +1,73 @@
...
+}) {
+    const dispatch = useThunkDispatch();
+    const [noteTemp, setNoteTemp] = React.useState(applicantSummary.note || "");
+
+    const updateApplicantNote = React.useCallback(
+        (note: string) => {
```
siefkenj: Why do you accept a `note` if you're not going to use it?

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
elaine-huynh: I think you're right about assign and unassign being a single hook/toggle and will update accordingly!
For hide/unhide from all positions (not just one), however, a toggle may be problematic?

For example, if you're hidden from one position and unhidden from the rest, "Hide from all courses" would cause you to be unhidden in that one position (but correctly hidden from the others).

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
elaine-huynh: useHideFromPosition's third argument is `isHidden: boolean` and is set to "true" in the cases where we want to hide from a single position / all positions, or "false" when we want to unhide instead.

(Not sure if I'm missing the point here, please clarify if so.)

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
```diff
@@ -33,59 +38,49 @@ export function GridItemDropdown({
...
-                stagedHoursAssigned: 0,
-            })
-        );
-    }, [baseMatchValues, dispatch]);
+    const canBeUnstarred = match.status === "starred";
 
-    const hideFromPosition = React.useCallback(() => {
-        dispatch(upsertMatch({ ...baseMatchValues, hidden: true }));
-    }, [baseMatchValues, dispatch]);
+    const toggleAssign = useToggleAssigned(
```
siefkenj: `toggleAssign` -> `toggleAssigned`

frontend/src/views/admin/matching/applicant-view/grid/grid-item/dropdown.tsx
siefkenj: The point is, when looking at the code it is not obvious what it is doing. If it were written as 
```
useHideFromPosition({positionCode: ..., utorid: ..., hide: ...})
```
it would be much clearer.

But, should this be a toggle like the others?

frontend/src/views/admin/matching/match-actions/view-application-detail.tsx
```diff
@@ -0,0 +1,44 @@
...
+    application: Application | null;
+    setShownApplication: (application: Application | null) => void;
+}) {
+    if (!application) {
+        return null;
+    }
+
+    return (
+        <Modal
+            show={!!application}
```
siefkenj: Use `Boolean(application)` instead.

frontend/src/views/admin/matching/reducers.ts
```diff
@@ -154,4 +156,114 @@ const matchingDataReducer = createReducer(initialState, {
...
+                matches: [...state.matches, newMatch],
+                updated: true,
+            };
+        }
+
+        // Item exists, so we have to update it
+        return {
+            ...state,
+            matches: state.matches
+                .map((match) => {
```
siefkenj: You can use `flatMap` for this and `return []` if you don't want to include an item.

frontend/src/views/admin/matching/reducers.ts
```diff
@@ -154,4 +156,114 @@ const matchingDataReducer = createReducer(initialState, {
...
+            };
+
+            return {
+                ...state,
+                matches: [...state.matches, newMatch],
+                updated: true,
+            };
+        }
+
+        return {
```
siefkenj, line 230: Code like this makes me wonder if we should just start using the `immer` library...Do you have experience with it? Did you want to investigate and see if it's worth the dependency?

siefkenj:
There appears to be a merge conflict to fix

frontend/src/views/admin/matching/match-actions/edit-applicant-note.tsx
```diff
@@ -0,0 +1,70 @@
...
+    setShow,
+}: {
+    applicantSummary: ApplicantSummary;
+    show: boolean;
+    setShow: (show: boolean) => void;
+}) {
+    const dispatch = useThunkDispatch();
+    const [noteTemp, setNoteTemp] = React.useState(applicantSummary.note || "");
+
+    const updateApplicantNote = React.useCallback(() => {
```
siefkenj: Why are you doing `useCallback` here? You're not getting the benefits of it on L60, since you're dynamically creating a surrounding callback.

siefkenj:
This is looking really close! Have you looked into immer yet? I think it would make the reducers look much nicer.

elaine-huynh:
I think immer looks good but perhaps should be a separate PR to incorporate it into other components as well? I do agree that it would probably make reducers look nicer.

siefkenj:
Okay. Immer can be done in a different PR.

PR #701, opened 2023-02-07
Switch to GitHub Actions

### LINKED ISSUE #699: Switch to Github Actions
Travis has caused a lot of headaches and stopped testing builds for my projects over the weekend (Feb 6, 2023). I suggest we move to Github actions.

@elaine-huynh Could you look into this? It would be high priority since CI needs to run before merging any future PRs.

elaine-huynh:
This PR resolves #699.

Travis CI configuration has been ported over to GitHub Actions and covers all tests previously handled on Travis. Production deployment should also work, automatically pushing things to the gh-pages branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,6 @@ jobs:
     - name: Clean up Docker
       run: docker-compose down
   production-build:
-    if: ${{ github.event.issue.push }}
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repo
@@ -58,7 +57,7 @@ jobs:
       if: ${{ failure() }}
       run: docker-compose logs
     - name: Deploy
-      if: ${{ success() }}
+      if: ${{ success() && github.event.issue.push }}
       uses: crazy-max/ghaction-github-pages@v3
       with:
         target_branch: gh-pages
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,6 +60,7 @@ jobs:
     - name: Clean up Docker
       run: docker-compose down
   production-build:
+    if: ${{ github.event.issue.push }}
     needs: [code-formatting, ruby-linting, unit-tests]
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Docker Image CI
+name: TAPP CI
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -7,12 +7,79 @@ on:
...
+    - name: Checkout repo
+      uses: actions/checkout@v3
+    - name: Build Docker backend
+      run: |
+        cp dev.env .env
+        docker-compose up --build -d backend
+        docker ps -a
+    - name: Run Ruby formatting tests
+      run: 
+        ./tests/run-ruby-linting-tests.sh
```
siefkenj: I don't think this test requires a build. You should be able to run both formatting tests in `code-formatting`

.github/workflows/build.yml
```diff
@@ -7,12 +7,79 @@ on:
...
+    - name: Checkout repo
+      uses: actions/checkout@v3
+    - name: Build Docker backend
+      run: |
+        cp dev.env .env
+        docker-compose up --build -d
+        docker ps -a
+        docker-compose run backend rake db:create
+        docker-compose run backend rake db:migrate
+        sleep 30
```
siefkenj: This worries me. Did the previous test scrip call `sleep`?

.github/workflows/build.yml
```diff
@@ -7,12 +7,79 @@ on:
...
+    - name: Build Docker
+      run: |
+        cp dev.env .env
+        docker-compose up --build -d
+        docker ps -a
+        docker-compose run frontend npm run dev-build
+    - name: Display Docker logs on failure
+      if: ${{ failure() }}
+      run: docker-compose logs
+    - name: Deploy
```
siefkenj, line 59: We don't want pull requests to run deploy, but we do want them to do a production build. Does the current script handle that?

Also, let's remove the `deploy` portion from the travis script so they don't fight with each other.

siefkenj (commented):
This is looking great!

.github/workflows/build.yml
elaine-huynh: It did! I'm removing this line in the newest push to see if it's fine without the call.
Edit: Seems fine. 👍

.github/workflows/build.yml
elaine-huynh: Initial PR would run deploy on every pull request, but a subsequent commit fixed it so that it only builds and deploys on push to master.

.github/workflows/build.yml
```diff
@@ -7,12 +7,62 @@ on:
...
+    - name: Run unit tests
+      run: ./tests/run-unit-tests.sh
+    - name: Display Docker logs on failure
+      if: ${{ failure() }}
+      run: docker-compose logs
+    - name: Clean up Docker
+      run: docker-compose down
+  production-build:
+    if: ${{ github.event.issue.push }}
+    needs: [code-formatting, unit-tests]
```
siefkenj: Let's drop this. It's useful to know if the build is broken separately from whether code formats correctly or unit tests run.

.github/workflows/build.yml
siefkenj: We want to build on all PRs, but we only want to deploy on pushes in master.

